Issue #1, opened 2022-10-10
Implement fundamental workflows


jorgepiloto:
🐞 **Problem**
There are common CI pipeline steps across different projects in PyAnsys. These steps are:

**Code style and tests**
- [x] Code style checking

**Documentation related**
- [x] Documentation style checking
- [x] Developers doc building and deployment
- [x] Stable doc building and deployment

**Software building related**
- [x] Creation of wheelhouse
- [x] Generation of wheel and sdist artifacts

*Software release related*
- [x] Private PyPI release publishing
- [x] Public PyPI release publishing
- [x] GitHub release

💡 **Solution**
Each one of previous steps should be a workflow itself. Reusable workflows should be seen like small modules.

Experience is that PyAnsys projects show small differences between them. These differences affect the workflows for the pipelines, making those kind of unique. 

We do not want to avoid opening a pull-request in a project, getting an error in the workflow, then coming to this repo to fix it and affecting lots of repositories.

PR #13, opened 2022-10-19
FEAT: add doc-build action

jorgepiloto:
Addresses for documentation building, as per #1.


jorgepiloto:
Merging this. I will validate this logic against https://github.com/ansys/ansys-templates/pull/164

PR #14, opened 2022-10-21
FEAT: add build-library, release-private and release-public actions

jorgepiloto:


build-library/action.yml
```diff
@@ -0,0 +1,54 @@
...
+      if: env.EXISTS_BUILD_REQUIREMENTS == 'true'
+      shell: bash
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements/requirements_build.txt
+
+    - name: "Install build dependencies from pyproject.toml"
+      if: env.EXISTS_BUILD_REQUIREMENTS == 'false' 
+      shell: bash 
+      run: python -m pip install .[build]
```
RobPasMue, line 42: Interesting... all projects now have to define the same targets?...

For example, in PyGeometry we only have docs and test extra targets. We don't have a "build" target.

release-github/action.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: "Display the structure of downloaded files"
+      shell: bash
+      run: ls -R
+
+    - name: "Release to GitHub"
+      uses: softprops/action-gh-release@v1
+      with:
+        files: |
+          ${{ inputs.library-name }}/**/*.whl
+          ${{ inputs.library-name }}/**/*.tar.gz
```
RobPasMue, line 29: We should also add the docs as part of the release (both in HTML and PDF version if possible...). That will force people to upload the documentation to the artifacts as a ZIP, and then download it here as well and push it to the release. What do you think? This may be left for a second round anyway - not a priority at all.

RobPasMue (approved):
Some minor comments but LGTM overall :) This is going to be so useful to homogenize all repos... love it!

release-github/action.yml
jorgepiloto: Oh, that's true! Good catch. So far, those have only been uploaded as artifacts during the build.

jorgepiloto: We can fix in this same pull-request, no worries.

build-library/action.yml
jorgepiloto: Yeah, this is something I have been thinking.

In the end, any Python library should be built using [build] and [twine], as these are the official libraries maintained by PyPA. Any build system should be compliant with those.

Maybe we could get rid of these checks and just impose:

```bash
python -m pip install build twine
python -m build && python -m twine check
```

RobPasMue: I agree. That would be the best default option I think =)

build-library/action.yml
```diff
@@ -0,0 +1,54 @@
...
+  
+    - name: "Setup Python"
+      uses: actions/setup-python@v4.2.0
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: "Check if requirements_build.txt file exists"
+      shell: bash
+      run: |
+        echo "EXISTS_BUILD_REQUIREMENTS=$(if [ -f requirements/requirements_build.txt ]; then echo 'true'; else echo 'false'; fi)" >> $GITHUB_ENV
```
MaxJPRey, line 30: Unfortunately, Sometimes the  `requirements_*.txt` files are not located into the `requirements` folder but directly at the root of the repo.
We could also check for that.

build-library/action.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: "Install documentation dependencies from requirements file"
```
MaxJPRey, line 32:
```suggestion
    - name: "Install build dependencies from requirements file"
```

release-private/action.yml
```diff
@@ -0,0 +1,62 @@
...
+runs:
+  using: "composite"
+  steps:
+  
+    - name: "Set up Python"
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ env.MAIN_PYTHON_VERSION }}
+
+    - name: "Install required dependencies"
```
MaxJPRey, line 38:
```suggestion
    - name: "Install twine"
```

release-public/action.yml
```diff
@@ -0,0 +1,51 @@
...
+runs:
+  using: "composite"
+  steps:
+  
+    - name: "Set up Python"
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ env.MAIN_PYTHON_VERSION }}
+
+    - name: "Install required dependencies"
```
MaxJPRey, line 28:
```suggestion
    - name: "Install twine"
```

MaxJPRey (approved):
This is great.

build-library/action.yml
RobPasMue: Well that is definitely something that has to be solved on a per  repo basis... It is one of the basic recommendations. Whoever still has it that way, we can open a PR to solve it.

jorgepiloto: I guess you raised this concern since PyAEDT collects its requirements at base root directory level, @MaxJPRey. It should be feasible for us, as @RobPasMue, to open a pull-request and move those to a requirements directory.

release-github/action.yml
jorgepiloto: I updated the actions so PDF doc gets generated and uploaded, see https://github.com/pyansys/actions/pull/14/commits/8ba021f810f460f2a056525ac925cdffb7007f47.

Issue #15, opened 2022-10-21
Add action validation in GH CI/CD


RobPasMue:
### 📝 Description of the feature

Don't know if it is possible, just thinking here, but... it would definitely be useful to have a way of testing the actions before committing them. Against a dummy repo extracted from ansys-templates or similar. What do you think @jorgepiloto? I can investigate as well.

jorgepiloto:
I agree, this is something we need to explore.

The documentation deploy and release actions are the non-trivial ones. They expect a `gh-pages` branch and a repository to be published to.

RobPasMue:
Yeah, not saying it is straightforward to test them. But we should try to figure something out. Planning on investigating it for next week on my side =)

jorgepiloto:
Solved in #38. Closing.

PR #18, opened 2022-10-26
FIX: doc-build syntax

jorgepiloto:
Continuation of #17.


jorgepiloto:
Implementing #15 will solve for these sort of issues. Also, syntax for GitHub actions is very weird and non-intuitive...

PR #27, opened 2022-10-27
FIX: simplify build-library workflow

jorgepiloto:


RobPasMue (approved):
Okay so, now we delegate in whatever the `pyproject.toml`` file or ``setup.*`` file specifies, right? Sounds good to me!

jorgepiloto:
Exactly! I forgot to add these changes when you suggested those.

PR #48, opened 2022-11-10
FEAT: wheelhouse action

### LINKED ISSUE #1: Implement fundamental workflows

jorgepiloto:
Fix #1 by adding the wheelhouse action.

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -1,0 +1,62 @@
+name: "Build the wheelhouse of a Python library"
+description: "Generate a Zip aritfacts for the wheelhouse of a Python library"
+
+inputs:
+  library-name:
+    description: 'Name of the Python library'
+    required: true
+    type: string
+  library-namespace:
+    description: 'Namespace of the Python library'
+    required: true
+    type: string
+  operating-system:
+    description: 'Name of the operating system'
+    required: true
+    type: string
+  python-version:
+    description: 'Desired Python version for build'
+    required: true
+    type: string
+
+
+runs:
+  using: "composite"
+  steps:
+
+    - name: "Install Git and clone project"
+      uses: actions/checkout@v3
+  
+    - name: "Set up Python ${{ inputs.python-version }}"
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: "Update pip and install dependencies"
+      shell: bash
+      run: pip install --upgrade pip build wheel
+
+    - name: "Install the library"
+      shell: bash
+      run: pip install .
+
+    - name: "Verify libray is properly installed and get its version number"
+      run: |
+        echo "library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)")" >> $GITHUB_ENV
+
+    - name: "Generate wheelhouse"
+      shell: bash
+      run: python -m pip wheel . -w wheelhouse
+
+    - name: "Zip wheelhouse"
+      uses: vimtor/action-zip@v1
+      with:
+        files: wheelhouse
+        dest: ${{ inputs.library-name }}-v${{ env.library_version }}-wheelhouse-${{ inputs.operating-system }}-${{ inputs.python-version }}.zip
+
+    - name: "Upload wheelhouse"
+      uses: actions/upload-artifact@v3
+      with:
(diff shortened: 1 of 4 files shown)


build-wheelhouse/action.yml
```diff
@@ -0,0 +1,62 @@
+name: "Build the wheelhouse of a Python library"
+description: "Generate a Zip aritfacts for the wheelhouse of a Python library"
```
RobPasMue, line 2:
```suggestion
description: "Generate ZIP aritfacts for the wheelhouse of a Python library"
```

build-wheelhouse/action.yml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: "Verify libray is properly installed and get its version number"
+      run: |
+        echo "library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)")" >> $GITHUB_ENV
+
+    - name: "Generate wheelhouse"
+      shell: bash
+      run: python -m pip wheel . -w wheelhouse
+
+    - name: "Zip wheelhouse"
```
RobPasMue, line 51: I thought we were going to proceed with ``cibuildwheel``

https://github.com/pyansys/pymapdl-reader/blob/8b0ace3f05e71b26fe632ab53538f2053bbee453/.github/workflows/testing-and-deployment.yml#L166-L167

doc/source/build-actions/index.rst
```diff
@@ -28,6 +28,38 @@ Here is a code sample for using this action:
...
+             os: [ubuntu-latest, windows-latest]
+             python-version: ['3.7', '3.8', '3.9', '3.10']
+      steps:
+        - name: "Build a wheelhouse of the Python library"
+          uses: pyansys/actions/build-wheelhouse@main
+          with:
+            library-name: "<ansys-product-library>"
+            library-namespace: "<ansys.product.libray>"
+            operating-system: ${{ matrix.os }}
+            python-version: ${{ matrix.python-version }}
```
RobPasMue, line 60: Why not just merge this action into a single call and provide the OS and Python versions as lists?

I guess that'd make it easier to use a plug-and-play stage of our workflow.

build-wheelhouse/action.yml
jorgepiloto: Hmm... I may be confused. I thought the wheelhouse packaged all the dependencies so you do not need to install those from an index while the cibuildwheel generated wheels for desired platform. Does the cibuildwheel also include all the dependencies?

doc/source/build-actions/index.rst
jorgepiloto: It is a great idea! Let me check if reusable workflows support matrix strategy, not sure about this...

RobPasMue: Let me know if it does! Sounds interesting =)

build-wheelhouse/action.yml
RobPasMue: Ah yeah... that's true. But I thought maybe cibuildwheel was also able to pack all dependencies. Let me check...

doc/source/build-actions/index.rst
jorgepiloto: It is possible to call a composite workflow inside a matrix strategy. However, it is not possible to execute a matrix inside a composite workflow.

RobPasMue: Damn... that's a shame =( thanks for looking into it anyway! =)

build-wheelhouse/action.yml
RobPasMue: Yeah, looks like it only builds the specific wheel, not the entire wheelhouse. Resolving comment.

RobPasMue (approved):

Issue #56, opened 2022-11-15
Consider adding caching mechanism for actions/setup-python


RobPasMue:
Action ``actions/setup-python@v4`` allows to implement an automated caching mechanism as follows:

```yml
      - name: Set up Python
        uses: actions/setup-python@v4
        with:
          python-version: '3.11'
          cache: 'pip'
          cache-dependency-path: 'pyproject.toml'
```

We should consider exposing our dependencies location (as well as the package manager, though this one is typically pip) so that the action can generate the hash adequately and cache the packages.

RobPasMue:
By default, I think it should be either the requirements files or the pyproject.toml.

jorgepiloto:
From private discussion, we agreed on using the `pyproject.toml` file as reference. Main reasons are:

* It is an official file as opposite to requirements.
* Dependabot supports now parsing of extra dependencies within `pyproject.toml`.

RobPasMue:
Sounds right! Let's start migrating our different requirements to pyproject.toml as well, now that dependabot and flit get along well :)

PR #58, opened 2022-11-16
FEAT: add private _setup-python action and force cache usage

jorgepiloto:
This pull-request solves #56 by adding a "private" action for setting up Python. This action is then reused across all the different actions in the repository.

### DIFF
--- a/_release-pypi/action.yml
+++ b/_release-pypi/action.yml
@@ -28,7 +28,7 @@ runs:
   steps:
   
     - name: "Set up Python"
-      uses: actions/setup-python@v4
+      uses: pyansys/actions/_setup-python@main
       with:
         python-version: ${{ env.MAIN_PYTHON_VERSION }}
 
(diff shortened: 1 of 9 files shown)


RobPasMue:
Looks like the best solution! Let me review it

_release-pypi/action.yml
```diff
@@ -28,7 +28,7 @@ runs:
   steps:
   
     - name: "Set up Python"
-      uses: actions/setup-python@v4
+      uses: pyansys/actions/_setup-python@main
       with:
```
RobPasMue, line 32: Remember that whenever we do a release, we should change all these "@main" tags in the release branch to point to its corresponding release :)

RobPasMue (approved):
LGTM!

_release-pypi/action.yml
RobPasMue: lol, didn't know there was a guy named "main" hahaha!

jorgepiloto: That's right! I noticed this too while working with this. Let's keep that in mind.

PR #64, opened 2022-11-16
FIX: secrets usage in doc-deploy-to-repo

jorgepiloto:
I just remembered that secrets cannot be used in reusable actions. They need to be passed as arguments. This solves the latest CI failures related with this issue.


RobPasMue:
Oh really? That's a shame...

PR #69, opened 2022-11-17
FIX: do not use cache in doc-deploy-stable

jorgepiloto:
As per [pygeomemtry failing actions](https://github.com/pyansys/pygeometry/actions/runs/3487427530/jobs/5835475424). This will be cherry-picked to the release/1.0. I will also move the v1 tag.


RobPasMue (approved):
Awesome @jorgepiloto, thanks!

PR #75, opened 2022-11-21
Accepting additional targets in wheel build

### LINKED ISSUE #74: Adding target (as optional argument) to build-wheel action
Although one typically builds the wheel of a project by pointing at its necessary dependencies, one could also build a wheel of a package taking into account optional dependencies. For example:

**Default (typical) case:**
```bash
pip wheel . -w wheelhouse
```
**Wheel with optional targets:**
```bash
pip wheel .[my-target] -w wheelhouse
```

This use-case is specially interesting for the ``pyansys`` metapackage, since we are interested in generating the wheels of the ``all`` target - that is, all PyAnsys packages inside it.

Consider adding it and bumping version. I will be implementing it shortly. This argument should definitely be optional - since it is not the typical use case.

RobPasMue:
Closes #74 (hopefully)

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -51,6 +51,7 @@ runs:
       shell: bash
       run: |
         echo "install_target=$(( ${{ inputs.target }} == '' ? '.' : '.[${{ inputs.target }}]'))" >> $GITHUB_ENV
+        echo "wheelhouse_target=$(( ${{ inputs.target }} == '' ? 'wheelhouse' : '${{ inputs.target }}-wheelhouse'))" >> $GITHUB_ENV
 
     - name: "Install the library"
       shell: bash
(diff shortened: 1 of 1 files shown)


jorgepiloto (approved):

build-wheelhouse/action.yml
```diff
@@ -42,9 +47,14 @@ runs:
       shell: bash
       run: python -m pip install --upgrade pip build wheel
 
+    - name: "Check if specific target is requested"
+      shell: bash
+      run: |
+        echo "install_target=$(( ${{ inputs.target }} == '' ? '.' : '.[${{ inputs.target }}]'))" >> $GITHUB_ENV
```
jorgepiloto, line 53: I had no idea this syntax could be used in shell script for conditionals. Nice.

PR #84, opened 2022-11-23
FEAT: enable outdated versions announcement logic

### LINKED ISSUE #80: Add Depreciation warning/page for very old version of docs.

Add the label depreciated/ add a depreciated pages for the very old versions of the version releases.

This needs to be added through the versions.json

 Adding numpy https://numpy.org/doc/1.14/index.html for ref.

jorgepiloto:
Fix #80.

Repositories should implement an `announcement.html` template in their `gh-pages` branch and within the `release/` folder.

The content of the `announcement.html` should look like:

```html
<p>You are not reading the most recent version of this documentation. Latest stable release is <a href="https://actions.pyansys.com/release/{version}">{version}</a></p>.
```

This way, the version script can format that template to the desired version.

### DIFF
--- a/scripts/version_mapper.py
+++ b/scripts/version_mapper.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import re
 from pathlib import Path
 
 
@@ -93,20 +94,18 @@ def update_switch_version_file(
         json.dump(new_content, switcher_file, indent=4)
 
     # Use the latest stable verion for formatting the announcement
-    with open(annoucement_filename, "r") as announcement_file:
+    with open(f"release/{annoucement_filename}", "r") as announcement_file:
         content = announcement_file.read()
         announcement_content = content.format(version=latest_stable_version)
 
     # Include the announcement in all available release folders. Note that
     # these are still accessible even if they are not included in the dropdown.
-    release_folders = [path for path in Path("./").iterdir() if path.is_dir()]
-    for release_folder in release_folders:
-        # Skip the latest stable version
-        if str(release_folder) == latest_stable_version:
-            continue
-
+    old_release_folders = [
+        path for path in Path("./").iterdir() if re.match("^[0-9]+.[0-9]+$", path.name)
+    ]
+    for release_folder in old_release_folders:
         # Create an 'announcement.html' file within each one of the old versions
-        with open(f"{str(release_folder)}/announcement.html", "w") as file:
+        with open(f"release/{release_folder.name}/announcement.html", "w") as file:
             file.write(announcement_content)
 
 
--- a/scripts/version_mapper.py
+++ b/scripts/version_mapper.py
@@ -20,7 +20,7 @@ def sort_versions_descending(versions_list):
 
 
 def update_switch_version_file(
-    json_filename, new_version, cname, render_last, annoucement_filename
+    json_filename, new_version, cname, render_last, announcement_filename
 ):
     """Add new version number and associated URL to JSON file.
 
@@ -34,7 +34,7 @@ def update_switch_version_file(
         The canonical name of the project's documentation website.
     render_last : int
         The number of stable releases to be shown in the version switcher.
-    announcement_filename : str
+    announcement_file : str
         Name of the HTML file controlling the outdated version announcement.
 
     """
@@ -94,7 +94,7 @@ def update_switch_version_file(
         json.dump(new_content, switcher_file, indent=4)
 
     # Use the latest stable verion for formatting the announcement
-    with open(f"release/{annoucement_filename}", "r") as announcement_file:
+    with open(f"release/{announcement_filename}", "r") as announcement_file:
         content = announcement_file.read()
         announcement_content = content.format(version=latest_stable_version)
 
--- a/scripts/version_mapper.py
+++ b/scripts/version_mapper.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+from pathlib import Path
 
 
 def sort_versions_descending(versions_list):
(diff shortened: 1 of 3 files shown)


scripts/version_mapper.py
```diff
@@ -80,14 +88,38 @@ def update_switch_version_file(filename, new_version, cname, render_last):
...
+        with open(f"{str(release_folder)}/announcement.html", "w") as file:
+            file.write(announcement_content)
+
 
 def parse_cli_arguments():
     """Parse all command line arguments."""
     parser = argparse.ArgumentParser(description="Version switcher JSON file updater.")
+    parser.add_argument(
+        "-a",
+        "--announcement_filename",
```
RobPasMue, line 118:
```suggestion
        "--announcement-filename",
```

I am not much of a fan of underscores for CLI arguments

RobPasMue: This applies to the other arguments as well if you wish to change them - but totally up to you =). You can disregard my suggestion

scripts/version_mapper.py
```diff
@@ -63,6 +68,9 @@ def update_switch_version_file(filename, new_version, cname, render_last):
         else:
             new_versions_list = current_versions_list
 
+        # Get the latest stable version
+        latest_stable_version = new_versions_list[0]
+
```
RobPasMue, line 73: Are we sure that element 0 is not the ``dev`` version? Just asking

RobPasMue (approved):
LGTM!

scripts/version_mapper.py
jorgepiloto: Yes, me neither... The issue is that, the name you place in here is the name for the variable and Python does not allow dash-separated variable names.

For example, the `--anouncement-filename` would be accessed with `args.announcement-file`, leading to an error.

scripts/version_mapper.py
jorgepiloto: Yes, we are. When collecting available version numbers, the `dev` is ignored. When building the new `versions.json`, the `dev` is added on its own. Thus, `new_versions_list` only contains version numbers in descending order:

```pycon
>>> new_versions_list
["0.3", "0.2", "0.1"]
```

RobPasMue: Okay perfect then! Thanks @jorgepiloto for confirming it!

scripts/version_mapper.py
```diff
@@ -80,14 +88,38 @@ def update_switch_version_file(filename, new_version, cname, render_last):
...
+        content = announcement_file.read()
+        announcement_content = content.format(version=latest_stable_version)
+
+    # Include the announcement in all available release folders. Note that
+    # these are still accessible even if they are not included in the dropdown.
+    release_folders = [path for path in Path("./").iterdir() if path.is_dir()]
+    for release_folder in release_folders:
+        # Skip the latest stable version
+        if str(release_folder) == latest_stable_version:
+            continue
```
Revathyvenugopal162, line 106: I was thinking, should we add announcement file for all the version, and on the latest release, the announcement will be "You are on the latest release" or any customized announcement ?

scripts/version_mapper.py
RobPasMue: I see... the way you access the variable is the same as the CLI name you provide... Okay, no problem then 😄 I'll live with it, no biggie

scripts/version_mapper.py
RobPasMue: I wouldn't do that. You only announce stuff if you have something to announce. If you are on the latest docs... I think there is no need to do that. I haven't seen that behaviour anywhere.

scripts/version_mapper.py
```diff
@@ -17,22 +19,26 @@ def sort_versions_descending(versions_list):
     return versions_list[::-1]
 
 
-def update_switch_version_file(filename, new_version, cname, render_last):
+def update_switch_version_file(
+    json_filename, new_version, cname, render_last, annoucement_filename
```
Revathyvenugopal162, line 23:
```suggestion
    json_filename, new_version, cname, render_last, announcement_filename
```

tests/tests_scripts/test_version_mapper.py
```diff
@@ -31,10 +31,11 @@ def test_update_switch_version_file(new_version, render_last, cname):
 
     # Execute the logic behind the version update script
     update_switch_version_file(
-        filename="versions.json",
+        json_filename="versions.json",
         new_version=new_version,
         cname=cname,
         render_last=render_last,
+        annoucement_filename="announcement.html",
```
Revathyvenugopal162, line 38:
```suggestion
        announcement_filename="announcement.html",
```

scripts/version_mapper.py
```diff
@@ -80,14 +89,36 @@ def update_switch_version_file(filename, new_version, cname, render_last):
 
     # Override the whole content of the version switches JSON file with the new
     # generated version data
-    with open(f"release/{filename}", "w") as switcher_file:
+    with open(f"release/{json_filename}", "w") as switcher_file:
         # Update JSON file with the new content
         json.dump(new_content, switcher_file, indent=4)
 
+    # Use the latest stable verion for formatting the announcement
+    with open(f"release/{annoucement_filename}", "r") as announcement_file:
```
Revathyvenugopal162, line 97:
```suggestion
    with open(f"release/{announcement_filename}", "r") as announcement_file:
```

RobPasMue (approved):
LGTM! That regex is nice! :)

Revathyvenugopal162 (approved):
LGTM :), i think the typo is intentional, so correcting is up to you. Thanks @jorgepiloto

jorgepiloto:
> LGTM! That regex is nice! :)

Thanks! It is for filtering other directories that may be present, like `__pycache__`. Just in case.

scripts/version_mapper.py
jorgepiloto: Ops! Good catch, @Revathyvenugopal162. I'm likely to have introduced this through the usage of my LSP. Let me fix it.

PR #86, opened 2022-11-24
DBG: check licenses

### LINKED ISSUE #85: Add license check action

Add a new action ``check-licenses``. This action should also be incorporated as the last step of the already existing action ``build-library``.

The content of this ``check-licenses`` action should be:

1. Define accepted licenses. For instance, contains ``MIT``, contains ``BSD``, etc. See the accepted licenses in pyansys/pyansys [here](https://github.com/pyansys/pyansys/blob/main/.github/workflows/dependency-reviewer.yml).
2. Use ``pip-licenses`` to list all packages with its licenses 
    2.1. Ignore some trusted packages with different licenses that are not accepted. This can be done using the [ignore package option](https://github.com/raimon49/pip-licenses#option-ignore-packages)
    2.2. List the remaining packages' licenses, see [DPF action](https://github.com/pyansys/pydpf-actions/blob/v2.2/check-licenses/action.yml). ``pip-licenses --summary`` can be used to only list the licenses and the number of packages under each license.
3. Parse the table content and check that all licenses are among the accepted ones (step 1).

- https://github.com/pyansys/pyansys/blob/main/.github/workflows/dependency-reviewer.yml
- https://github.com/pyansys/pydpf-actions/blob/v2.2/check-licenses/action.yml
- https://github.com/raimon49/pip-licenses#option-ignore-packages

GuillemBarroso:
Fix #85.

Debugging check licenses behavior. 

The ci_cd,yml has been modified to only execute this action for debugging purposes.


check-licenses/action.yml
```diff
@@ -0,0 +1,34 @@
...
+
+runs:
+  using: "composite"
+  steps:
+
+    - name: "Check licences of packages"
+      shell: bash
+      run: |
+        echo ${{ inputs.accepted-licenses }}
+        pip install pip-licenses
```
jorgepiloto, line 19: Try to execute any Python module using `python -m ...`. This ensures that you use the modules within your environment and not the binaries.

check-licenses/action.yml
```diff
@@ -0,0 +1,34 @@
...
+        input=${{ inputs.accepted-licenses }}
+        accepted_licenses=''
+        for license in ${input[*]}; do accepted_licenses+="$license\|"; done
+        echo $accepted_licenses
+        accepted_licenses=${accepted_licenses::-2}
+        echo $accepted_licenses
+        echo pip-licenses
```
jorgepiloto, line 26: No need to echo a CLI tool that already outputs to stdout.

jorgepiloto (changes requested):

check-licenses/action.yml
GuillemBarroso: Got it!

check-licenses/action.yml
GuillemBarroso: Thats why I added both, with echo (in line 26) and without it (in line 27), just to see if any of them was printing in the logs

GuillemBarroso: I've removed the unnecessary echos. However, it looks like it outputs to stdout an empty string:

image

I am able to print the ``accepted_licenses`` variable (strings to find in the ``pip_licenses --summary``) before and after removing the last two characters. But then I get 3 empty lines for the 3 prints that was supposed to show.

jorgepiloto: The issue is that you have a clean Python environment without any package installed in the CI.

jorgepiloto: It is working fine. Try to install numpy and matplotlib, so the Python environment gets cluttered with dependencies and packages.

jorgepiloto:
To complete this action:

- [ ] Add a `python-version` as input.
- [ ] Add also the `skip-install` option, to avoid installing the project. This is just in case we want to reuse this action inside another action which already installs the project.

check-licenses/action.yml
```diff
@@ -0,0 +1,28 @@
+name: "Check licenses"
+description: "Check if package licenses in the current environment are among the accepted ones"
+
+inputs:
+  accepted-licenses:
+    description: "List of accepted licenses"
+    default: "MIT BSD Apache PSF HPND"
+    required: false
+    type: string
```
RobPasMue, line 9: What about ignored packages? Shouldn't we provide them as a list as well? The actions should also have a record file of accepted packages whose licenses are accepted (even if they are not of the types accepted by default). Don't know if you follow me but we can talk about this offline =)

GuillemBarroso: yep, I'm on it. So far I'm only excluding matplotlib for testing purposes.

Now I'm working on making it fail if bad licenses are found. After that I will stop hard-coding the inputs and pass them properly. I am planning on adding these inputs:
- ``accepted-licenses``: with some default well-known licenses (I took the ones you had [here](https://github.com/pyansys/pyansys/blob/main/.github/workflows/dependency-reviewer.yml#L22), @RobPasMue).
- ``additional-accepted-licenses``: just in case some projects want to add one or two more to the ``accepted-licenses`` list.  Instead of having to re-write the entire list.
- ``ingored-packages``: list of packages that are trusted but have invalid licenses.
-  ``python-version``
- ``skip-install``: to avoid installing the package if this action is used inside another one that already installs the package

Then I will add the install package step before checking the licenses.

Do you think I'm missing something? Comments?

RobPasMue: Well... It's confusing for me to have two arguments called ``accepted-licenses`` and ``additional-accepted-licenses``.

* The first ones (``accepted-licenses``) should be hard-coded in the action. In fact we should be reading them from a file (same way you execute a ``check-licenses.py`` file) so that it is more maintainable. We should keep track of them in a txt file or similar of the accepted licences by the PyAnsys repositories.
* The ``additional-accepted-licenses`` approach is good for those repos which have specific needs.
* The ``ignored-packages`` option should behave similarly to the previous two. We should have a list of generally accepted packages that have "wrong" licenses. And we should also allow users to provide additional packages (if needed)

Keep in mind that this action is oriented for PyAnsys repositories. There is no problem on tailoring it for our purpose - with these ``accepted-licenses.txt`` and ``ignored-packages.txt`` files, for example.

RobPasMue: Don't know if you got my point but basically I am proposing to:
* Delete ``accepted-licenses`` and ``ignored-packages`` options
* Only accept ``additional-accepted-licenses`` and ``additional-ignored-packages``
* Write to a TXT file those licenses accepted by default, and read them from there in the action
* Write to a TXT file those packages accepted by default, and read them from there in the action

GuillemBarroso: After an off-line conversation, we've decided to go for the txt approach.

Thus, the inputs of this actions are doing to be:
- ``python-version``
- ``skip-install``: to avoid installing the package if this action is used inside another one that already installs the package

Inside the action we're going to hard-code the most common licenses. Then, two txt files (``accepted-licenses.txt`` and ``ignored-packages.txt``) will be used in order to tweak which extra licenses are also accepted and which packages can be ignored, depending on the particularities of each repo.

GuillemBarroso: pinging @jorgepiloto and @MaxJPRey for further input on the final decision.

RobPasMue: My main point is that the default accepted licenses and ignored packages should already be defined inside this action (if possible, as TXT files). The way in which users provide additional licenses and ignored packages is up for discussion, either with an additional TXT file which they maintain in their own repos or directly in the YML file calling this action.

GuillemBarroso:
The final approach that I've followed is to only accept licenses appearing in ``accepted-licenses.txt`` and ignore packages in ``ignored-packages.txt``. No additional licenses or packages can be added.

By doing this we force all pyansys repos to follow the same licenses. If a particular project wants to add a new license or ignored packages, they can do so by proposing a PR adding them in the txt files.

check-licenses/action.yml
MaxJPRey: If you do so, just add a comment to explain why we need matplotlib and numpy.
@jorgepiloto Could numpy be enough?

GuillemBarroso: Maybe I'm wrong, but this was just to test the action. Now I'm going to add all the libraries for PyAnsys2023.1.1 and I will remove the matplotlib and numpy imports. It shouldn't be required anymore when calling the action from other repos.

RobPasMue: Yes, AFAIK, it was only for testing purposes.

GuillemBarroso:
I have tested this action with the pyansys/pyansys repo, see [PR](https://github.com/pyansys/pyansys/pull/97).

They're both passing, so I believe we should be good to go. Please, let me know if you think something in the action could be improved.

.github/workflows/ci_cd.yml
```diff
@@ -19,264 +19,23 @@ concurrency:
 
 jobs:
 
-  label-syncer:
-    name: Label syncer
+  debug:
+    name: "Debug"
     runs-on: ubuntu-latest
```
RobPasMue, line 24: Remember to revert the main workflow.

accepted-licenses.txt
```diff
@@ -0,0 +1,8 @@
+MIT
```
RobPasMue, line 1: I think this file should just be deleted... right? It is duplicated. The only one that should live on is the one inside the ``check-licenses`` folder

ignored-packages.txt
```diff
@@ -0,0 +1,2 @@
+matplotlib
+pypiwin32
```
RobPasMue, line 2: Same comment as for ``accepted-licenses.txt`` applies here.

check-licenses/action.yml
```diff
@@ -0,0 +1,68 @@
...
+    description: 'Whether to use Python cache or not. Requires a pyproject.toml file.'
+    required: false
+    default: true
+    type: boolean
+  skip-install:
+    description: 'Skip project installation.'
+    default: false
+    required: false
+    type: boolean
+
```
RobPasMue, line 20: I left you a comment on the ``pyansys/pyansys`` PR. Please have a look at it: https://github.com/pyansys/pyansys/pull/97#issuecomment-1328592298

We might need to provide an additional optional argument called ``target``. Plan on implementing it.

RobPasMue (changes requested):
I am missing documentation on how to use this action. Are you planning on adding it? I think it might be a good addition to your PR. Apart from my comments, LGTM @GuillemBarroso! Re-request review once the comments are implemented.

check-licenses/action.yml
```diff
@@ -0,0 +1,68 @@
...
+      if: inputs.skip-install == 'false'
+      shell: bash
+      run: python -m pip install .
+
+    - name: "Check licences of packages"
+      shell: bash
+      run: |
+        python -m pip install pip-licenses
+        wget https://raw.githubusercontent.com/pyansys/actions/f06b839725de78136d4f797b95d5c6458c689cd3/check-licenses/accepted-licenses.txt
+        wget https://raw.githubusercontent.com/pyansys/actions/f06b839725de78136d4f797b95d5c6458c689cd3/check-licenses/ignored-packages.txt
```
jorgepiloto, line 50: Note that these URLs need are currently pointing to the files in the branch you created, @GuillemBarroso. Once everything is ready to be merged, they should be changed assuming that the files are in `main`. This will make the actions to fail in this PR but pass once merged.

jorgepiloto: Also, could we isolate the download of the files in a separate step? Just to split de setup process and the final logic of the licensing checking.

jorgepiloto (changes requested):

check-licenses/action.yml
RobPasMue: Oh true! Yeah, those URLs have to be modified =)

And yeah, we might want to break down the entire step into more isolated ones. Good idea @jorgepiloto.

jorgepiloto (changes requested):

accepted-licenses.txt
GuillemBarroso: Yes, that's right.

ignored-packages.txt
GuillemBarroso: Done.

check-licenses/action.yml
GuillemBarroso: Done.

check-licenses/action.yml
GuillemBarroso: @jorgepiloto, yes. This applies to both the links in ``action.yml`` and the @feat/chech-licenses in ``ci_cd.yml``.

What's the proper way of doing that? I was thinking on first merging the PR (should pass since the branch will still be "alive"), and then proposing another PR changing the two files mentioned above to point to @main with the correct links.

Finally, I think the links in ``action.yml`` have to be updated every time that the files are modified. Is this correct? (For example, now the PR in pyanys/pyansys is failing because it does not download the updated version of the files).

check-licenses/accepted-licenses.txt
```diff
@@ -0,0 +1,8 @@
+MIT
```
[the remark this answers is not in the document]
GuillemBarroso: Done.

doc/source/check-licenses-action/index.rst
```diff
@@ -0,0 +1,44 @@
+Check licenses action
```
jorgepiloto, line 1: Let us rename this to `Licenses actions` to be consistent with the rest of the headers. Rename the folder as well.

doc/source/index.rst
```diff
@@ -7,5 +7,6 @@
    style-actions/index
    tests-actions/index
    doc-actions/index
+   check-licenses-action/index
```
jorgepiloto, line 10: See previous comment about renaming this folder to unify the directories names.

.github/workflows/ci_cd.yml
```diff
@@ -101,6 +100,15 @@ jobs:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           use-python-cache: false
 
+  check-licenses:
```
jorgepiloto, line 103: We can not add this action within the CI pipeline of this repository. The reason is that this is not a Python package. Even if we use the `skip-install`, we are not creating any Python virtual environment or publishing third party packages.

This is the reason of the failing CI.

check-licenses/accepted-licenses.txt
```diff
@@ -0,0 +1,10 @@
+Ansys
+Apache
+BSD
+GNU
+HPND
+ISC
+MIT
+MPL
+PSF
+
```
jorgepiloto, line 10:
```suggestion
```

jorgepiloto (changes requested):

doc/source/check-licenses-action/index.rst
```diff
@@ -1,4 +1,4 @@
-Check licenses action
+Licenses actions
 =====================
```
RobPasMue, line 2:
```suggestion
================
```

doc/source/index.rst
GuillemBarroso: I have renamed the doc section to Licenses actions. However, I left the action itself as check-licenses. Let me know if you're OK with that @jorgepiloto.

doc/source/check-licenses-action/index.rst
RobPasMue: Remember to align title underline to title length

doc/source/index.rst
RobPasMue: Yeah that sounds fine, don't worry. One thing is the action itself, something different are the docs. The action name should be as explicit and standalone as possible

doc/source/licenses-actions/index.rst
```diff
@@ -0,0 +1,43 @@
...
+================
+
+The ``pyansys/check-licenses`` action allows to verify that the project's dependencies
+only contain valid licenses.
+
+The valid licenses are defined in the ``accepted-licenses.txt``. Moreover, an additional txt
+file ``ignored-packages.txt`` contains a list of packages that are trusted but may not have a
+valid license associated.
+
+Use of the check-licenses action
```
jorgepiloto, line 11: We are using one section per action. In addition, the title of the section is the title of the action, see the rest of the documentation files. Name this section to something like:

```suggestion
Check licenses action
```

All the content related with this action should be described below this section. For the guides on how to update the TXT files, take advantage of a `.. note::` directive.

jorgepiloto (changes requested):

check-licenses/action.yml
```diff
@@ -0,0 +1,82 @@
...
+    - name: "Install Python library"
+      if: inputs.skip-install == 'false'
+      shell: bash
+      run: pip install ${{ env.install_target }}
+
+    - name: "Download txt files so they are found from other repos"
+      shell: bash
+      run: |
+        wget https://raw.githubusercontent.com/pyansys/actions/feat/check-licenses/check-licenses/accepted-licenses.txt
+        wget https://raw.githubusercontent.com/pyansys/actions/feat/check-licenses/check-licenses/ignored-packages.txt
```
RobPasMue, line 59: Are you sure that the moment we merge this PR (and the branch gets deleted), those links will still work? I'd say they won't... 
```suggestion
    - name: "Download txt files so they are found from other repos"
      shell: bash
      run: |
        wget https://raw.githubusercontent.com/pyansys/actions/main/check-licenses/check-licenses/accepted-licenses.txt
        wget https://raw.githubusercontent.com/pyansys/actions/main/check-licenses/ignored-packages.txt
```

You'll probably have to merge this PR with a failing status due to this breaking change. But once it's on main, it should work.

In fact, this link will have to be edited every time we do a release so that it points to whatever release branch we have... It ain't gonna be easy to maintain... We should think something... 

What about using this action to do that same operation? https://github.com/marketplace/actions/checkout-files

GuillemBarroso: I have a question here that I was discussing with Maxime on Friday.

Won't it work if we leave this as feat/check-licenses for now? It will still point to the feat branch, but this branch will still be available until it gets removed right? I was thinking about first merging and then opening another PR to change from feat/check-licenses to main. Am I missing something?

I really like the approach you suggested with the checkout files action. I am going to implement it now.

RobPasMue: Sure, we can do it in 2 PRs =) there is no issue in that. I was just trying to optimize. But at least we will get our green ticks dopamine 😄 

And yeah, let's try to use the action. I think it'll be more clear for everybody to understand what this step is doing. I know what you're doing, and probably anybody that understand what ``wget`` is knows what you're doing... but just in case =)

doc/source/licenses-actions/index.rst
GuillemBarroso: Done.

check-licenses/action.yml
GuillemBarroso: For now, I'll leave it like this since I want to test the action from the PR in ``pyansys/pyansys``. After checking that everything works, I'll modify all the references to @feat/check-licenses to @main.

doc/source/licenses-actions/index.rst
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - name: "PyAnsys check_licenses action"
+          uses: pyansys/actions/check-licenses@main
+          with:
+            python-version: ${{ env.MAIN_PYTHON_VERSION }}
+
+.. note::
+
+  If a certain repository requires a different license or packages
+  that is not included in the original list, a PR can be proposed to ``pyansys/actions`` in order to
```
RobPasMue, line 38:
```suggestion
  that is not included in the original list, a PR can be proposed at ``pyansys/actions`` to
```

check-licenses/action.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    - name: "Check licences of packages"
+      shell: bash
+      run: |
+        python -m pip install pip-licenses
+        mapfile licenses_from_txt < accepted-licenses.txt
+        echo "Accept licences if they contain:"
+        accepted_licenses=''
+        for license in ${licenses_from_txt[*]}; do echo $license && accepted_licenses+="$license\|"; done
+        accepted_licenses=${accepted_licenses::-2}
```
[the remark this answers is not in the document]
GuillemBarroso: Nice, thanks for the suggestion.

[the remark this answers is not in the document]
GuillemBarroso: hm.. wait. Beacuse ``licenses_from_txt`` is an array... it's not a string with spaces between licenses. Will this work in this case? Same for ``ignored_packages_from_txt``.

check-licenses/action.yml
GuillemBarroso: @RobPasMue. I have implemented the [action](https://github.com/marketplace/actions/checkout-files) following the documentation. However, when using the ``check-licenses`` action in the [pyansys/pyansys PR](https://github.com/pyansys/pyansys/pull/97) it looks like it is trying to find the files in the ``pyansys/pyansys`` repo, see url that is trying to access:

image

I don't see an option to define the repo in which the files are located... Is it possible that this action is only capable of checking out files that are in the same repo where the action is called?

jorgepiloto: I think there shouldn't be a problem to define these files in the `main` branch, even if those point from a release branch. They are expected to be quite stable.

RobPasMue: Ha true... it looks like this action is only valid for files in the same repo... sorry for that =( you might need to revert to the ``wget`` mechanism

RobPasMue: Summarizing:
* Let's revert to the ``wget`` mechanism
* Point directly to its expected location in the ``main`` branch
* Merge whenever you feel it's ready =)

RobPasMue:
Should we merge it then? Or are we waiting for something else? Let me do a final review just in case

doc/source/licenses-actions/index.rst
```diff
@@ -0,0 +1,43 @@
...
+---------------------
+
+The ``pyansys/check-licenses`` action allows to verify that the project's dependencies
+only contain valid licenses.
+
+The valid licenses are defined in the ``accepted-licenses.txt``. Moreover, an additional TXT
+file ``ignored-packages.txt`` contains a list of packages that are trusted but may not have a
+valid license associated.
+
+This acction accepts the following inputs.
```
RobPasMue, line 16:
```suggestion
This action accepts the following inputs.
```

doc/source/licenses-actions/index.rst
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: "PyAnsys check_licenses action"
+          uses: pyansys/actions/check-licenses@main
+          with:
+            python-version: ${{ env.MAIN_PYTHON_VERSION }}
+
+.. note::
+
+  If a certain repository requires a different license or packages
```
RobPasMue, line 37:
```suggestion
  If a certain repository requires a different license or package
```

RobPasMue (approved):
Some comments, otherwise LGTM! Thanks @GuillemBarroso for all the iterations!

jorgepiloto (changes requested):

check-licenses/action.yml
```diff
@@ -0,0 +1,82 @@
...
+    - name: "Install Python library"
+      if: inputs.skip-install == 'false'
+      shell: bash
+      run: pip install ${{ env.install_target }}
+
+    - name: "Checkout TXT files"
+      uses: Bhacaz/checkout-files@v2
+      with:
+        files: check-licenses/accepted-licenses.txt check-licenses/ignored-packages.txt
+        branch: feat/check-licenses
```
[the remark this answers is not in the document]
GuillemBarroso: No probs! Done.

GuillemBarroso:
I've checked that the action works properly when calling it from ``pyansys/pyansys``. 

I will now change all references to feat/check-licenses to main and it'll be ready to merge.

RobPasMue (approved):
LGTM! Let's go!

jorgepiloto (approved):

PR #87, opened 2022-11-28
Rename namespace for argument

Revathyvenugopal162:
Namespace object has no attribute 'announcement_file', which will create error. Resolves the above.


jorgepiloto (approved):
Great catch, @Revathyvenugopal162. Feel free to merge 🚀

PR #88, opened 2022-11-28
FIX: ignore missing announcement.html

jorgepiloto:
The "version_mapper.py" script gets downloaded from "main" branch. The latest changes introduced in #84 were being used, thus, by the "v1" branch.

This changes check whether the announcement file exists or not. If it is not present, the script terminates and forgets about the announcement.


GuillemBarroso (approved):
LGTM

RobPasMue (approved):
Yep, good call @jorgepiloto

MaxJPRey (commented):
LGTM.

PR #92, opened 2022-11-29
FIX: recursively add announcement.html to the different folders

jorgepiloto:
Last additions to #90. The action will include now the "announcement.html" rendered file within each key subdir of the outdated release folders.

### DIFF
--- a/scripts/version_mapper.py
+++ b/scripts/version_mapper.py
@@ -105,13 +105,16 @@ def update_switch_version_file(
     # Include the announcement in all available release folders. Note that
     # these are still accessible even if they are not included in the dropdown.
     old_release_folders = []
-    for path in Path("release").iterdir():
-        if path.is_dir() and path.name != latest_stable_version:
+    for path in Path("release").glob("**/*"):
+        has_invalid_name = path.name == latest_stable_version and path.name.startswith(
+            "_"
+        )
+        if path.is_dir() and not has_invalid_name:
             old_release_folders.append(path)
 
-    for release_folder in old_release_folders:
+    for folder in old_release_folders:
         # Create an 'announcement.html' file within each one of the old versions
-        announcement_file = release_folder / "announcement.html"
+        announcement_file = folder / "announcement.html"
         with open(announcement_file, "w") as file:
             print(f"Writing content to {announcement_file}.")
             file.write(announcement_content)


scripts/version_mapper.py
```diff
@@ -105,13 +105,14 @@ def update_switch_version_file(
     # Include the announcement in all available release folders. Note that
     # these are still accessible even if they are not included in the dropdown.
     old_release_folders = []
-    for path in Path("release").iterdir():
-        if path.is_dir() and path.name != latest_stable_version:
+    ignore_folders = ["_static", "_sources", latest_stable_version]
```
RobPasMue, line 108:
```suggestion
    ignore_folders = ["_static", "_sources", latest_stable_version]
```

I think you might want to ignore any folder that starts with an underscore... Those two are the typical ones but there are some docs that have more:
image

RobPasMue (commented):
If I understand correctly this PR... the announcement will appear in any HTML page of previously released docs? That's why you are looping through all the existing folders?

jorgepiloto:
> If I understand correctly this PR... the announcement will appear in any HTML page of previously released docs? That's why you are looping through all the existing folders?

Exactly. The algorithm works as follows:

1. With the future ansys-sphinx-theme==0.7.3, all pages will look for an "announcement.html" next to them.
2. If not found, they will ignore it. If found, the announcement gets displayed as a banner in the top.
3. The announcement is updated according to the latest stable version. A file named "announcement.html" living in the "release/" folder is used as template. It contains something like:

```text
<p>This version is outdated. Please visit <a>{version}</a>.<p>
```


4. This file gets copied to the "outdated" documentation folders and their subdirectories containing the HTML files to be displayed.
5. If any new release, everything updates automatically.

RobPasMue:
> > If I understand correctly this PR... the announcement will appear in any HTML page of previously released docs? That's why you are looping through all the existing folders?
> 
> Exactly. The algorithm works as follows:
> 
> 1. With the future ansys-sphinx-theme==0.7.3, all pages will look for an "announcement.html" next to them.
> 2. If not found, they will ignore it. If found, the announcement gets displayed as a banner in the top.
> 3. The announcement is updated according to the latest stable version. A file named "announcement.html" living in the "release/" folder is used as template. It contains something like:
> 
> ```
> <p>This version is outdated. Please visit <a>{version}</a>.<p>
> ```
> 
> 4. This file gets copied to the "outdated" documentation folders and their subdirectories containing the HTML files to be displayed.
> 5. If any new release, everything updates automatically.

Awesome @jorgepiloto! Thanks for the thorough explanation!

RobPasMue (approved):
LGTM!

PR #93, opened 2022-11-29
FIX: valid name logic

jorgepiloto:
I introduced a bug in previous pull-request. The logic required an "or" instead of an "and". Nevertheless, this changes may be more readable than using negations.


MaxJPRey:
Good catch @jorgepiloto .

PR #94, opened 2022-11-29
FIX: recursively add announcement.html

jorgepiloto:
This pull-request fixes one last bug regarding the announcment. The new code guarantees that not only private folders are excluded but also their sub-directories.

Also, the whole stable release folder gets ignored by default.

### DIFF
--- a/scripts/version_mapper.py
+++ b/scripts/version_mapper.py
@@ -105,15 +105,23 @@ def update_switch_version_file(
     # Include the announcement in all available release folders. Note that
     # these are still accessible even if they are not included in the dropdown.
     old_release_folders = []
-    for path in Path("release").glob("**/*"):
-        path_has_valid_name = (
-            path.name != latest_stable_version and not path.name.startswith("_")
-        )
-        if path.is_dir() and path_has_valid_name:
+    for path in Path("release").glob(f"**/[!{latest_stable_version}]*"):
+
+        # Skip the path if it not a directory
+        if not path.is_dir():
+            continue
+
+        # Append the folder unless it is contained within a private directory
+        add_path = True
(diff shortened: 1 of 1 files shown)


scripts/version_mapper.py
```diff
@@ -105,15 +105,23 @@ def update_switch_version_file(
...
+            continue
+
+        # Append the folder unless it is contained within a private directory
+        add_path = True
+        for dirname in str(path.absolute()).split("/"):
+            if dirname.startswith("_"):
+                add_path = False
+                break
+        if add_path:
             old_release_folders.append(path)
```
RobPasMue, line 121: Algorithm logic:
* You get all the elements inside the ``release`` folder except the ones belonging to the latest stable release folder and loop them
  * If your element is a directory, you proceed, else you continue the loop
  * You get the absolute path of your directory (i.e. MYDISK/folder/just/created/release/0.1) and split it by "/" (i.e. ``["MYDISK", "folder", "just", "created", "release", "0.1"]``)
  * And now... You are checking if any of the elements of that list start with ``_``?

I think the last point is wrong... You should only check those inside each of the release folders. Am I missing something?

RobPasMue: If you don't understand my comment let me know and we can talk about it.

jorgepiloto: I see your point. Let me try to force relative path to the release folder instead of asking for the absolute one.

RobPasMue: Perfect! Thanks for addressing this @jorgepiloto

RobPasMue (approved):

Revathyvenugopal162:
make sure that we put one announcement file inside the release/0.1, before checking for directory, near to index.html, if it already added perfect.

PR #108, opened 2022-12-01
MAINT: update cname and versions.json content

### LINKED ISSUE #107: Update conf.py and ci_cd.yml to use DOCUMENTATION_CNAME env variable
🐞 **Problem**
Ideally, we should define key variables in one single place. As per @RobPasMue suggestion some time ago, let us use a variable named `DOCUMENTATION_CNAME` in our `ci_cd.yml` files to specify the value for the canonical name pointing to the documentation website. This value should be used for the `version.json` file too.

jorgepiloto:
Fix #107 by:

- Adding a `DOUMENTATION_CNAME` within the `ci_cd.yaml` file.
- Reading its value from the `conf.py` file.
- Using it in the `versions` vairable.
- Deprecating the `get_version_match` function in favor of the one implemented in the theme.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   MAIN_PYTHON_VERSION: '3.10'
-  CNAME: 'actions.docs.pyansys.com'
+  DOCUMENTATION_CNAME: 'actions.docs.pyansys.com'
   test-library-name: 'pyansys-actions'
 
 concurrency:
@@ -159,7 +159,7 @@ jobs:
         if: github.event_name == 'push'
         uses: pyansys/actions/doc-deploy-to-repo@main
         with:
-          cname: ${{ env.CNAME }}
+          cname: ${{ env.DOCUMENTATION_CNAME }}
           repository: "pyansys/actions"
           bot-id: ${{ secrets.BOT_APPLICATION_ID }}
           bot-token: ${{ secrets.BOT_APPLICATION_PRIVATE_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   MAIN_PYTHON_VERSION: '3.10'
-  CNAME: 'actions.docs.pyansys.com'
+  DOCUMENTATION_CNAME: 'actions.docs.pyansys.com'
```
RobPasMue, line 13: Understood. Only in this ``ci_cd.yml`` file 😄

doc/source/conf.py
```diff
@@ -18,29 +20,17 @@
...
 html_favicon = ansys_favicon
 
 # Specify the location of your GitHub repo
-version_mapper = (
-    "https://raw.githubusercontent.com/pyansys/actions/gh-pages/release/versions.json"
-)
 html_theme_options = {
     "github_url": "https://github.com/pyansys/actions",
-    "show_prev_next": False,
+    "show_prev_next": True,
```
RobPasMue, line 31: Interesting... I thought we weren't using this feature. In any case, if you want to set it to ``True`` there is no need to do that. It is set by default to ``True``. See https://pydata-sphinx-theme.readthedocs.io/en/v0.9.0/user_guide/configuring.html?highlight=show_prev_next#hiding-the-previous-and-next-buttons

RobPasMue (commented):
It's looking good =)

RobPasMue (approved):

doc/source/conf.py
jorgepiloto: Oh! I see... Then I guess we can get rid of this line then.

PR #109, opened 2022-12-02
DOC: add more examples and automate their rendering

### LINKED ISSUE #82: Add more examples on how to use the actions
🐞 **Problem**
So far, each one of the actions contains a small gist explaining how to use it. These examples are quite trivial, containing only one step.

However, most of the projects within PyAnsys require to install a server to access all the logic through the Python library. How to combine this step with the pyansys/actions is not explained, even though it is a simple task.

💡 **Idea**
~We could create a `Examples` section showing more complex jobs or add more examples next to each one of the actions.~

Let us collect all the examples under each action, as similar to what we do with docstring examples. I am proposing this because some of the examples may not show all the possible inputs an action can take.

jorgepiloto:
Fix #82 by adding more examples and using Jinja to create the dropdowns and fixing the version. In addition, the accepted licenses and ignored packages are dynamically loaded too, see:

Screenshot 2022-12-06 at 13-37-29 Tests actions — PyAnsys Actions 2 0 dev0 documentation

Screenshot 2022-12-06 at 13-38-34 Licenses actions — PyAnsys Actions 2 0 dev0 documentation

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 88
 ignore = E501
-extend-ignore = E203
+extend-ignore = E203, W503
--- a/.gitignore
+++ b/.gitignore
@@ -162,5 +162,5 @@ cython_debug/
 
 # End of https://www.toptal.com/developers/gitignore/api/python
 
-# Ignore baked output cookies from debugging
-output/
+# Ignore rendered examples
+*-rendered-example.yml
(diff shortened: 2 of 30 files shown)


RobPasMue:
I'll try to review at some point today =)

jorgepiloto:
Merging this so we do not get blocked from implementing the latest changes in other projects. If some issues appear, we can do a patch release and keep updating the v2 tag.

PR #120, opened 2022-12-19
Adding skip artifact download

germa89:
I'm working in a workflow which downloads the already build documentation from another workflow:

https://github.com/pyansys/pymapdl/pull/1707

Because I download the artifact from other workflow, the default ``actions/download-artifact`` does not work. Hence I would like to skip it in the current pyansys action.

Other option is for me to just copy this workflow but given the complexity of the latter steps, I would like to use this action as much as I can.

### DIFF
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -16,7 +16,7 @@ inputs:
     required: true
     type: string
   use-downloaded-artifact:
-    description: "Skip downloading the artifact."
+    description: "Skip downloading the artifact. Uses the file <doc-artifact-name> at the working directory."
     required: false
     type: bool
     default: false
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -15,6 +15,11 @@ inputs:
     description: "Required token for documentation deployment."
     required: true
     type: string
+  use-downloaded-artifact:
+    description: "Skip downloading the artifact."
+    required: false
+    type: bool
+    default: false
 
 runs:
   using: "composite"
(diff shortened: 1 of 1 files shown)


germa89:
There is one concern:

The ``doc-artifact-name`` does not include file extension. I believe the ``actions/download-artifact`` does unzip the file. Then we need to add the ``zip`` extension at the ``run`` step and the ``unzip`` command.

Revathyvenugopal162 (approved):
LGTM. Thanks @germa89

doc-deploy-dev/action.yml
```diff
@@ -15,6 +15,11 @@ inputs:
     description: "Required token for documentation deployment."
     required: true
     type: string
+  use-downloaded-artifact:
```
jorgepiloto, line 18: May we renamed this input variable as `skip_artifact_download`? This is similar to the `skip_installation` input variable that we have in other actions.

jorgepiloto (changes requested):
Thanks for opening this, @germa89. I just left one comment about variable naming.

Also, do you think this should be implemented in the `doc-deploy-stable` action?

doc-deploy-dev/action.yml
germa89: I agree with this. I will also check if it is a good addition to the stable too.

I just hope I'm not customizing a general action such as this!

jorgepiloto:
From the changes you made in https://github.com/pyansys/pymapdl/pull/1707, I do not see any reason behind implementing this. 

The pyansys/actions allow to download the artifact based on their name. If you need to point to a particular YML file, let us add that as an input rather than skipping the artifact download, @germa89.

germa89:
> From the changes you made in [pyansys/pymapdl#1707](https://github.com/pyansys/pymapdl/pull/1707), I do not see any reason behind implementing this.
> 
> The pyansys/actions allow to download the artifact based on their name. If you need to point to a particular YML file, let us add that as an input rather than skipping the artifact download, @germa89.

I think that is even better. 

I will change this accordingly.

doc-deploy-dev/action.yml
germa89: Resolving because we are changing this whole thing

germa89:
Closing because it wont be implemented.

PR #121, opened 2022-12-19
Allowing download artifacts from other workflows

germa89:
As the title.

We should probably give many options for filtering the artifacts.

I'm using the default ``actions/download-artifact`` action unless one of the arguments is used, but I'm happy to drop it.

### DIFF
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -20,6 +20,34 @@ inputs:
     required: false
     default: ""
     type: string
+  repo:
+    description: Repository name with owner (like actions/checkout)
+    required: false
+    default: ${{ github.repository }}
+  pr:
+    description: Pull request number
+    required: false
+  commit:
+    description: Commit hash
+    required: false
+  branch:
+    description: Branch name
+    required: false
+  event:
+    description: Event type
+    required: false
+  run_id:
+    description: Workflow run id
+    required: false
+  run_number:
+    description: Workflow run number
+    required: false
+  name:
+    description: Artifact name (download all artifacts if not specified)
+    required: false
+  dry_run:
+    description: Check the existence of artifact(s) without downloading
+    required: false
 
 runs:
   using: "composite"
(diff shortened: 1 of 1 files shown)


doc-deploy-dev/action.yml
```diff
@@ -15,6 +15,39 @@ inputs:
...
+    required: false
+  run_number:
+    description: Workflow run number
+    required: false
+  name:
+    description: Artifact name (download all artifacts if not specified)
+    required: false
+  dry_run:
+    description: Check the existence of artifact(s) without downloading
+    required: false
```
germa89, line 50: Copy/pasted from the ``https://github.com/pyansys/actions/pull/121`` args

doc-deploy-dev/action.yml
```diff
@@ -34,11 +67,29 @@ runs:
       shell: bash
       run: rm -rf dev/ && mkdir dev/
 
-    - name: "Download the HTML documentation artifact"
+    - name: "Download the HTML documentation artifact from the same workflow"
       uses: actions/download-artifact@v3
+      if: ${{ inputs.workflow == "" }}
+      with:
+        name: ${{ inputs.doc-artifact-name }}
+        path: dev
```
germa89, line 75: This branching can be deleted if we are happy with using ``dawidd6/action-download-artifact@v2`` for everything... I think we can be confident enough we should not mess it up.

germa89:
Interesting read:

https://www.legitsecurity.com/blog/github-actions-that-open-the-door-to-cicd-pipeline-attacks

germa89:
https://github.com/dawidd6/action-download-artifact

germa89:
Replace #120

doc-deploy-dev/action.yml
jorgepiloto, line 50: Please, make sure that each input has the following fields:

* Description: "...." (Use single quotes
* Default (if required): default-value
* Required: true/false
* Type: [boolean, string...]

jorgepiloto (changes requested):

doc-deploy-dev/action.yml
germa89: @jorgepiloto  What about this?

doc-deploy-dev/action.yml
germa89: This is now resolved.

doc-deploy-dev/action.yml
germa89: I'm just going to use that action for everything.

doc-deploy-dev/action.yml
```diff
@@ -15,6 +15,48 @@ inputs:
...
+  run_number:
+    description: 'Workflow run number'
+    required: false
+    type: string
+  name:
+    description: 'Artifact name (download all artifacts if not specified)'
+    required: false
+    type: string
+  dry_run:
+    description: 'Check the existence of artifact(s) without downloading'
```
MaxJPRey, line 57: This could be very useful. 👍

doc-deploy-dev/action.yml
```diff
@@ -15,6 +15,48 @@ inputs:
     description: "Required token for documentation deployment."
     required: true
     type: string
+  workflow:
+    description: 'Workflow name from where the artifacts are downloaded. By default, the same running workflow'
```
MaxJPRey, line 19:
```suggestion
    description: 'Workflow name from where the artifacts are downloaded. By default, the same running workflow.'
```

MaxJPRey (approved):
I needed something very similar few days ago. Thanks @germa89 for implementing it.

germa89:
> I needed something very similar few days ago. Thanks @germa89 for implementing it.

It is my pleasure @MaxJPRey !! :)

jorgepiloto (approved):
Awesome, @germa89!

PR #136, opened 2023-01-16
Adding 'setup-python' and 'download-artifacts' options

germa89:
As the title.

Now if you want to use artifacts different than the defaults, you need to download them by yourself and locate them in the right folder. Then execute the appropriate action.

### DIFF
--- a/_release-pypi/action.yml
+++ b/_release-pypi/action.yml
@@ -22,12 +22,24 @@ inputs:
     description: 'Desired Python version for using twine'
     required: true
     type: string
+  setup-python:
+    description: 'Whether install python or not.'
+    required: false
+    type: bool
+    default: true
+  download-artifacts:
+    description: 'Whether or not download the artifacts. By default download an artifact named "``library-name``-artifacts"'
+    required: false
+    type: bool
+    default: true
+
 
 runs:
   using: "composite"
   steps:
   
     - name: "Set up Python"
+      if: inputs.setup-python == 'true'
       uses: pyansys/actions/_setup-python@main
       with:
         python-version: ${{ inputs.python-version }}
@@ -40,6 +52,7 @@ runs:
         python -m pip install twine
 
     - name: "Download the library artifacts from build-library step"
+      if: inputs.download-artifacts == 'true'
       uses: actions/download-artifact@v3
       with:
         name: ${{ inputs.library-name }}-artifacts
--- a/_release-pypi/action.yml
+++ b/_release-pypi/action.yml
@@ -22,11 +22,6 @@ inputs:
     description: 'Desired Python version for using twine'
     required: true
     type: string
-  setup-python:
-    description: 'Whether install python or not.'
-    required: false
-    type: bool
-    default: true
   download-artifacts:
     description: 'Whether or not download the artifacts. By default download an artifact named "``library-name``-artifacts"'
     required: false
@@ -39,7 +34,6 @@ runs:
   steps:
   
     - name: "Set up Python"
-      if: inputs.setup-python == 'true'
       uses: pyansys/actions/_setup-python@main
       with:
         python-version: ${{ inputs.python-version }}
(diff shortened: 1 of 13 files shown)


germa89:
Some actions has ``skip-install`` which depending on the action can do just the ``pip install`` (i.e. ``doc-build`` action) or also the checkout, python install and project install (i.e. ``check-license`` action).

We should probably be consistent with its usage.

germa89:
The ``release-to-PyPI`` and ``release-to-Github`` are not consistent.

The PyPI action download the artifacts (all) to a folder called ``${{ inputs.library-name}}-artifacts`` but then, the ``release-to-github`` action check the folder ``other-artifacts``.

RobPasMue:
This is in the lines of #140.. we should figure out the strategy to follow first before merging anything. Pinging @jorgepiloto as well for acknowledgement.

germa89:
> This is in the lines of #140.. we should figure out the strategy to follow first before merging anything. Pinging @jorgepiloto as well for acknowledgement.

I think this PR is fine for now. We can work out the path structure later.

doc-build/action.yml
```diff
@@ -38,6 +38,7 @@ inputs:
       required: false
       type: boolean
 
+
```
RobPasMue, line 41:
```suggestion
```

code-style/action.yml
```diff
@@ -18,6 +18,7 @@ inputs:
       required: false
       type: boolean
 
+
```
RobPasMue, line 21:
```suggestion
```

doc-deploy-stable/action.yml
```diff
@@ -86,8 +91,8 @@ runs:
         echo "VERSION=$major.$minor" >> $GITHUB_ENV
 
     - name: "Download the HTML documentation artifact in release/${{ github.ref_name }}"
-      if: env.ACCEPTED_FORMAT == 'true'
-      uses: actions/download-artifact@v3
+      if: ${{ env.ACCEPTED_FORMAT == 'true' && inputs.download-artifacts }}
```
RobPasMue, line 94: Not sure if this will work. Pinging @jorgepiloto

_release-pypi/action.yml
```diff
@@ -22,6 +22,12 @@ inputs:
     description: 'Desired Python version for using twine'
     required: true
     type: string
+  download-artifacts:
+    description: 'Whether or not download the artifacts. By default download an artifact named "``library-name``-artifacts"'
+    required: false
```
RobPasMue, line 27:
```suggestion
    required: true
```

_release-pypi/action.yml
```diff
@@ -22,6 +22,12 @@ inputs:
...
+    type: bool
+    default: true
```
RobPasMue, line 29:
```suggestion
```

_release-pypi/action.yml
```diff
@@ -22,6 +22,12 @@ inputs:
     description: 'Desired Python version for using twine'
     required: true
     type: string
+  download-artifacts:
```
RobPasMue, line 25: For "private" actions (i.e. starting with "_"), it is better not to have optional args.

RobPasMue (commented):
I'm fine with it (for now)... but only because it is needed for your release @germa89.

Honestly, I am looking forward to #141 ... When the time comes, this might be reverted. Just letting you know.

I'll leave it for @jorgepiloto to take the final call.

_release-pypi/action.yml
germa89: Fair enough

doc-deploy-stable/action.yml
germa89: Why? I think it should (if you mean the ``if`` syntax)

https://github.com/orgs/community/discussions/25641#discussioncomment-3248571

germa89: You are right...

It should be without the ``${{ }}``.

germa89:
Where are we on this??

jorgepiloto:
Discarding these changes based on the strict philosophy of the actions.

PR #144, opened 2023-01-18
FIX: indicate whether to compress/decompress artifacts

### LINKED ISSUE #143: Add decompress_artifact option
🐞 **Problem**
Some projects like PyAEDT and PyFluent generate heavy documentation due to the amount of examples. This situation forces them to zip the HTML documentation. If these artifacts are downloaded a double-zipped file is obtained.

💡 **Idea**
Let us add a `decompress_artifact` option to ensure that projects are able to indicate if an action should decompress an artifact.

jorgepiloto:
Fixes #143 by allowing users to specify if documentation artifacts should be decompressed or not.

### DIFF
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -7,6 +7,11 @@ inputs:
     required: false
     default: 'documentation-html'
     type: string
+  decompress_artifact:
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):
LGTM @jorgepiloto!

PR #146, opened 2023-01-23
Skip json build for docs build.

Revathyvenugopal162:
Fix #145

Issue #134, opened 2023-01-10
Migrate all versioning logic to a Python script


jorgepiloto:
🐞 **Problem**
The [logic behind the doc-deploy-stable](https://github.com/pyansys/actions/blob/main/doc-deploy-stable/action.yml#L53-L86) for multi-version mixes between Shell and Python script.

I think it would make more sense to delegate all the logic in a Python script that:

* Performs the proper tag check
* Crops the version to X.Y numbers
* Applies all the version update logic

Note that this change does not affect the logic itself, so no breaking changes are introduced when implementing this.

RobPasMue:
I totally agree yeah... better all this scripting to be in one place rather than multiple...

MaxJPRey:
+1 for moving the logic in a Python script.

jorgepiloto:
Closing this after #163. Everything is now in the form of a bash workflow.

PR #147, opened 2023-01-23
FIX: shell script syntax

jorgepiloto:
Fixes the issues found in https://github.com/ansys/ansys-sphinx-theme/actions/runs/3987699947/jobs/6837892451. A continuation of the work made in #146.

### DIFF
--- a/doc-build/action.yml
+++ b/doc-build/action.yml
@@ -110,7 +110,7 @@ runs:
       run: |
         make -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
         make -C doc pdf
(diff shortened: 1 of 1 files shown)


Revathyvenugopal162 (approved):

MaxJPRey (approved):

RobPasMue (commented):
Thanks for taking care of this @jorgepiloto - I must have overseen it

jorgepiloto:
No worries, @RobPasMue. As I was discussing with @Revathyvenugopal162, the syntax for GitHub actions conditionals is confusing. On top of that, we need to switch it when working with shell script.

This is the reason behind my proposal in #134, so everything gets collected under Python.

PR #150, opened 2023-01-24
FIX: download artifacts from desired workflow

jorgepiloto:
Fixes the recent issues in:

* Pyfluent visualization: https://github.com/pyansys/pyfluent-visualization/actions/runs/3992912227/jobs/6849174814
* PyAEDT: https://github.com/pyansys/pyaedt/actions/runs/3992907653
* Other internal repositories that just upgraded to pyansys/actions@v3.

I feel the conflict is being caused because the value for the `workflow` is not behaving as expected. Hence, I went with this "duplicated" logic. This is the original logic from pyansys/actions@v2.0.7, which has been proven to work.

I do not like to keep patching stuff in here but I did not see other option for the moment. Once we agree on https://github.com/pyansys/actions/discussions/141, everything will be unified for @v4.

### DIFF
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -23,6 +23,7 @@ inputs:
   workflow:
     description: 'Workflow name from where the artifacts are downloaded. By default, the current running workflow.'
     required: false
+    default: ''
     type: string
   checkout:
     description: 'If ``True``, the project is checked out in the ``gh-pages`` branch. If ``False``, users are the ones in charge of performing the checkout step.'
@@ -51,13 +52,22 @@ runs:
       run: |
         rm -rf dev/ && mkdir dev/
 
-    - name: "Download the HTML documentation artifact"
+    - name: "Download the HTML documentation artifact from desired workflow"
       uses: dawidd6/action-download-artifact@v2
+      if: inputs.workflow != ''
       with:
         github_token: ${{ inputs.token }}
         name: ${{ inputs.doc-artifact-name }}
         path: dev
         workflow: ${{ inputs.workflow }}
+
+    - name: "Download the HTML documentation artifact from current workflow"
+      uses: dawidd6/action-download-artifact@v2
+      if: inputs.workflow != ''
+      with:
+        github_token: ${{ inputs.token }}
+        name: ${{ inputs.doc-artifact-name }}
+        path: dev
 
     - name: "Decompress artifact content if required"
       shell: bash
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -25,6 +25,11 @@ inputs:
     required: false
     default: ''
     type: string
+  workflow-conclusion:
+    description: 'Workflow conclusion (options: "failure", "success", "neutral", "cancelled", "skipped", "timed_out", "action_required") or status (options: "completed", "in_progress", "queued") to search for. Use the empty string ("") to ignore status or conclusion in the search.'
+    required: false
+    default: ''
+    type: string
   checkout:
     description: 'If ``True``, the project is checked out in the ``gh-pages`` branch. If ``False``, users are the ones in charge of performing the checkout step.'
     required: false
@@ -60,6 +65,7 @@ runs:
         name: ${{ inputs.doc-artifact-name }}
         path: dev
         workflow: ${{ inputs.workflow }}
+        workflow_conclusion: ${{ inputs.workflow-conclusion }}
 
     - name: "Download the HTML documentation artifact from current workflow"
       uses: dawidd6/action-download-artifact@v2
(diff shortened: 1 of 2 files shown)


doc-deploy-dev/action.yml
```diff
@@ -50,14 +51,23 @@ runs:
...
         workflow: ${{ inputs.workflow }}
 
+    - name: "Download the HTML documentation artifact from current workflow"
+      uses: dawidd6/action-download-artifact@v2
+      if: inputs.workflow != ''
+      with:
+        github_token: ${{ inputs.token }}
+        name: ${{ inputs.doc-artifact-name }}
+        path: dev
+
```
RobPasMue, line 70:
```suggestion

    - name: "Download the HTML documentation artifact from current workflow"
      uses: dawidd6/action-download-artifact@v2
      if: inputs.workflow == ''
      with:
        github_token: ${{ inputs.token }}
        name: ${{ inputs.doc-artifact-name }}
        path: dev

```
I am guessing this should be like this... but please correct me if I'm wrong

RobPasMue (approved):

doc-deploy-dev/action.yml
RobPasMue: I changed the ``!=`` for ``==``

RobPasMue:
> I feel the conflict is being caused because the value for the `workflow` is not behaving as expected. Hence, I went with this "duplicated" logic. This is the original logic from pyansys/actions@v2.0.7, which has been proven to work.

I think PyGeometry uses this actions version and we weren't really seeing any issues... let me double check it.

> I do not like to keep patching stuff in here but I did not see other option for the moment. Once we agree on #141, everything will be unified for @v4.

I feel you... I don't like patching stuff either. Let's try to get some consensus on #141

RobPasMue:
Yes... see https://github.com/pyansys/pygeometry/actions/runs/3993924998/jobs/6851241074

PyGeometry isn't affected by this issue. We might want to look into whether this is really what's causing it or not.

doc-deploy-dev/action.yml
MaxJPRey: Yes, thanks @RobPasMue

MaxJPRey (approved):
LGTM.

germa89:
@jorgepiloto @RobPasMue @MaxJPRey

## What is happening?

The reason why the artifact is not being found is because the workflow conclusion. 

If we examine the workflow log:
https://github.com/pyansys/pyfluent-visualization/actions/runs/3992912227/jobs/6849174814#step:11:480
```console
Run dawidd6/action-download-artifact@v2
==> Repository: pyansys/pyfluent-visualization
==> Artifact name: HTML-Documentation-tag-v23.1.0
==> Local path: dev
==> Workflow name: 27198130
==> Workflow conclusion: success
==> (found) Run ID: 3964634582
==> (found) Run date: 2023-01-20T04:05:04Z
==> (not found) Artifact: HTML-Documentation-tag-v23.1.0
==> Found the following artifacts instead:
Error: no artifacts found
```
It says the ``Workflow conclusion: success``, and then it picks the run ``3964634582``, which is **Nightly Documentation Build #239**:

https://github.com/pyansys/pyfluent-visualization/actions/runs/3964634582

And it is the **last successful** run:

image

See: https://github.com/pyansys/pyfluent-visualization/actions/workflows/nightly-doc-build.yml

Picking a successful workflow run is based on the default behavior of the [dawidd6 download action](https://github.com/dawidd6/action-download-artifact) and it was already noted as an issue in here: 

https://github.com/dawidd6/action-download-artifact/issues/176

In Pyfluent repo, it was actually **picking** the wrong workflow.... it was picking the latest successful workflow, until the artifacts expired, hence the error message.

I face that myself in PyMAPDL. See: https://github.com/pyansys/pymapdl/issues/1792

## Solution:
You can change the ``workflow_conclusion`` option to be ``in_progress``, so it picks the current run. However, that will probably break other workflows which use artifacts from other workflows (they are not in progress anymore).

We could go for the "skip-all-together-this-option", detailed in the following PR:
https://github.com/dawidd6/action-download-artifact/pull/185

I will outline the PR in a few minutes.

# TLDR
The default ``workflow_conclusion`` argument is ``successful`` so it does not work for downloading the artifacts from an artifact which is running already.

RobPasMue:
I must say I really "hate" this action we are using behind the scenes... I understand the need of using but it is overcomplicating things a lot... I would be in favor of just removing its use and if somebody needs an artifact from another workflow then you just generate that artifact in your new workflow - especially since public repos are not charged by minutes consumed on GH Actions... and they are the ones using it the most.

jorgepiloto:
Thanks for the deep explanation, @germa89.

I saw the changes you made in #156 and decided to add those in here. Even if this pull-request introduces some duplicated logic, it isolates the one from version "pyansys/actions@v2.0.7", which is the one before introducing the `workflow` input.

Isolation ensures that users are able to upgrade to the latest "v3" and that we do not run into additional issues. The PR also provides the logic for those who want to use the `workflow` and `workflow_conclussion` option.

RobPasMue (approved):

germa89:
Fair enough! Thank you for the fix!

PR #156, opened 2023-01-27
Setting a compatible value for ``workflow_conclusion`` argument

germa89:
See https://github.com/pyansys/actions/pull/150#issuecomment-1406795131

Replace #150 

I'm always into exposing things (APIs!! 😄), but this PR could work too just doing:

```yml
    - name: "Download the HTML documentation artifact"
      uses: dawidd6/action-download-artifact@v2
      with:
        github_token: ${{ inputs.token }}
        name: ${{ inputs.doc-artifact-name }}
        path: dev
        workflow: ${{ inputs.workflow }}
        workflow_conclusion: ""
 ```
 
 But I think this is a good oportunity to expose... howeeeeever... I cannot really see in which situations this argument might be useful.

### DIFF
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -29,6 +29,11 @@ inputs:
     required: false
     default: true
     type: string
+  workflow_conclusion:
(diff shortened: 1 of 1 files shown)


RobPasMue:
Closing in favor of implementation in #150

PR #167, opened 2023-02-09
Wheels should be uploaded - error in naming

RobPasMue:
Seen in https://github.com/pyansys/pygeometry/actions/runs/4132811802/jobs/7142191585#step:3:1437

When attempting to perform GitHub release, the wheels are not found because of the artifact naming convention. This has to be fixed and patched released

### DIFF
--- a/release-github/action.yml
+++ b/release-github/action.yml
@@ -61,8 +61,8 @@ runs:
       uses: softprops/action-gh-release@v1
       with:
         files: |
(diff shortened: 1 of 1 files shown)


RobPasMue:
This is done fine here, for example:
https://github.com/pyansys/actions/blob/b791328faa631064bc830a7a3f4b20cc2cd85a58/_release-pypi/action.yml#L52-L56

jorgepiloto (approved):

Issue #142, opened 2023-01-17
Improve error handling


jorgepiloto:
🐞 **Problem**
Actions are useful but act like a black box unsless you refer to their YML files. In addition, actions expecting artifacts with a particular name make the CI/CD pipeline to fail if such artifacts are not found. Error messages for this situations are poor and not explanatory.

💡 **Idea**
Previous situation could be improved with a better error handling:

- For actions looking for particular artifacts name, check that those artifacts exist before running any logic. If they do not exist, raise a proper error indicating how to solve the situation.

PR #183, opened 2023-02-22
FEAT: add logging action

jorgepiloto:
Related with #142. Includes a new action named `_logging` that contains three different levels: error (red), warning (yellow) and info (green).

As an example, consider this actions log: https://github.com/pyansys/test-multiversion/actions/runs/4245246509/jobs/7380502179

Screenshot_20230222_182141

Update. Now the color code is echoed is included directly in the message to avoid the logs showing any variables.

### DIFF
--- a/_logging/action.yml
+++ b/_logging/action.yml
@@ -1,0 +1,47 @@
+name: >
+  Utils logging action
+
+description: >
+  This action is used to generate logging messages during the execution of other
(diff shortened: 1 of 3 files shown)


jorgepiloto:
Once merged, this can be used for general echoing, warning and error messages. Not sure how useful it will be for these last two as you cannot have nested actions.

RobPasMue (approved):
This is supercool @jorgepiloto!

PR #184, opened 2023-02-22
CI: verify private actions do not point to main when releasing

jorgepiloto:
This pull-request adds an extra step in the code style to verify that private actions do not point to main. By private actions I mean reusable actions that we implemented to simplify more complex ones.


RobPasMue (approved):
Interesting! LGTM @jorgepiloto

Issue #178, opened 2023-02-22
Allow PDF to be optional in actions/release-github


iboyd-ansys:
### 📝 Description of the feature

In my project PDF doc builds fail, for reasons that no one has been able to understand. It has not been considered critical because the primary documentation is HTML and this is fine. However, the lack of PDF doc causes the `release-github` action to fail.

It would be useful to allow PDF to be optional, to reflect that it is a secondary requirement.

### 💡 Steps for implementing the feature

Add a new optional input, boolean valued with default `true`, that specifies whether PDF is expected to be included in the release.

Adjust the logic of the steps to account for the flag.

jorgepiloto:
Hi @iboyd-ansys, thanks for opening this issue.

The philosophy for `pyansys/actions` is very strict about existant artifacts and their naming. When we implemented the `release-github` action, we imposed the following criteria:

* Always deploy `wheels` and `sdist` artifacts
* Always deploy `documentation-html`
* Always deploy `documentation-pdf`
* Always deploy `wheelhouse`
* Allow to deploy any additional artifacts

**If a Python library can not generate this set of artifacts, it means there is some issue on their side. This forces us to go and explore such issue.**

Hence, I am closing this request.

However, let's work on solving this issue on your side.

PR #186, opened 2023-02-23
FEAT: allow additional artifacts and dry-runs for critical workflows

### LINKED ISSUE #138: Lack of consistency on paths for released artifacts
The ``release-to-PyPI`` and ``release-to-Github`` are not consistent.

The PyPI action download the artifacts (all) to a folder called ``${{ inputs.library-name}}-artifacts`` but then, the ``release-to-github`` action check the folder ``other-artifacts``.

### LINKED ISSUE #139: Implement ``dry-run`` option for critical actions
Our actions should have a ``dry-run`` option for the important/only-can-do-once things... like release to PyPI.

### LINKED ISSUE #140: Allow to publish additional artifacts in release-github action
So far, the action uploads everything. I don't think this is a good approach. Everything should be optional.

```yml

    - name: "Release to GitHub"
      uses: softprops/action-gh-release@v1
      with:
        files: |
          ${{ inputs.library-name }}/*.whl
          ${{ inputs.library-name }}/*.tar.gz
          documentation-html.zip
          documentation-pdf.zip
          other-artifacts/**/*-wheelhouse-*.zip
```

jorgepiloto:
Fix #138, fix #139 and fix #140 by allowing to download additional artifacts. Related to #178 too.

### DIFF
--- a/_release-pypi/action.yml
+++ b/_release-pypi/action.yml
@@ -43,6 +43,15 @@ inputs:
     required: true
     type: string
 
+  dry-run:
+    description: >
+      Whether to run or not this action in testing mode. Testing mode executes
+      all the action steps except the releasing artifacts one. This allows
+      developers to verify the health of the action once integrated in their
+      CI/CD pipelines without actually publishing artifacts.
+    required: true
+    type: string
+
 runs:
   using: "composite"
   steps:
@@ -70,6 +79,7 @@ runs:
 
     - name: "Upload artifacts to private PyPi"
       shell: bash
+      if: ${{ inputs.dry-run != "false" }}
       run: |
         python -m twine upload --verbose --skip-existing ${{ inputs.library-name }}-artifacts/*.whl
         python -m twine upload --verbose --skip-existing ${{ inputs.library-name }}-artifacts/*.tar.gz
--- a/release-github/action.yml
+++ b/release-github/action.yml
@@ -100,11 +100,28 @@ runs:
           Create a dist/wheelhouse directory and move all wheelhouse artifacts
           inside this folder.
 
-    - name: "Move documentation artifacts to dist/documentation directory"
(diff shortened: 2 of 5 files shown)


_release-pypi/action.yml
```diff
@@ -43,6 +43,15 @@ inputs:
     required: true
     type: string
 
+  dry-run:
+    description: >
+      Whether to run or not this action in testing mode. Testing mode executes
+      all the action steps except the releasing artifacts one. This allows
+      developers to verify the health of the action once integrated in their
+      CI/CD pipelines without actually publishing artifacts.
+    required: true
```
MaxJPRey, line 52: I am wondering here if we should have the same arguments as for `release-github` action.
I was thinking that usually when you do a release, you want to publish on PyPI -so the logic should be similar-.
But it is actually not always true. So I am open to any decisions.

```suggestion
    required: false
    default: false
```

RobPasMue: Dry-run should definitely be optional. This is must be optional.

RobPasMue: Ahh now I see... This is the private action. In that case, that's fine

RobPasMue (approved):

_release-pypi/action.yml
jorgepiloto: That's it, @RobPasMue. This is the private action. The convention is that private actions do not have optional arguments to be transparent when called from public actions.

By default, the "dry-run" is disabled in:

* `release-pypi-public`
* `release-pypi-private`
* `release-pypi-test`
* `release-github`

MaxJPRey: OK make sense. I did not notice that it was private. Thanks for the clarification.

MaxJPRey (approved):

PR #189, opened 2023-02-24
Allowing last N-Days on hk-package actions

RobPasMue:
As title says

### DIFF
--- a/hk-package-clean-except/action.yml
+++ b/hk-package-clean-except/action.yml
@@ -20,7 +20,7 @@ inputs:
     default: 'pyansys'
     type: string
   allow-last-days:
-    description: "Avoid removing the last N days image: e.g. '2'."
+    description: "Avoid removing the last N days images: e.g. '2'."
     required: false
     default: ''
     type: string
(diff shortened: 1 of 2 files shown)


RobPasMue:
Pinging @greschd for visibility on implementation. I think this is just what we needed. Something simple.

hk-package-clean-except/action.yml
```diff
@@ -19,6 +19,11 @@ inputs:
     required: false
     default: 'pyansys'
     type: string
+  allow-last-days:
+    description: "Avoid removing the last N days images: e.g. '2'."
+    required: false
+    default: ''
+    type: string
```
greschd, line 26: Is there no 'int' type here?

RobPasMue: Sadly, no... I wish there were. I think the types allowed are these only https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputsinput_idtype

Even if it is not for the same location, that's the types they accept. In fact, everything is a string behind the scenes...

PR #198, opened 2023-02-27
FIX: release-github syntax

jorgepiloto:
As per the issues found in https://github.com/ansys/ansys-sphinx-theme/actions/runs/4284451985/jobs/7461503095.


Revathyvenugopal162 (approved):
Thanks Jorge for fixing this

PR #205, opened 2023-02-28
Allow skip of the PDF build

### LINKED ISSUE #204: Add ability to disable building pdf on `doc-build`
`doc-build-action` needs an option to not build the pdf. It's not always necessary and sometimes a pain to setup.

https://actions.docs.pyansys.com/version/stable/doc-actions/index.html#doc-build-action

akaszynski:
Resolve #204 by allowing the user to skip the pdf build.

### DIFF
--- a/doc-build/action.yml
+++ b/doc-build/action.yml
@@ -81,6 +81,14 @@ inputs:
     required: false
     type: boolean
 
+  skip-pdf-build:
+    description: >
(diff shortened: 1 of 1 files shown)


RobPasMue (changes requested):
Though the implementation is correct, I am blocking this PR until we finalize discussion on the issue related #204

Issue #188, opened 2023-02-24
Explore commit convention and towncrier for building changelogs in an automated way


jorgepiloto:
💡 **Idea**
@AlejandroFernandezLuces suggested to follow [conventional commits] to unify commit messages. This is a great idea as it is an accepted standard. On top of that, this convention is used by changelog tools to create this document based on those commits.

- [x] **Implement a `check-conventional-commit` action**
  The goal is to make use of [amannn/action-semantic-pull-request](https://github.com/amannn/action-semantic-pull-request). By using this action together with [squash and merging](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) and the "commit message as pull-request title", we could ensure [conventional commits] without annoying developers when actually making commits. Everything happens in the pull-request.
  
- [x] **Implement a `check-changelog-fragments` action**
  The goal is to ensure that any pull-request (unless some condition) contains a changelog fragment requested by [towncrier](https://github.com/twisted/towncrier). [The Astropy project](https://www.astropy.org/) uses this tool in a very smart way together with  the [action-towncrier](https://github.com/pllim/actions-towncrier-changelog) action.
  
- [x] **Implement a `build-changelog` action**
  This action generates the changelog file following the configuration declared in the `pyproject.toml` file of every project. Ideally, we should enforce the [keep a changelog](https://keepachangelog.com/en/1.0.0/) convention.

[conventional commits]: https://www.conventionalcommits.org/en/v1.0.0/

AlejandroFernandezLuces:
I just did the first point today in a personal repo, to avoid testing in our public repositories. I also seized the opportunity to implement a branch naming convention checker, to avoid people from merging un labeled branches. @MaxJPRey is taking a look at it, I'll share the link and if you think it's appropriate we can open a PR with those actions here: https://github.com/AlejandroFernandezLuces/trame-slides/blob/main/.github/workflows/check-name.yml

jorgepiloto:
Thanks for linking this, @AlejandroFernandezLuces. Let us explore [amannn/action-semantic-pull-request](https://github.com/amannn/action-semantic-pull-request) in favor of [aslafyz/conventional-pr-title](https://github.com/aslafy-z/conventional-pr-title-action), as the first one supports more options.

The one I am interested the most is the `ignore labels` one.

klmcadams:
@jorgepiloto yep that sounds good!

PR #209, opened 2023-03-06
FEAT: Add PR title name checker and branch name checker

AlejandroFernandezLuces:
This PR adds two new actions:
- One checks that the title of the PR follows conventional commits.
- The other checks if the branch name follows our branch naming convention.

Related to #188


check-branch-name/action.yml
```diff
@@ -0,0 +1,22 @@
+name: "Check if the branch name follows our naming convention."
+description: "Action that checks if the name of the branch is following the our branch naming convention."
+
+runs:
+  using: composite
+  steps:
+    - uses: jungwinter/split@master
+      id: split
+      with:
+        msg: ${{ env.BRANCH_NAME }}
```
jorgepiloto, line 10: As far as I know, there is no `env.BRANCH_NAME` in Github actions, see the [GitHub actions context](https://docs.github.com/en/actions/learn-github-actions/contexts).

There is a variable named `github.ref_name`, that refers to the branch or tag that triggered the workflow. If this variable is used with `github.ref_type` (possible values are "branch" or "tag"), the branch name can be extracted.

```yml
if: github.ref_type == 'branch':
  ...
```

Maybe, we could delegate this conditional to the user. What do you think?

check-branch-name/action.yml
```diff
@@ -0,0 +1,22 @@
+name: "Check if the branch name follows our naming convention."
+description: "Action that checks if the name of the branch is following the our branch naming convention."
+
+runs:
+  using: composite
+  steps:
+    - uses: jungwinter/split@master
```
jorgepiloto, line 7: Instead of using a third party action for splitting a string, let us go with the `cut` command. For using this command, you specify the "delimiter" and the "field" you want to retain after cutting the string.

```yml
- name: "Get the prefix of the branch"
  if: github.ref_type == 'branch'
  shell: bash
  run: |
    branch_prefix=$(cut -d '/' -f 1)
    echo "BRANCH_PREFIX=$branch_prefix" >> $GITHUB_ENV
```

jorgepiloto (changes requested):

check-branch-name/action.yml
```diff
@@ -0,0 +1,22 @@
...
+
+    - name: Check branch name has slash
+      shell: bash
+      if: contains('/', env.BRANCH_NAME) == false
+      run: echo "[ERROR] Branch name ${{ env.BRANCH_NAME }} does not contain a prefix and a backslash." && exit 1
+
+    - name: Check branch prefix naming
+      shell: bash
+      if: contains(fromJson('["fix", "feat", "junk", "maint", "doc", "no-ci", "testing", "release"]'), steps.split.outputs._0) == false
+      run: echo "[ERROR] Branch name ${{ env.BRANCH_NAME }} prefix is not valid." && exit 1
```
jorgepiloto, line 22: I would suggest to use the recently added `pyansys/actions/logging` action. It allows to send ANSI color logging messages that can be quickly identified through  all the actions report.

check-branch-name/action.yml
```diff
@@ -0,0 +1,22 @@
...
+        maxsplit: -1
+
+    - name: Check branch name has slash
+      shell: bash
+      if: contains('/', env.BRANCH_NAME) == false
+      run: echo "[ERROR] Branch name ${{ env.BRANCH_NAME }} does not contain a prefix and a backslash." && exit 1
+
+    - name: Check branch prefix naming
+      shell: bash
+      if: contains(fromJson('["fix", "feat", "junk", "maint", "doc", "no-ci", "testing", "release"]'), steps.split.outputs._0) == false
```
MaxJPRey, line 21: @AlejandroFernandezLuces @jorgepiloto Isn't `test` accepted for the branch name?
If we look for `test` then both `test` and `testing` will work.
```suggestion
      if: contains(fromJson('["fix", "feat", "junk", "maint", "doc", "no-ci", "test", "release"]'), steps.split.outputs._0) == false
```

check-branch-name/action.yml
AlejandroFernandezLuces: I don't see why we would want to delegate this conditional to the user. We always want to use this action on branches, not on tags, right? We can't perform any check on the tags with this action.

RobPasMue:
@AlejandroFernandezLuces @MaxJPRey @jorgepiloto what's the status here?

AlejandroFernandezLuces:
> @AlejandroFernandezLuces @MaxJPRey @jorgepiloto what's the status here?

I was just checking the errors in the actions. Also, I was waiting for @jorgepiloto feedback on the remaining open conversation.

AlejandroFernandezLuces:
It seems that the doc building isn't handling actions without inputs properly. Maybe I'm missing something here, should all actions have at least one input? If not, I can try to fix the doc generation

RobPasMue:
Just did a trial attempt to define "empty" inputs. Let's see if it works

RobPasMue:
Reverted changes and went down the robust approach. See commit: https://github.com/pyansys/actions/pull/209/commits/753e2b07bb3adee5eb49133cea89e0367b15118c

* You were not including them in the docs. You actually had to include a new section or similar for the ``Other actions``.
* Naming convention was wrong for ``pr-conventional...``, has to start with one of our allowed prefixes. I added the ``check-`` prefix.
* Also modified the ``conf.py`` to accept no-input actions. That way everything is more robust.

IMO, ready to merge

RobPasMue (approved):

MaxJPRey (approved):
LGTM.

check-branch-name/action.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Check branch name has slash
+      shell: bash
+      env:
+        BRANCH_NAME: ${{ github.head_ref || github.ref_name }} 
+      if: contains(env.BRANCH_NAME, '/') == false
+      uses: pyansys/actions/_logging@main
+      with:
+        level: "ERROR"
+        message: >
+          "Branch name ${{ env.BRANCH_NAME }} does not contain a prefix and a backslash."
```
jorgepiloto, line 25:
```suggestion

    - name: "Get the prefix of the branch"
      env:
        BRANCH_NAME: ${{ github.head_ref || github.ref_name }} 
      if: github.ref_type == 'branch'
      shell: bash
      run: |
        branch_prefix=$(echo "${{ BRANCH_NAME }}" | cut -d '/' -f 1)
        echo "BRANCH_PREFIX=$branch_prefix" >> $
        
    - name: Check branch name has slash
      shell: bash
      env:
        BRANCH_NAME: ${{ github.head_ref || github.ref_name }} 
      if: contains(env.BRANCH_NAME, '/') == false
      uses: pyansys/actions/_logging@main
      with:
        level: "ERROR"
        message: >
          "Branch name ${{ env.BRANCH_NAME }} does not contain a prefix and a backslash."
```

PR #218, opened 2023-04-05
feat: implement version/index.html page

### LINKED ISSUE #117: Implement "Older versions" link in multi-version drop down
💡 **Idea**
The main idea behind this issue is to have a link named "Older versions" in the drop down button for the multi-version documentation. This link must only appear after more than 3 stable versions are found for a project.

The idea is simple, but the implementation is a bit tricky. The main issue is that rendered documentation are not aware about the version context. This implies, that, we need to create the page and inject it after the content has been rendered.

1. The `ansys-sphinx-theme` needs to generate a page that is not linked to any other one. This page must be named `releases.html` and should contain a `Jinja` section to be rendered after the documentation has been generated. This is the only way we can ensure that the documentation sections and style match the latest stable release of a project.
2. In the `doc-deploy-stable`, it renders previous Jinja context with the name and URL for each existing version. 
3. The HTML page is placed in the `gh-pages` in the `release/index.html` path.

jorgepiloto:
Fix #117 by adding a `version/index.html` page that reads the content of the `versions.json` file and creates a table showing all the versions available in the documentation.


doc-deploy-dev/versions.js
```diff
@@ -0,0 +1,41 @@
...
+        document.documentElement.innerHTML = html;
+      }),
+    fetch("versions.json")
+      .then((res) => res.json())
+      .then((data) => {
+        var table = "<h2>Older versions</h2>";
+        table += "<table class='table'>";
+        table += "<th> versions </th>";
+        table += "<th> url </th>";
+        for (let versions of data) {
```
jorgepiloto, line 20: Let's play also with a variable named `ith_version` so we can add the `row-odd` and `row-even` classes to the right rows in the table. This guarantees that the final style of the table matches the one from the ansys-sphinx-theme.

Revathyvenugopal162:
keeping the ss of output with script for reference.
### expected output

image

jorgepiloto:
Thanks for adding that script, @Revathyvenugopal162. Let me integrate it inside of the actions.

jorgepiloto:
I just realized that we are forced to collect the directories next to the 'version/index.html' to check which versions are available. The 'versions.json' file should only contain the latest three and a link to the 'version/index.html' page with the title `Older versions`.

Revathyvenugopal162:
> I just realized that we are forced to collect the directories next to the 'version/index.html' to check which versions are available. The 'versions.json' file should only contain the latest three and a link to the 'version/index.html' page with the title `Older versions`.

Yes true, i was just reading the json to write the table.
But reading directories with js will give duplicate the stable version and its actual version number.because the stable doc have the duplicated files in the latest version folder i think(0.64 and stable have same thing in mapdl)

doc-deploy-dev/action.yml
```diff
@@ -211,6 +211,144 @@ runs:
...
+    - name: "Update all references to point to 'stable/_static' or 'dev/_static'"
+      shell: bash
+      run: |
+        if [[ -f 'version/stable/index.html' ]]; then
+          sed -i "s|_static|stable/_static|g" version/index.html
+        else
+          sed -i "s|_static|dev/_static|g" version/index.html
+        fi
+
+    - name: "Declare the script to read the 'versions.json' file"
```
jorgepiloto, line 240: No need to scrap anything. The repository gets checkout first of executing any logic. Therefore, just a list of desired folders versions and pass it to the desired JS code.

jorgepiloto (commented):
Leaving some comments before I forget about cleaning this implementation.

doc-deploy-dev/action.yml
jorgepiloto: Thank you, Jorge from the past.

RobPasMue (approved):
Looks great @jorgepiloto - let's just modify the description of the action and add my minor typo suggestion =)

_doc-gen-versions-page/action.yml
```diff
@@ -0,0 +1,80 @@
...
+description: |
+  This action generates the ``version/index.html`` page. This page collects all
+  the releases available for a library. This feature allows to not overload the
+  content of the dropdown button for the multi-version.
+
+  .. warning::
+
+      This is a private action and its meant to be reused by ``doc-deploy-*``
+      actions.
+
```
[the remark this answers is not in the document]
RobPasMue: I mean it is not the "Doc deploy dev action"

PR #227, opened 2023-04-17
FIX: conventional commits

jorgepiloto:
This pull-request collects the style checkers for branch and commit names into the "style" group. It also fixes some issues regarding the input variables, as those were listed as secrets.


check-pr-conventional-name/action.yml
```diff
@@ -1,9 +0,0 @@
-name: "Check if the pull request title follows conventional commits."
-description: "Action that checks if the title of the pull request is following the conventional commits standard."
-
-runs:
-  using: composite
-  steps:
-    - uses: amannn/action-semantic-pull-request@v5
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jorgepiloto, line 9: We do not inherit secrets. Instead, we are explicit when passing those as inputs. This is why the action was failing.

jorgepiloto: Now, this is reading `${{ inputs.token }}`.

doc/source/style-actions/examples/commit-style-basic.yml
```diff
@@ -0,0 +1,5 @@
+commit-style:
+  name: "Run commit style checks"
+  runs-on: ubuntu-latest
+  steps:
+    - uses: pyansys/actions/commit-style@{{ branch }}
```
RobPasMue, line 5:
```suggestion
    - uses: pyansys/actions/commit-style@{{ branch }}
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
```

RobPasMue (approved):
LGTM!

PR #235, opened 2023-04-18
FIX: update apt-get

RobPasMue:
Following https://github.com/actions/runner-images/issues/7452


RobPasMue:
Doing a hotfix right away with this change

PR #268, opened 2023-05-22
feat: Adding option for repository when cleaning packages

germa89:
Currently you cannot clean packages under a repository. Only if they belong to the organization.
This is fine, but it is a bit limiting when we have packages like [pymapdl/mapdl](https://github.com/ansys/pymapdl/pkgs/container/pymapdl%2Fmapdl) (At PyMAPDL we do things differently to everybody else it seems)

This PR adds the possibility to include the repo where you are cleaning the packages.

Based on:
https://github.com/fastai/ghapi/issues/84

### DIFF
--- a/hk-package-clean-except/action.yml
+++ b/hk-package-clean-except/action.yml
@@ -19,6 +19,11 @@ inputs:
     required: false
     default: 'ansys'
     type: string
+  repository:
+    description: "Repository at which packages are published."
+    required: false
+    default: ''
+    type: string
   allow-last-days:
     description: "Avoid removing the last N days images: e.g. '2'."
     required: false
@@ -62,6 +67,7 @@ runs:
         PACKAGE_TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
         org_str = "${{ inputs.package-org }}"
+        repo_str = "${{ inputs.repository }}"
         pck_str = "${{ inputs.package-name }}"
         last_days_str = "${{ inputs.allow-last-days }}"
         valid_tags_str = "${{ inputs.tags-kept }}"
@@ -69,7 +75,7 @@ runs:
         last_days = int(last_days_str) if last_days_str != "" else None
         valid_tags = [x.strip() for x in valid_tags_str.split(",")]
 
-        api = GhApi(debug=print_summary, token=os.getenv("PACKAGE_DELETION_TOKEN"))
+        api = GhApi(debug=print_summary, repo=repo_str, token=os.getenv("PACKAGE_DELETION_TOKEN"))
 
         if last_days:
             delete_before_date = datetime.now() - timedelta(days=last_days)
--- a/hk-package-clean-untagged/action.yml
+++ b/hk-package-clean-untagged/action.yml
@@ -15,6 +15,11 @@ inputs:
     required: false
     default: 'ansys'
     type: string
+  repository:
+    description: "Repository at which packages are published."
+    required: false
+    default: ''
+    type: string
   allow-last-days:
     description: "Avoid removing the last N days images: e.g. '2'."
     required: false
@@ -58,6 +63,7 @@ runs:
         PACKAGE_TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
         org_str = "${{ inputs.package-org }}"
+        repo_str = "${{ inputs.repository }}"
         pck_str = "${{ inputs.package-name }}"
         last_days_str = "${{ inputs.allow-last-days }}"
 
@@ -65,7 +71,7 @@ runs:
         if last_days:
             delete_before_date = datetime.now() - timedelta(days=last_days)
 
-        api = GhApi(debug=print_summary, token=os.getenv("PACKAGE_DELETION_TOKEN"))
+        api = GhApi(debug=print_summary, repo=repo_str, token=os.getenv("PACKAGE_DELETION_TOKEN"))
 
         paged_packages = paged(
             api.packages.get_all_package_versions_for_package_owned_by_org,


greschd:
Are you sure this is necessary? In the `ansys-internal` org we have packages that are linked to the repository, but cleaned by the action nonetheless.

germa89:
> Are you sure this is necessary? In the `ansys-internal` org we have packages that are linked to the repository, but cleaned by the action nonetheless.

Really? Can you send me the link to those workflows?

RobPasMue:
I think I know understand your point... But it might be more complicated than just doing this.

RobPasMue:
For those cases, your package name is actually ``repo/yourname`` such that you have ``pyansys/mapdl``... In any case, we should try avoid doing that. Packages should be called ``ghcr.io/<org>/<package-name>`` and avoid using ``/`` inside ``<org>`` or ``<package-name>``. I've experienced issues in the past because of this

greschd:
> Really? Can you send me the link to those workflows?

Summary of "offline" discussion: The packages I had in mind in `ansys-internal` don't apply, since they are not using an extra `/` in the package name.

greschd:
> I've experienced issues in the past because of this

Can you expand on this, besides the limitation this PR is trying to address?

RobPasMue:
> > I've experienced issues in the past because of this
> 
> Can you expand on this, besides the limitation this PR is trying to address?

Yep... see https://github.com/ansys-internal/api-eigen-example/actions/workflows/package_cleanup.yml

I tried in multiple ways to provide it but the way in which the request is parsed somehow dealt poorly with the ``/`` sybol. Maybe @germa89's PR will solve this. I need to test it out locally first.

germa89:
> For those cases, your package name is actually `repo/yourname` such that you have `pyansys/mapdl`... In any case, we should try avoid doing that. Packages should be called `ghcr.io/<org>/<package-name>` and avoid using `/` inside `<org>` or `<package-name>`. I've experienced issues in the past because of this

I kind of agree.... I was thinking about leaving `ansys/mapdl` for the vanilla docker image. And then using `ansys/pymapdl/mapdl` for a customized version of that docker image for running CICD and/or do development.


> I tried in multiple ways to provide it but the way in which the request is parsed somehow dealt poorly with the / sybol. Maybe @germa89's PR will solve this. I need to test it out locally first.

That was the goal (deal with `/`). It worked for me locally, but I might have missed something. Please test it.

Also we might want to rename the ``repository`` argument for something more generic? Just in case someone wants their image to be `ansys/ace/mapdl` for example. (Or maybe that case should not be covered)

greschd:
> Also we might want to rename the repository argument for something more generic? Just in case someone wants their image to be ansys/ace/mapdl for example. (Or maybe that case should not be covered)

I tend to think we should just use the `package-name` argument for that, and make it accept names with `/`. If directly using the ghapi in this way doesn't work, we could just split the argument in our logic.
As far as I understand (might be mistaken..) the `/` is part of the package name in GHCR's terminology.

RobPasMue:
> I tend to think we should just use the `package-name` argument for that, and make it accept names with `/`. If directly using the ghapi in this way doesn't work, we could just split the argument in our logic. As far as I understand (might be mistaken..), `pyansys/mapdl` _is_ the package name in GHCR's terminology.

I agree with @greschd - let's do the name mangling on the script but accept it as "package-name" directly.

greschd:
@RobPasMue see my sneak edit above..

RobPasMue:
>Hmm no.. what's the difference between containers that don't render their `/` on the "packages" list (https://github.com/orgs/ansys/packages) such as `pyansys/mapdl`, and those that do (see https://github.com/orgs/ansys-internal/packages)?

I do think it's just the name of the package.

RobPasMue:
@greschd See my also sneak edit above 😄

germa89:
Soo..... do we approve this or not? XD

RobPasMue:
> Soo..... do we approve this or not? XD
> 
> I think having `/` is interesting. For example we could have `pymapdl/mapdl` for the docker image for developing on PyMAPDL (we could include the repo, currently we dont). Anyways this could be solved also with `mapdl-pymapdl`..... so it is not strictly needed.

We should approve it but I need to test it first, sorry... havent gotten to it yet

RobPasMue:
Getting back to this

RobPasMue:
The following script "should" work for the following package, but it isn't working. Please provide a token with the required auth.

https://github.com/ansys/api-eigen-example/pkgs/container/api-eigen-example%2Fpython-grpc-server

```py
from datetime import datetime, timedelta
import os

from ghapi.all import GhApi
from ghapi.core import print_summary
from ghapi.page import paged

PACKAGE_TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"

org_str = "ansys"
repo_str = "api-eigen-example"
pck_str = "python-grpc-server"
last_days_str = "7"
valid_tags_str = "latest"

last_days = int(last_days_str) if last_days_str != "" else None
valid_tags = [x.strip() for x in valid_tags_str.split(",")]

api = GhApi(debug=print_summary, repo=repo_str, token=os.getenv("PACKAGE_DELETION_TOKEN"))

if last_days:
    delete_before_date = datetime.now() - timedelta(days=last_days)

paged_packages = paged(
    api.packages.get_all_package_versions_for_package_owned_by_org,
    org=org_str,
    package_name=pck_str,
    package_type="container",
    state="active",
    per_page=100,
)

# Loop over all pages
for page in paged_packages:
    for package in page:
        # If keeping last days is requested, then check
        if (
            last_days
            and datetime.strptime(package.created_at, PACKAGE_TIME_FORMAT).timestamp()
            > delete_before_date.timestamp()
        ):
            continue

        # Check if the given package has valid tags
        package_tags = package.metadata.container.tags
        delete = True
        for tag in package_tags:
            if tag in valid_tags:
                delete = False
                break
        # In case it should, delete it
        if delete:
            print(f"Deleting package: {package}")
            api.packages.delete_package_version_for_org(
                org=org_str,
                package_name=pck_str,
                package_type="container",
                package_version_id=package["id"],
            )
```

Issue #269, opened 2023-05-29
Build wheelhouse action piping unintended output and breaking tests


raph-luc:
The following line is trying to pipe the entire output of `import` and `print` commands into the `GITHUB_ENV` variable:
https://github.com/ansys/actions/blob/a78b246d2b1319a6ed1bc66e1fdd0e635f66b660/build-wheelhouse/action.yml#L92

If on `import` the package outputs anything, it breaks the action as it all gets merged into the `GITHUB_ENV` var. I believe that is the reason for this failure:
https://github.com/ansys/pyfluent/actions/runs/5114660367/jobs/9195134134?pr=1651
Due to:
![image](https://github.com/ansys/actions/assets/132297401/3cefe1f2-7fe7-4a07-9fbd-e01df5ca297c)

I get `bash: $GITHUB_ENV: ambiguous redirect` if I try piping the entire output into a variable too, not sure how/why it works on the github runners (is it not bash?).

I believe we only want the output from the `print(__version__)` command piped into the `GITHUB_ENV` variable, in which case my suggestion would be to change [line 92](https://github.com/ansys/actions/blob/a78b246d2b1319a6ed1bc66e1fdd0e635f66b660/build-wheelhouse/action.yml#L92) to: 

-  `echo "library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(); print(__version__)" | tail -1)" >> $GITHUB_ENV` so that we only pipe the last line of the output, which should include only the output from the `print(version)` command we want. Also that extra `print();` to make sure output doesn't include any previous std output that did not end with a line break (shouldn't happen but probably better be safe).

An example commit where this was fixed and then the action `Build Documentation` which had previously failed for the same reason (and does not use `build-wheelhouse`) passed: https://github.com/ansys/pyfluent/pull/1651/commits/c65473dd550f322f9087cc8416f6f74677fb26d2

RobPasMue:
My first question here @raph-luc is... where is the first print statement coming from? Do you print information in import? That's not really a good idea IMO, letting @jorgepiloto provide feedback here.

Also, why would I want a pyfluent log file when the only thing I am doing is importing the version?

For me the real problem is there, rather than adapting the version catcher statement... Of course we could adapt it but I still think that we should address the real issue here... why are you printing information on import and why are you creating a log file right after import... when maybe it isn't needed (for example, after retrieving the version I won't be doing anything else so... no need for logger).

jorgepiloto:
[...]

PR #276, opened 2023-07-04
fix: check proper import and version variable

### LINKED ISSUE #275: Add `import` testing inside the smoke tests actions

In PyMAPDL we ran across a simple issue #2161 where because of changes in Python we couldn't catch it. Testing is done in Python 3.9, whereas this issue is raised only 3.8.

A simple module import would have raise an error which then would have shown in the CICD.

Therefore, I would propose to have an option for add a simple smoke test (simple `python -c "import library"`) step (conditional) in the CICD.

I would draft a PR soon

_No response_

germa89:
Because of `$()` if there is an error while importing the library, the exception/error is hidden because it does not affect the parent process. 

With this PR, we check that the variable has ben properly created which means that the package can be imported properly.

Closes #275

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -89,7 +89,7 @@ runs:
     - name: "Verify library is properly installed and get its version number"
       shell: bash
       run: |
-        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)" || exit 1 )
+        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(); print(__version__)" | tail -1 || exit 1 )
         if [ -z "$library_version" ]; then
             echo "Problem getting the library version"
             exit 1;
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -89,7 +89,14 @@ runs:
     - name: "Verify library is properly installed and get its version number"
       shell: bash
       run: |
-        echo "library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)")" >> $GITHUB_ENV
+        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)" || exit 1 )
+        if [ -z "$library_version" ]; then
+            echo "Problem getting the library version"
+            exit 1;
+        else
+            echo "The library version is: $library_version";
+        fi;
+        echo "$library_version" >> $GITHUB_ENV
 
     - name: "Generate the wheelhouse"
       shell: bash


germa89:
@jorgepiloto and @RobPasMue right now...

build-wheelhouse/action.yml
```diff
@@ -89,7 +89,14 @@ runs:
...
       shell: bash
       run: |
-        echo "library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)")" >> $GITHUB_ENV
+        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)" || exit 1 )
+        if [ -z "$library_version" ]; then
+            echo "Problem getting the library version"
+            exit 1;
+        else
+            echo "The library version is: $library_version";
+        fi;
```
RobPasMue, line 98: Well it seems to me that this "double exit" is oging to make parts of the code unusable... Shouldn't it be enough to just change line 92?
```suggestion
        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)" || exit 1 )
```

RobPasMue (changes requested):

build-wheelhouse/action.yml
RobPasMue: Or the other way around, I don't really care. In fact, it's probable better
```suggestion
        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)")
        if [ -z "$library_version" ]; then
            echo "Problem getting the library version"
            exit 1;
        else
            echo "The library version is: $library_version";
        fi;
```

RobPasMue:
I saw this issue some weeks ago, yeah... Thanks for opening the PR @germa89!

build-wheelhouse/action.yml
```diff
@@ -89,7 +89,14 @@ runs:
     - name: "Verify library is properly installed and get its version number"
       shell: bash
       run: |
-        echo "library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)")" >> $GITHUB_ENV
+        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)" || exit 1 )
+        if [ -z "$library_version" ]; then
+            echo "Problem getting the library version"
+            exit 1;
+        else
+            echo "The library version is: $library_version";
```
raph-luc, line 97: I think this `echo` is going to be helpful, this should help address https://github.com/ansys/actions/issues/269 as well

RobPasMue: Yep, agreed... if we get a multiline result, it should print it out as well. This will close #269

build-wheelhouse/action.yml
raph-luc: That exit inside `library_version=$(... || exit 1)` works fine here, in my understanding it only exits out of the `$(command)` expression, all the lines below it should still be processed

raph-luc: In my preliminary tests here the exit inside `library_version=$(... || exit 1)` works fine, in my understanding it is only exiting out of the `$(command)` expression, not out of the bash shell.

(Had previously deleted this comment by mistake, sorry for the repeated emails/notifications)

build-wheelhouse/action.yml
```diff
@@ -89,7 +89,14 @@ runs:
     - name: "Verify library is properly installed and get its version number"
       shell: bash
       run: |
-        echo "library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)")" >> $GITHUB_ENV
+        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(__version__)" || exit 1 )
```
raph-luc, line 92:
```suggestion
        library_version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(); print(__version__)" | tail -1 || exit 1 )
```
@RobPasMue @jorgepiloto FYI This is what I had suggested in my wall of text here: https://github.com/ansys/actions/issues/269, to avoid failing due to extra unexpected stdout

To be clear, just a suggestion/food for thought, and I am not aware of any downsides to this change

RobPasMue: Yep, I think we can add it as well. Should be okay.

build-wheelhouse/action.yml
RobPasMue: Let's leave it as is. It's fine 😄

RobPasMue (approved):
All good @germa89 - merging and patch releasing

build-wheelhouse/action.yml
germa89: My thoughts were that the exit inside the `$()` should avoid having something in that variable as a consequence of the error... like the the traceback. But probably it does not make. I'm Ok with whatever you decide guys. You know better. Although it is already merged xD

PR #277, opened 2023-07-05
feat: enable link check as an optional input

### LINKED ISSUE #273: Implementing link check in docs build action as an optional argument
We should enable some kind of optional argument (true by default, maybe?) to perform a linkcheck using Sphinx

```sphinx-build -M linkcheck $(SOURCEDIR) $(BUILDDIR) $(SPHINXOPTS)```

```yml
    - name: "Build HTML, PDF, and JSON documentation"
      if: inputs.requires-xvfb == 'false'
      shell: bash
      run: |
        make -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
        make -C doc pdf
        if [[ ${{ inputs.check_links }} == 'true' ]];
        then
          make -C doc linkcheck SPHINXOPTS="${{ inputs.sphinxopts }}"
        fi
        if [[ ${{ inputs.skip-json-build }} == 'false' ]];
        then
          make -C doc json SPHINXOPTS="${{ inputs.sphinxopts }}"
        fi
```

RobPasMue:
Closes #273

### DIFF
--- a/doc-build/action.yml
+++ b/doc-build/action.yml
@@ -98,7 +98,7 @@ inputs:
 
   check-links:
     description: >
-      Whether to perform external links checks during the generation of
+      Whether to perform external link checks during the generation of
       documentation. Default value is ``false``.
     default: false
     required: false
--- a/doc-build/action.yml
+++ b/doc-build/action.yml
@@ -96,6 +96,14 @@ inputs:
     required: false
     type: boolean
 
+  check-links:
+    description: >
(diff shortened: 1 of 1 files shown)


MaxJPRey:
I think that it was the rigth approach to set the default value to `False` at first.
Let's see how it behave on some repos and potentially set it to `True`.

RobPasMue:
Yep, follow up issue on #278

PR #285, opened 2023-07-19
feat: stop asking for library-namespace

### LINKED ISSUE #280: Detect project name and namespace from pyproject.toml

Currently we need to manually specify the project name and namespace in the workflow file in order to run `build-wheelhouse`. This is a bit of a shame, as we have the above information stored in `pyproject.toml`, it is not beyond the realm of possiblity to read this file...

Investigate whether we can read this information in a sensible way, perhaps with importlib-metadata, or by using tomli

https://python-poetry.org/docs/pyproject/
https://flit.pypa.io/en/stable/pyproject_toml.html

jorgepiloto:
Fix #280 by retrieving the library version from its name using the `importlib.metadata` module. Once this is merged, a deprecation should be added in @v4.0.

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -99,7 +99,7 @@ runs:
       run: |
         library_name=${{ inputs.library-name }}
 
-        if [ ${{ env.needs_importlib_metadata }} ]; then 
+        if [ ${{ env.needs_importlib_metadata }} ]; then
           version=$(python -c "import importlib_metadata; print(importlib_metadata.version('$library_name'))")
         else
           version=$(python -c "import importlib.metadata as importlib_metadata; print(importlib_metadata.version('$library_name'))")
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -19,14 +19,6 @@ inputs:
     required: true
     type: string
 
-  library-namespace:
-    description: >
-      Namespace of the Python library. The namespace is the path used when
-      importing the library in Python. The library requires to have a
-      ``__version__`` variable at its top ``__init__.py`` file.
-    required: true
-    type: string
-
   operating-system:
     description: >
       Name of the operating system where the library is installed.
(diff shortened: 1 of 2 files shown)


build-wheelhouse/action.yml
```diff
@@ -89,7 +81,8 @@ runs:
     - name: "Verify library is properly installed and get its version number"
       shell: bash
       run: |
-        version=$(python -c "from ${{ inputs.library-namespace }} import __version__; print(); print(__version__)" | tail -1 || exit 1 )
+        library_name=${{ inputs.library-name }}
+        version=$(python -c "import importlib.metadata as metadata; print(metadata.version('$library_name'))")
```
RobPasMue, line 85: This will only be allowed from Python 3.8 onwards. Maybe we should specify it in the description.

RobPasMue (changes requested):

build-wheelhouse/action.yml
jorgepiloto: You are right. Let me fix this by installing the `importlib-metadata` package if `inputs.python-version` is lower than Python 3.8. Also, I will update the version retrieval command to be compliant with these new changes.

RobPasMue (approved):
LGTM now @jorgepiloto! Thanks!

germa89:
This is an API change, I would probably bump to v5.0 instead of adding anything to v4.0. But up to you guys.

RobPasMue:
But it is not a braking change. Keep in mind that 5.0 is a major version. It is still backwards compatible. It won't break your existing workflows, it will just throw a warning saying your action does not need that input 

See https://semver.org/#summary

RobPasMue:
Bottom line: it is not an incompatible API change. Your action definition will still run just fine. You will just get a warning saying that a certain input is not accepted or something like that. I've seen it before.

germa89:
oh... I thought if you remove the argument from the action, it will prompt an error....  Nevermind then.

PR #287, opened 2023-07-19
feat: add check-licenses to wheelhouse

### LINKED ISSUE #203: Force the license check action in the wheelhouse action
💡 **Idea**

Let us include the check of licenses as a step in the wheelhouse action.

RobPasMue:
Closes #203

### DIFF
--- a/check-licenses/action.yml
+++ b/check-licenses/action.yml
@@ -59,12 +59,29 @@ inputs:
     default: ''
     type: string
 
+  skip-install:
+    description: >
+      Whether to skip the installation of the project. The default is ``false``.
+      If used in combination with other actions, it might not be needed to install
+      the project.
+    default: false
+    required: false
+    type: boolean
+
+  checkout:
+    description: >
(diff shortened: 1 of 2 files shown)


check-licenses/action.yml
```diff
@@ -59,12 +59,29 @@ inputs:
...
+    default: true
+    required: false
+    type: boolean
 
 runs:
   using: "composite"
   steps:
     - name: "Install Git and clone project"
       uses: actions/checkout@v3
+      if: ${{ inputs.checkout == 'true' }}
```
jorgepiloto, line 84: I think this should be:

```suggestion
      if: inputs.checkout == 'true'
```

jorgepiloto (approved):

check-licenses/action.yml
RobPasMue: I think you can do it either way. We've got other actions doing it this way too. Let me merge and check if it is doing it properly

PR #302, opened 2023-07-31
docs: add warning admonition

### LINKED ISSUE #290: Add a deprecation warning in v4.0 because of the project name auto detection.
This will require to add a deprecation warning in v4.0.

_Originally posted by @jorgepiloto in https://github.com/ansys/actions/issues/280#issuecomment-1641442872_

jorgepiloto:
Fixes #290 by introducing a warning in the documentation. Since we removed the ``library-name`` input parameter, it is no longer possible to reference it in the workflow. Thus, we can not check if the user just passed it. GitHub is the one taking care of this warning.

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -7,6 +7,11 @@ description: >
   This allows for local installations and avoids the need to connect to the
   online PyPI index.
 
+  .. warning::
+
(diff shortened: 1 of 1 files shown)


RobPasMue (approved):
This warning will be raised by the workflow anyway, but it's fine to add it to the documentation :)

PR #313, opened 2023-08-02
feat: add reuse and order ignored packages

RobPasMue:
As title says


RobPasMue:
``reuse`` has an APACHE license so it works for us

Revathyvenugopal162 (approved):

AlejandroFernandezLuces (approved):

PR #320, opened 2023-08-08
fix: doc-deploy-index in CI/CD

jorgepiloto:
Fixes the issue raised in https://github.com/ansys/actions/actions/runs/5782950550/job/15670887761


RobPasMue (approved):
Thanks @jorgepiloto

PR #321, opened 2023-08-09
feat: adding ansys corba to ignored licenses

germa89:
The package [ansys-corba](https://github.com/ansys/pymapdl-corba) is currently archived (only-read) and it does not have a valid license (no license at all).

I think, we should add it to the exceptions. Or unarchive it and add a license. I don't mind either.

Reason for this PR:

https://github.com/ansys/pymapdl/actions/runs/5799747218/job/15720419673?pr=2234#step:2:3884

Pinging @jorgepiloto @RobPasMue

### DIFF
--- a/check-licenses/ignored-packages.txt
+++ b/check-licenses/ignored-packages.txt
@@ -7,3 +7,4 @@ reuse
 typing
 typing_extensions
 typing-extensions
+ansys-corba


RobPasMue (approved):
Thanks @germa89 - adding it is fine since it is already archived =)

germa89:
@RobPasMue can we get a patch release soon?

RobPasMue:
These changes go in right away. We are checking the "main" version of the accepted licenses, not the ones on each released version. See: 

https://github.com/ansys/actions/blob/a0cf1dc32aa06fd70909818c669f35c1236aefa3/check-licenses/action.yml#L128-L133

RobPasMue:
It's also a convenience mechanism. There is no need to release in case we are just adding a new accepted package. That'd be an overkill

germa89:
Clever idea indeed! Perfect! Thank you!

PR #322, opened 2023-08-09
fix: tag trigger in workflow

jorgepiloto:
Fixes the issues pointed out by @Revathyvenugopal162. This ensures that only for tags following `v*.*.*` the CI/CD pipelines get triggered:

- Fails because multi-version does not ingest `vX`, see https://github.com/ansys/actions/actions/runs/5795010942
- Generates draft PRs because some candidates were already released

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


RobPasMue (approved):
Sounds good! =)

Revathyvenugopal162 (approved):

PR #323, opened 2023-08-09
fix: add additional options to pymeilisearch

Revathyvenugopal162:
Add stop urls as sphinx options

### DIFF
--- a/doc-deploy-index/action.yml
+++ b/doc-deploy-index/action.yml
@@ -77,8 +77,8 @@ inputs:
 
   stop-urls:
     description: >
-      URLs to stp scraping using `pymeilisearch <https://pymeilisearch.docs.ansys.com/version/stable/>`_
-      while scraping. Default is ``None``.
+      A list of stop points when scraping URLs. If specified,
+      crawling will stop when encountering any URL containing any of the strings in this list.
     default: ''
     required: false
     type: string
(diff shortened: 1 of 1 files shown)


doc-deploy-index/action.yml
```diff
@@ -75,10 +75,10 @@ inputs:
...
     type: string
 
-  pymeilisearchopts:
+  stop-urls:
     description: >
-      Set of options to pass to `pymeilisearch <https://pymeilisearch.docs.ansys.com/version/stable/>`_
-      while scraping. Default is ``None``.
+      A list of stop points when scraping URLs. If specified,
+      crawling will stop when encountering any URL containing any of the strings in this list.
     default: ''
```
RobPasMue, line 82: Wouldn't it be better to just allow all the peimiliesarch possible options like it was done previously?

RobPasMue (changes requested):
LGTM but I have my concerns... see comment

doc-deploy-index/action.yml
jorgepiloto: You are right, Roberto. I did not knew this input was available. It looks like it is not being used in the final command.

Let us go with this input and enable it in the command, @Revathyvenugopal162.

Revathyvenugopal162: Sure revert as options and added it in command

RobPasMue (approved):
Way better now! :)

jorgepiloto:
I will cherry-pick the latest two fixes and perform a new patch release.

PR #328, opened 2023-08-22
fix: run linkcheck on xvfb-run for doc-build action

RobPasMue:
As ttile says - we were missing it when running it with ``xvfb-run``

Also, linkcheck should be the first run - even before html or pdf


RobPasMue:
This will require a patch release - planning on it as soon as it's merged

Revathyvenugopal162 (approved):

PR #329, opened 2023-08-23
fix: doc version page generator

jorgepiloto:
Fix the issues raised in https://github.com/ansys/actions/actions/runs/5948898594/job/16133718203


jorgepiloto:
Just moved to `awk` instead of `sed`. Working as expected.

PR #330, opened 2023-08-25
feat: adding a new vulnerability detection action

RobPasMue:
As title says.

### DIFF
--- a/check-vulnerabilities/action.yml
+++ b/check-vulnerabilities/action.yml
@@ -1,7 +1,8 @@
 name: "Check library and third-party vulnerabilities."
 description: >
   Action actively checking for library and third party vulnerabilities by
-  means of the `bandit <>`_ and `safety <>`_ Python packages. These packages
+  means of the `bandit <https://bandit.readthedocs.io/en/latest/>`_ and
+  `safety <https://pyup.io/safety/>`_ Python packages. These packages
   allow you to identify security issues and vulnerabilities inside your code.
 
   This action has two running modes:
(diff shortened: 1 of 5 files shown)


check-vulnerabilities/action.yml
```diff
@@ -0,0 +1,390 @@
...
+                pass
+
+
+        EOF
+        cat dependency-check.py
+
+    - name: "Run safety advisory checks"
+      shell: bash
+      run: |
+        # python dependency-check.py
```
RobPasMue, line 389: Will be activated once the action is reviewed and validated

check-vulnerabilities/action.yml
```diff
@@ -0,0 +1,390 @@
...
+    required: false
+    type: string
+
+runs:
+  using: "composite"
+  steps:
+    - name: "Declare needed environment variables"
+      shell: bash
+      run: |
+        echo "DEPENDENCY_CHECK_TOKEN=${{ inputs.token }}" >> $GITHUB_ENV
```
jorgepiloto, line 90: I am worried about the secret nature of the token getting removed by copying its value to an environment variable.

Could we use "INPUT_TOKEN"? This one should be available, see https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#example-specifying-inputs

jorgepiloto (changes requested):

check-vulnerabilities/action.yml
RobPasMue: Yep, definitely! :)

jorgepiloto (approved):

PR #340, opened 2023-09-12
fix: sort ignored packages and add WMI

jorgepiloto:
This package is required by one of our libraries. This library is still under development. The WMI package is using an MIT, see https://pypi.org/project/WMI/. However, because it does not ship with a LICENSE file, it needs to be manually ignored.

```console
Invalid licenses found:

WMI 1.4.9 has http://www.opensource.org/licenses/mit-license.php license

 Name                                   Version   License                                            
 WMI                                    1.4.9     http://www.opensource.org/licenses/mit-license.php 
```

### DIFF
--- a/check-licenses/ignored-packages.txt
+++ b/check-licenses/ignored-packages.txt
@@ -1,3 +1,4 @@
+ansys-corba
 defusedxml
 filelock
 matplotlib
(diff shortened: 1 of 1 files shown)


RobPasMue:
Thanks @jorgepiloto!

Just so you know. If you add ignored packages there is no need to do a release. We are looking into the main branch's file for this operation.

PR #343, opened 2023-09-29
fix: bug in wheelhouse generation

RobPasMue:
Reported by @greschd - related to https://github.com/ansys/pyansys/discussions/373#discussioncomment-7141170

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -134,7 +134,7 @@ runs:
     - name: "Generate the wheelhouse"
       shell: bash
       run: |
-        python -m pip wheel . -w wheelhouse
+        python -m pip wheel ${{ env.install_target }} -w wheelhouse
 
     - name: "Compress the wheelhouse"
       uses: vimtor/action-zip@v1.1


RobPasMue:
This will require a patch release

build-wheelhouse/action.yml
```diff
@@ -134,7 +134,7 @@ runs:
     - name: "Generate the wheelhouse"
       shell: bash
       run: |
-        python -m pip wheel . -w wheelhouse
+        python -m pip wheel ${{ env.install_target }} -w wheelhouse
```
jorgepiloto, line 137: This ensures that if no optional deps are declared, the project gets installed. Also, since this is an env var, it may be better to use upper case:

```suggestion
        python -m pip wheel .[${{ env.OPTIONAL_DEPENDENCIES }}] -w wheelhouse
```

Note that developers can pass multiple optional deps, like "OPTIONAL_DEPENDENCIES: 'doc,tests'".

jorgepiloto (changes requested):

build-wheelhouse/action.yml
RobPasMue: @jorgepiloto - this is handled here already

https://github.com/ansys/actions/blob/7988e4d1644a4077e88eaa92fe99d5dfb41c6e99/build-wheelhouse/action.yml#L84-L89

RobPasMue: It's not a real env var, that's why I defined it in small caps in the past. It's coming from an input.

jorgepiloto: Oh! My fault. Approving this.

jorgepiloto (approved):

PR #353, opened 2023-10-16
feat: adding step for importing library

### LINKED ISSUE #347: Library import when smoke tests/wheel build
I think we should explicitly import the library during the smoke testing step. Maybe we are importing it when using `importlib_metadata` but I'm not sure/it is not clear.

Like in here: https://github.com/ansys/pymapdl/pull/2399

germa89:
Add an explicit step for testing the import of the library.

Close #347 

Improve #276

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -92,6 +92,18 @@ runs:
       run: |
         python -m pip install ${{ env.install_target }}
 
+    - name: "Verify library is properly installed"
+      shell: bash
+      run: |
+        library_name=${{ inputs.library-name }}
+
+        if python -c "import $library_name"; then
+            echo "Python module '$library_name' successfully imported."
+        else
+            echo "Failed to import Python module '$library_name': Something went wrong."
+            exit 1
+        fi
+
     - name: "Verify if importlib-metadata needs to be installed"
       shell: bash
       run: |
@@ -112,7 +124,8 @@ runs:
       run: |
         python -m pip install importlib-metadata
 
-    - name: "Verify library is properly installed and get its version number"
+
+    - name: "Get library version number"
       shell: bash
       run: |
         library_name=${{ inputs.library-name }}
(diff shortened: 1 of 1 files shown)


build-wheelhouse/action.yml
```diff
@@ -92,6 +92,18 @@ runs:
...
+      shell: bash
+      run: |
+        library_name=${{ inputs.library-name }}
+
+        if python -c "import $library_name"; then
+            echo "Python module '$library_name' successfully imported."
+        else
+            echo "Failed to import Python module '$library_name': Something went wrong."
+            exit 1
+        fi
```
RobPasMue, line 105: Not sure this would work... Your library name is ``ansys-mapdl-core`` but your import statement should be ``import ansys.mapdl.core as pymapdl`` for example

RobPasMue (changes requested):

build-wheelhouse/action.yml
germa89: veeeery good point.

germa89: We use to have the other library name for that.... It's ok :)

germa89: Fixed now in here: https://github.com/ansys/actions/pull/353/commits/b5b4a3eeab7f3c4c279b005a8618c4002b14560b


Code for local testing:

```bash
library_name="asdf-QERQER-asdf.asdfas"

library_name="${library_name//-/.}"
echo "Modified the library name for importing: '$library_name'"
```

build-wheelhouse/action.yml
```diff
@@ -92,6 +92,21 @@ runs:
       run: |
         python -m pip install ${{ env.install_target }}
 
+    - name: "Verify library is properly installed"
+      shell: bash
+      run: |
+        library_name=${{ inputs.library-name }}
+
+        library_name="${library_name//-/.}"
```
RobPasMue, line 100: I'm not sure if this will always work. 90% of the cases it will... but I'm guessing some will fail

germa89: I kind of agree.. but I cannot come up with something better 🤷🏻‍♂️

jorgepiloto:
Closing this as it is no longer required.

PR #357, opened 2023-10-25
feat: add changelog fragment action

klmcadams:
Creates a fragment based on the title of the pull request, and pushes the fragment to the PR branch. 

- Checks if `inputs.pr-title` has double quotes around it. If not, it adds the quotes
- Checks the PR titles for keywords: fix, add, change, remove, bump (for dependencies). If it does not start with a keyword, it fails the action and says to change the title
- Removes any fragments that exist in the PR containing the PR number. For example, if the PR number was 10, it searches for files containing "10." since the fragment format would be 10.added.md, or whichever other keyword
- Creates the fragment with the pull request title, number, and keyword. 
      `towncrier create -c "my title" 10.added.md`
- If the file was modified, it adds the file and pushes it to the pull request's branch. Otherwise, it does nothing

To do:
- Figure out if there's a way to only run this action when the PR is first created, and any time the PR title is updated


doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
+name: >
+  Documentation Changelog
+
+description: |
+  Adds a newsfragment to the target repository using ``towncrier``
+  containing the title and number of the pull request.
```
germa89, line 6: It will be nice to add an example here or a link to the documentation where you can see an actual implementation of it.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+
+        # Save title into env variable
+        echo "PR-TITLE=$title" >> $GITHUB_ENV
+
+        # Check if pr-title string starts with Add, Remove, Change, Fix, or Bump
+        added=`if [[ $title =~ ^[a,A]dd ]]; then echo "True"; fi`
+        removed=`if [[ $title =~ ^[r,R]emove ]]; then echo "True"; fi`
+        changed=`if [[ $title =~ ^[c,C]hange ]]; then echo "True"; fi`
+        fixed=`if [[ $title =~ ^[f,F]ix ]]; then echo "True"; fi`
+        dependency=`if [[ $title =~ ^[b,B]ump ]]; then echo "True"; fi`
```
germa89, line 82: Ok.... shall we really use the PR title?? the PR branch name?? or the PR labels?? Being aware the labels tends to be applied by the bot in many cases.

I'm in for the labels. It is easy to change... many times you start a branch with a purpose and ends for something else.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+        elif [ ! -z "$changed" ]; then
+          echo "PR-TYPE=changed" >> $GITHUB_ENV
+        elif [ ! -z "$fixed" ]; then
+          echo "PR-TYPE=fixed" >> $GITHUB_ENV
+        elif [ ! -z "$dependency" ]; then
+          echo "PR-TYPE=dependencies" >> $GITHUB_ENV
+        else
+          echo "The pull request title ${{ inputs.pr-title }} is invalid. Please ensure it starts with one of the following keywords:"
+          echo  "Add, Remove, Change, Fix, or Bump (for dependencies)."
+          exit 1
```
germa89, line 98: yeah... I prefer to not enforce titles.. but labels. But this is a matter of taste, I'm open to discuss it or consider other options.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+    - name: "Create and commit towncrier fragment"
+      shell: bash
+      run: |
+        file="${{ inputs.pr-number }}.${{ env.PR-TYPE }}.md"
+        changelog_file="${{ inputs.path-to-fragment }}/$file"
+
+        # If fragment already exists with ${{ inputs.pr-number }}, remove it and create a new one
+        if [[ "$changelog_file" == *"${{ inputs.pr-number }}."* ]]; then
+          rm ${{ inputs.path-to-fragment }}/${{ inputs.pr-number }}.*
+        fi
```
germa89, line 110: What if I'm doing my fragment in each PR, because maybe I want something more customizable or similar.

I would attach the one that is already in place.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+
+        # Create changelog fragment with towncrier
+        towncrier create -c "${{ env.PR-TITLE }}" $file
+
+        # Configure git username & email
+        git config user.name 'github-actions[bot]'
+        git config user.email 'github-actions[bot]@users.noreply.github.com'
+
+        modified=`git diff HEAD --name-only`
+        echo "modified: $modified"
```
germa89, line 120: Good call!

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+        else
+          echo "MODIFIED=False" >> $GITHUB_ENV
+        fi
+
+    - name: "Push fragment to PR branch"
+      if: ${{ env.MODIFIED }} == True
+      uses: ad-m/github-push-action@master
+      with:
+        github_token: ${{ inputs.token }}
+        branch: ${{ inputs.pr-branch }}
```
germa89, line 138: You could do all of this in one step:

```yml
      - name: "Attaching modified files to PR"
        id: attatch-to-pr
        uses: EndBug/add-and-commit@v9
        with:
          message: Update the image cache
          committer_name: GitHub Actions
          committer_email: actions@github.com
          add: './tests/.image_cache/*.png'
```

germa89 (changes requested):
This is an interesting and useful PR.

I'm not sure if I am grasping all the context needed for a proper review (I had few questions about its intended use), but Github seems I should review it... so yeah.. here we go. I'm aware this is a draft by the way.

I think overall looks good. I have several comments, and now a general one:

* Commit and pushing using a `GITHUB_TOKEN` does not trigger CICD, that is why I had to add myself a comment in the PR: https://github.com/ansys/pymapdl/blob/ffa3d294e87b07df874936b51fc570751c2a4368/.github/workflows/ci.yml#L551-L552
  Presumably, the fragment does not require CI. But it will still show the actions as "waiting" ... I think. Probably I'm wrong.

doc-changelog/action.yml
germa89: I'm not sure about what email to use... if you are happy with yours, keep yours.

doc-changelog/action.yml
jorgepiloto: You can create a new example for this action by creating a file named `doc-changelog-basic.yml` inside the `doc/source/doc-actions/examples/` directory, @klmcadams.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+
+inputs:
+
+  # Required inputs
+
+  path-to-fragment:
+    description: >
+      Path to where the towncrier fragment is stored
+    required: true
+    type: string
```
jorgepiloto, line 17: I think this is not required. Projects using this action should configure `towncrier` in their `pyproject.toml`. This is where the fragments directory gets specified. This ensures that running any `towncrier <commnad>` renders a fragment in the desired dir.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+
+  pr-branch:
+    description: >
+      Name of the pull request
+    required: true
+    type: string
```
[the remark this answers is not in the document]
jorgepiloto: That's a good point, @germa89. Could we check if the environment gets shared with the reused actions, @klmcadams? If so, we could skip these inputs.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: "Install Git and clone project"
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ inputs.pr-branch }}
```
jorgepiloto, line 49: No need to indicate the branch. By default, this action clones the branch in the commit that triggered the workflow.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+
+    - name: "Set up Python ${{ inputs.python-version }}"
+      uses: ansys/actions/_setup-python@main
+      with:
+        python-version: ${{ inputs.python-version }}
+        use-cache: ${{ inputs.use-python-cache }}
```
jorgepiloto, line 55: This input is missing.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+      shell: bash
+      run: |
+        # Check inputs.pr-title is surrounded by double quotes
+        title='${{ inputs.pr-title }}'
+        if [[ "$title" =~ ^\".*\"$ ]]; then
+          echo "PR title is wrapped in double quotes"
+        else
+          echo "PR title is not wrapped in double quotes"
+          title="${{ inputs.pr-title }}"
+        fi
```
jorgepiloto, line 72: Could you give some information on why this check is required?

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+
+        # Save title into env variable
+        echo "PR-TITLE=$title" >> $GITHUB_ENV
+
+        # Check if pr-title string starts with Add, Remove, Change, Fix, or Bump
```
jorgepiloto, line 77: For this checks, we may consider using https://github.com/ansys/actions/blob/release/4.1/commit-style/action.yml. This way, we enforce pull-requests to follow conventional commits.

doc-changelog/action.yml
```diff
@@ -0,0 +1,138 @@
...
+        else
+          echo "The pull request title ${{ inputs.pr-title }} is invalid. Please ensure it starts with one of the following keywords:"
+          echo  "Add, Remove, Change, Fix, or Bump (for dependencies)."
+          exit 1
+        fi
+
+    - name: "Create and commit towncrier fragment"
+      shell: bash
+      run: |
+        file="${{ inputs.pr-number }}.${{ env.PR-TYPE }}.md"
```
jorgepiloto, line 104: Let's use towncrier to create the fragment. This ensures that users configuration applies during the whole process:

```console
towncrier create --content 'Can also be ``rst`` as well!' 7890.doc.rst
```

doc-changelog/action.yml
jorgepiloto, line 138: Is this really required? You already configure the Git environment with a user. Running the usual `git push` should be enough without needing third party actions.

jorgepiloto (changes requested):
This is looking good, I can see its power. Some changes are required though.

doc-changelog/action.yml
klmcadams: If double quotes aren't wrapped around the title, then the towncrier command isn't going to generate correctly

`towncrier create -c my message prnumber.added.md` vs. `towncrier create -c "my message" prnumber.added.md`

I think it throws an error if the title isn't surrounded by quotes

doc-changelog/action.yml
klmcadams: That would be great! I didn't know that existed

doc-changelog/action.yml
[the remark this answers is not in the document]
klmcadams: @germa89 @jorgepiloto Oh yeah, I must have forgotten to make a new description. The description should be the name of the PR branch. The reason I had those inputs was because I didn't know that you could do ${{ github... }} within the action, and that it would refer to the context of the branch it was being run on

This is the ci_cd file I was using to test the action: https://github.com/ansys-internal/actions-sandbox/blob/15e61871243b198706aae00c8eb4bf9dd90bd6db/.github/workflows/ci_cd.yml#L29

I think it would be nice if this only ran on PRs when they are first created, and any time the title is changed. I'm not sure if there's a title change event but I can do more research. Is there a way to specify within the action to only run in certain scenarios?

doc-changelog/action.yml
klmcadams: I think it would be best to make sure the PR title follows certain naming conventions (like in this action https://github.com/ansys/actions/blob/release/4.1/commit-style/action.yml) because if someone names it incorrectly, they can just fix it and re-run the action

doc-changelog/action.yml
klmcadams: Sure I can use mine

doc-changelog/action.yml
klmcadams: I think this is intended for only generating fragments in PRs. I was following these docs for setting up the towncrier command

https://towncrier.readthedocs.io/en/stable/markdown.html

This line `towncrier create -c "Added a cool feature!" 1.added.md` would result in this section in the changelog after doing towncrier build. 

```
### Added

- Added a cool feature! ([#1](https://github.com/twisted/my-project/issues/1))
```

I was taking in the pr-number, the action of the PR in the title like add, fix, etc., and the title of the PR for the phrase that comes before the link to the PR so then it would generate the fragment similar to how it looks when we manually update the changelog (at least in pymechanical).

In the above example, if 1.added.md already existed in the branch, then it would create another file named 1.added.1.md (or something like that), which would then in the end add another line to the changelog with the same PR number. To get around that, I was deleting any fragment that had the same number, so there would only be one fragment per PR number

doc-changelog/action.yml
klmcadams: I think I do use towncrier here: https://github.com/ansys/actions/blob/513e6bfdc7a73057009bc4bc0c4ab32aad246081/doc-changelog/action.yml#L113

I was following the file name format in this tutorial: https://towncrier.readthedocs.io/en/stable/markdown.html

doc-changelog/action.yml
klmcadams: I'm not sure! I'll try git push instead of this

RobPasMue:
@klmcadams - what's the status with this PR¿?

klmcadams:
@RobPasMue I haven't had a chance to work on it, but I can after I'm done with the pre-commit hook PR

PR #360, opened 2023-11-03
chore: bump doc-style vale version (2.29.6)

SMoraisAnsys:
Multiple repos use the ansys/doc-style action without providing specific inputs. Since using "2.29.6" and updating accepts.txt file fixes the warning many developers faced, I would recommend "2.29.6" to be the default version used.


RobPasMue (approved):
Let's go!

RobPasMue:
Changing PR name to follow conventional commits convention

PR #363, opened 2023-11-14
feat: Add requirements file custom path in doc building

AlejandroFernandezLuces:
## Overview
 
This PR adds an option to specify the location of the `requirements.txt` dependency file in the doc building action, since in some cases, we are building documentation in non-python projects and we want to avoid having a requirements file outside of the scope where it's needed.

### DIFF
--- a/doc-build/action.yml
+++ b/doc-build/action.yml
@@ -82,7 +82,7 @@ inputs:
 
   requirements-file:
     description: >
-      Path to the requirements file in case it needs to be in an specific location.
+      Path to the requirements file in case it needs to be in a specific location.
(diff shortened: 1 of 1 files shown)


doc-build/action.yml
```diff
@@ -80,6 +80,15 @@ inputs:
     required: false
     type: boolean
 
+  requirements-file:
+    description: >
+      Path to the requirements file in case it needs to be in an specific location.
```
RobPasMue, line 85:
```suggestion
      Path to the requirements file in case it needs to be in a specific location.
```

RobPasMue (approved):
LGTM! Doing patch release right away

PR #366, opened 2023-11-27
fix: add option to skip-existing in PyPI releases

jorgepiloto:
As per @greschd request, this pull-request introduces the ability to skip or not an existing artifact in a PyPI index.


_release-pypi/action.yml
```diff
@@ -52,6 +52,11 @@ inputs:
     required: true
     type: boolean
 
+  skip-existing:
+    description: >
+      Continue uploading files if one already exists. Only valid when uploading
+      to PyPI. Other implementations may not support this.
+
```
klmcadams, line 59: Does this need a required or type value like the previous inputs?

RobPasMue: Yes it does. Please add it as a boolean @jorgepiloto

RobPasMue (changes requested):

RobPasMue (approved):

MaxJPRey (approved):
LGTM.

PR #377, opened 2023-12-20
chore: update upload-artifacts to v4

germa89:
As the title.


jorgepiloto (approved):
Thanks for fixing this, @germa89.

Issue #381, opened 2024-01-03
Add Dockerfile linting


SMoraisAnsys:
### 📝 Description of the feature

After using `hadolint` on a project, I was reminded of some good practices when writing a Dockerfile.
Adding this lint to ansys repos should be of great help to increase the quality of our products.

After some discussion with @RobPasMue the idea would be to:
- add a step to the `code-style` action with some checking on the directory to check
- create an independent public action as `code-style` action gets removed in favor of using pre-commit.ci.

Note: the action can either leverage [hadolint-action](https://github.com/hadolint/hadolint-action) or [hadolint-gh-action](https://github.com/jbergstroem/hadolint-gh-action).

### 💡 Steps for implementing the feature

We have to be carefull of the way we want to discover and check multiple dockerfiles as `hadolint-action` and `hadolint-gh-action` do not work in the same way when trying to deal with multiple dockerfiles.

RobPasMue:
My suggestion would be to:

* Add the checks for the ``docker`` folder inside the wrapped hadolint-action/hadolint-gh-action (e.g. ``docker-style``?).
* From the ``code-syle`` action, simply call the ``docker-style`` action
* The ``docker-style`` action should verify that a user has (at least) a "docker" folder that contains the repository's Dockerfiles. Dockerfiles shouldn't be placed anywhere else. Then, you simply run the hadolint-* action with some predefined inputs to ensure consistency across our repos.

SMoraisAnsys:
Closed by #382

PR #382, opened 2024-01-03
feat: add dockerfile linting

SMoraisAnsys:
This PR addresses #381 by :
1. adding an extra action ("docker-style") that can be used to evaluate the Dockerfile(s) associated to a repo;
2. extending action "code-style" by triggering "docker-style" on "docker/Dockerfile" and ".devcontainer/Dockerfile" if the associated directories exist.

It was tested and validated in a sandbox. If you are interested, you can see the CI jobs failing and succeeding when expected (https://github.com/SMoraisAnsys/pyansys-geometry/actions/runs/7628030247/job/21173613810?pr=1).

Here is a summary of the changes:

TLDR
------
- docker-style looks for files named "Dockerfile" in the provided directory (default to "docker"). If one wants to propose and evaluate multiple docker files, then the associated repo must follow a directory oriented architecture (e.g. docker/linux/Dockerfile and docker/windows/Dockerfile) and option `recursive` must be set to `true`.
- docker-style fails based on `hadolint` output and one can define at which ouput level the action fails through option `error-level`.
- code-style can be extend to perform docker linting on "docker" and ".devcontainer" directories through option `docker-lint` which default value is False to not introduce breaking changes.
- code-style docker linting can be tunned through options `docker-recursive` and `docker-error-level`.

docker-style
--------------

This action leverages the action [jbergstroem/hadolint-gh-action](https://github.com/jbergstroem/hadolint-gh-action) and requires a directory to look for the docker file. This directory can be anything but we emphasize to store docker files in a directory named "docker" at the root of the repository. If no "docker" directory exists, this action will raise a warning through the CI.

If one wants to evaluate multiple docker files nested in the provided directory, option `recursive` must be set to `True` and :warning: each file expected to be evaluated must be named Dockerfile. Notice that this constraints your repo to follow a directory oriented architecture, i.e. to split the Dockerfiles into multiple directories (e.g. docker/linux/Dockerfile and docker/windows/Dockerfile).

By default the action has a strict error level based on the outputs of [hadolint](https://github.com/hadolint/hadolint). Any info caught into this ouput will make the action failt. It can be lightened by changing the value of the option `error-level`. The associated value can be -1, 0, 1 or 2 where:
- -1: never fail;
- 0: fail on errors;
- 1: fail on warnings;
- 2: fail on infos.

code-style
------------

This action was modified to allow users to integrate docker linting (using previous docker-style action). By default this is not activated but one can activate it with option `docker-lint` set to True. Directory to be linted are set to "docker" and ".devcontainer", and the options used are `recursive=false` and `error-level=2`.
The options passed to the action "docker-style" can be modified through the new options `docker-recursive` and `docker-error-level`.

Notes
------

Github `working-directory` cannot be used with `uses` (I tried to use it to specify where to look for docker files).
At first we used [hadolint-action](https://github.com/hadolint/hadolint-action) but is wouldn't work as wanted when looking for multiple docker files. For example, if we wanted to look for multiple docker files, we would need to use the recursive option. However, if we wanted to look into an empty directory "docker" while having docker files in a directory "some_other_directory",  then the action would evaluate the docker files inside of "some_other_directory". Indeed, option `recursive` searches for specified dockerfile recursively, **from the project root**.

We also had to fix `hadolint-gh-action` issue through (https://github.com/jbergstroem/hadolint-gh-action/pull/135)

### DIFF
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -29,7 +29,7 @@ inputs:
 
   docker-recursive:
     description: >
-      Search for Dockerfile(s) recursively. Only applies to directory 'docker'.
+      Search for Dockerfile(s) recursively. Only applies to allowed directories, i.e. docker and .devcontainer.
     required: false
     default: false
     type: bool
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -5,7 +5,8 @@ description: >
   This action evaluates the code quality of your project by using `pre-commit
   <https://github.com/pre-commit/pre-commit/>`_. The action installs and runs
   ``pre-commit``. It is assumed that your project contains a
-  ``.pre-commit-config.yaml`` file in the root directory.
+  ``.pre-commit-config.yaml`` file in the root directory. The action can also
+  be extended to lint docker files.
 
 inputs:
 
@@ -48,6 +49,13 @@ inputs:
     default: 2
     type: int
 
+  docker-directory:
+    description: >
+      Directory from which to search for Dockerfile(s).
+    required: false
+    default: docker
+    type: string
+
 runs:
   using: "composite"
   steps:
@@ -73,91 +81,29 @@ runs:
 
     # ------------------------------------------------------------------------
 
-    # The following steps are applying docker-style action to allowed working dirs if they exist.
-    # Changes performed should coherent with the 'allowed_working_dirs' in the docker-style action.
+    # Trigger docker-style action if docker-lint is set to true.
     - uses: ansys/actions/_logging@main
+      if: ${{ inputs.docker-lint == 'true' }}
       with:
         level: "INFO"
         message: >
-          Extend the workflow to lint Dockerfile(s) if option docker-lint is
-          set to true, exit otherwise.
-
-    - name: "Exit if docker-lint is false"
-      shell: bash
-      run: |
-        if [[ "${{ inputs.docker-lint }}" == "true" ]]; then
-          echo "RUN_DOCKER_LINT=true" >> $GITHUB_ENV
-        fi
-
-    # ------------------------------------------------------------------------
-
-    - uses: ansys/actions/_logging@main
-      if: env.RUN_DOCKER_LINT == 'true'
-      with:
-        level: "INFO"
-        message: >
-          Check that directory 'docker' exists at the root of the repository.
-          If it does not, then it means that the rest of the action should not
-          be performed.
-
-    - name: "Check if docker directory exists"
-      if: env.RUN_DOCKER_LINT == 'true'
-      shell: bash
-      run: |
-        if [ -d "${{ github.workspace }}/docker" ]; then
-          echo "HAS_DOCKER_DIR=true" >> $GITHUB_ENV
-        else
-          echo "HAS_DOCKER_DIR=false" >> $GITHUB_ENV
-        fi
-
-    # ------------------------------------------------------------------------
-
-    - uses: ansys/actions/_logging@main
-      if: env.HAS_DOCKER_DIR == 'true' && env.RUN_DOCKER_LINT == 'true'
-      with:
-        level: "INFO"
-        message: >
-          Lint directory '${{ github.workspace }}/docker'.
+          Extend the workflow to lint Dockerfile(s) in
+          '${{ github.workspace }}/docker'.
 
     - name: "Run Dockerfile linting in docker directory"
-      if: env.HAS_DOCKER_DIR == 'true' && env.RUN_DOCKER_LINT == 'true'
-      uses: ansys/actions/docker-style@main
+      if: ${{ inputs.docker-lint == 'true' }}
+      uses: ansys/actions/docker-style@feat/lint_dockerfile
       with:
-        directory: docker
+        directory: ${{ inputs.docker-directory }}
         recursive: ${{ inputs.docker-recursive }}
         error-level: ${{ inputs.docker-error-level }}
 
     # ------------------------------------------------------------------------
 
+    # End action as docker-lint is set to false.
     - uses: ansys/actions/_logging@main
-      if: env.RUN_DOCKER_LINT == 'true'
+      if: ${{ inputs.docker-lint == 'false' }}
       with:
         level: "INFO"
         message: >
-          Check if directory '.devcontainer' exists.
-
-    - name: "Check if .devcontainer directory exists"
-      if: env.RUN_DOCKER_LINT == 'true'
-      shell: bash
-      run: |
-        if [ -d "${{ github.workspace }}/.devcontainer" ]; then
-          echo "HAS_DEVCONTAINER_DIR=true" >> $GITHUB_ENV
-        else
-          echo "HAS_DEVCONTAINER_DIR=false" >> $GITHUB_ENV
-        fi
-
-    # ------------------------------------------------------------------------
-
-    - uses: ansys/actions/_logging@main
-      if: env.HAS_DEVCONTAINER_DIR == 'true' && env.RUN_DOCKER_LINT == 'true'
-      with:
-        level: "INFO"
-        message: >
-          Lint directory '${{ github.workspace }}/.devcontainer'.
-
-    - name: "Run Dockerfile linting in .devcontainer directory"
-      if: env.HAS_DEVCONTAINER_DIR == 'false' && env.RUN_DOCKER_LINT == 'true'
-      uses: ansys/actions/docker-style@main
-      with:
-        directory: .devcontainer
-        error-level: ${{ inputs.docker-error-level }}
+          Docker lint disabled.
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -93,7 +93,7 @@ runs:
 
     - name: "Run Dockerfile linting in directories 'docker' and '.devcontainer'"
       if: ${{ inputs.docker-lint == 'true' }}
-      uses: ansys/actions/docker-style@main
+      uses: ansys/actions/docker-style@feat/lint_dockerfile
       with:
         directory: "docker .devcontainer"
         recursive: ${{ inputs.docker-recursive }}
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -85,7 +85,7 @@ runs:
     - name: "Exit if docker-lint is false"
       shell: bash
       run: |
-        if [ ${{ inputs.docker-lint }} == false ]; then
+        if [[ "${{ inputs.docker-lint }}" == "false" ]]; then
           exit 0
         fi
 
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -84,6 +84,7 @@ runs:
           be performed.
 
     - name: "Check if docker directory exists"
+      id: checkdocker
       shell: bash
       run: |
         if [ -d "${{ github.workspace }}/docker" ]; then
@@ -95,7 +96,7 @@ runs:
     # ------------------------------------------------------------------------
 
     - uses: ansys/actions/_logging@main
-      if: env.HAS_DOCKER_DIR == 'false'
+      if: steps.checkdocker.outputs.HAS_DOCKER_DIR == 'false'
       with:
         level: "WARNING"
         message: >
@@ -105,22 +106,22 @@ runs:
           directory at the root of the repository.
 
     - name: "Exit workflow"
-      if: env.HAS_DOCKER_DIR == 'false'
+      if: steps.checkdocker.outputs.HAS_DOCKER_DIR == 'false'
       shell: bash
       run: exit 0
 
     # ------------------------------------------------------------------------
 
     - uses: ansys/actions/_logging@main
-      if: env.HAS_DOCKER_DIR == 'true'
+      if: steps.checkdocker.outputs.HAS_DOCKER_DIR == 'true'
       with:
         level: "INFO"
         message: >
           Lint directory '${{ github.workspace }}/docker'.
 
     - name: "Run Dockerfile linting in docker directory"
+      if: steps.checkdocker.outputs.HAS_DOCKER_DIR == 'true'
       uses: ansys/actions/docker-style@feat/lint_dockerfile
-      if: env.HAS_DOCKER_DIR == 'true'
       with:
         directory: docker
         recursive: ${{ inputs.docker-recursive }}
@@ -136,6 +137,7 @@ runs:
 
     - name: "Check if .devcontainer directory exists"
       shell: bash
+      id: checkdevcontainer
       run: |
         if [ -d "${{ github.workspace }}/.devcontainer" ]; then
           echo "HAS_DEVCONTAINER_DIR='true'" >> $GITHUB_OUTPUT
@@ -146,7 +148,7 @@ runs:
     # ------------------------------------------------------------------------
 
     - uses: ansys/actions/_logging@main
-      if: env.HAS_DEVCONTAINER_DIR == 'false'
+      if: steps.checkdevcontainer.outputs.HAS_DEVCONTAINER_DIR == 'false'
       with:
         level: "INFO"
         message: >
@@ -154,22 +156,22 @@ runs:
           the repository.
 
     - name: "Exit workflow"
-      if: env.HAS_DEVCONTAINER_DIR == 'false'
+      if: steps.checkdevcontainer.outputs.HAS_DEVCONTAINER_DIR == 'false'
       shell: bash
       run: exit 0
 
     # ------------------------------------------------------------------------
 
     - uses: ansys/actions/_logging@main
-      if: env.HAS_DEVCONTAINER_DIR == 'true'
+      if: steps.checkdevcontainer.outputs.HAS_DEVCONTAINER_DIR == 'true'
       with:
         level: "INFO"
         message: >
           Lint directory '${{ github.workspace }}/.devcontainer'.
 
     - name: "Run Dockerfile linting in .devcontainer directory"
+      if: steps.checkdevcontainer.outputs.HAS_DEVCONTAINER_DIR == 'false'
       uses: ansys/actions/docker-style@feat/lint_dockerfile
-      if: env.HAS_DEVCONTAINER_DIR == 'true'
       with:
         directory: .devcontainer
         error-level: ${{ inputs.docker-error-level }}
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -87,9 +87,9 @@ runs:
       shell: bash
       run: |
         if [ -d "${{ github.workspace }}/docker" ]; then
-          echo "HAS_DOCKER_DIR=true" >> $GITHUB_OUTPUT
+          echo "HAS_DOCKER_DIR='true'" >> $GITHUB_OUTPUT
         else
-          echo "HAS_DOCKER_DIR=false" >> $GITHUB_OUTPUT
+          echo "HAS_DOCKER_DIR='false'" >> $GITHUB_OUTPUT
         fi
 
     # ------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


RobPasMue:
Is the PR ready for review? I see it's still in draft mode

code-style/action.yml
```diff
@@ -27,6 +27,20 @@ inputs:
     default: true
     type: boolean
 
+  docker-recursive:
+    description: >
+      Search for Dockerfile(s) recursively. Only applies to directory 'docker'.
```
germa89, line 32: This is false, because you can also analyse `.devcontainer` dir.

code-style/action.yml
```diff
@@ -27,6 +27,20 @@ inputs:
...
+    default: false
+    type: bool
+
+  docker-error-level:
+    description: >
+      Fail docker-style action based on hadolint output (-1: never, 0: error, 1: warning, 2: info)
+    required: false
+    default: 2
+    type: int
+
```
germa89, line 43: I would also add the option
```yml
docker-directories:
   description: >
       Check these directories for docker files and lint them....
...
```

germa89 (changes requested):
I would personally change the approach as per my comments. I'm happy to discuss this further, or compromise if my approach is not valid.
The approach proposed is not bad, I just feel we should be more flexible anticipating that this action will be used in many repos with many different configurations.

Other than that, I like the emphasis given to the documenting part. It makes the action reads very well.

germa89:
> Is the PR ready for review? I see it's still in draft mode

Opsss.... I did it again (I review it)

SMoraisAnsys:
@germa89 Following your feedback, we have to decide on the scope associated with this action.
*Reminder*: we emphasize the use of a `docker` directory where every docker file should be name `Dockerfile`. In the case of [pyansys-geometry](https://github.com/ansys/pyansys-geometry) that would lead to `docker/windows/Dockerfile` and `docker/linux/Dockerfile` instead of `docker/Dockerfile.windows` and `docker/Dockerfile.linux`

From my previous discussions on docker files with other developers, I had the feeling that it would be great that every pyansys project could share the same structure for those files. Currently, one can share docker files in any directory, with whatever name (compatible with docker ofc) they want. I think that it would be clearer for our users if some pattern was also respected for our products. Does anyone has a specific reason to not follow the proposed structure ?

code-style/action.yml
```diff
@@ -49,3 +63,36 @@ runs:
...
       run: pre-commit run --all-files --show-diff-on-failure
+
+    # The followings steps are applying docker-style action to allowed working dirs if the exist.
+    # Changes performed here should coherent with the 'allowed_working_dirs' in the docker-style action.
+    - name: "Check if docker directory exists"
+      shell: bash
+      id: checkdocker
+      run: |
+        if [ -d "${{ github.workspace }}/docker" ]; then
+          echo "exists=true" >> $GITHUB_OUTPUT
```
jorgepiloto, line 74: I would rename this var to:

```suggestion
          echo "HAS_DOCKER_DIR=true" >> $GITHUB_OUTPUT
```

code-style/action.yml
```diff
@@ -49,3 +63,36 @@ runs:
...
+        fi
+
+    - name: "Run Dockerfile linting in docker directory"
+      uses: ansys/actions/docker-style@main
+      if: steps.checkdocker.outputs.exists == 'true'
```
jorgepiloto, line 79: Following on previous suggestion, this would be:

```suggestion
      if: env.HAS_DOCKER_DIR == 'true'
```

code-style/action.yml
```diff
@@ -49,3 +63,36 @@ runs:
...
+        directory: docker
+        recursive: ${{ inputs.docker-recursive }}
+        error-level: ${{ inputs.docker-error-level }}
+
+    - name: "Check if .devcontainer directory exists"
+      shell: bash
+      id: checkdevcontainer
+      run: |
+        if [ -d "${{ github.workspace }}/.devcontainer/Dockerfile" ]; then
+          echo "exists=true" >> $GITHUB_OUTPUT
```
jorgepiloto, line 90:
```suggestion
          echo "HAS_DEVCONTAINER_DIR=true" >> $GITHUB_OUTPUT
```

code-style/action.yml
```diff
@@ -49,3 +63,36 @@ runs:
...
+        fi
+
+    - name: "Run Dockerfile linting in .devcontainer directory"
+      uses: ansys/actions/docker-style@main
+      if: steps.checkdevcontainer.outputs.exists == 'true'
```
jorgepiloto, line 95:
```suggestion
      if: env.HAS_DEVCONTAINER_DIR == 'true'
```

doc/source/style-actions/examples/docker-style-basic.yml
```diff
@@ -0,0 +1,10 @@
+code-style:
```
jorgepiloto, line 1:
```suggestion
docker-style:
```

doc/source/style-actions/examples/docker-style-basic.yml
```diff
@@ -0,0 +1,10 @@
...
+  name: "Running docker style checks"
```
jorgepiloto, line 2:
```suggestion
  name: "Docker style"
```

docker-style/action.yml
```diff
@@ -0,0 +1,80 @@
...
+
+runs:
+  using: "composite"
+  steps:
+
+    - name: "Check docker directory existence"
+      shell: bash
+      run: |
+        if [ ! -d "${{ github.workspace }}/docker" ]; then
+          echo "Directory 'docker' does not exist. Except in special cases, e.g. '.devcontainer', we emphasize that dockerfiles are expected to be in the docker directory in the root of the project."
```
jorgepiloto, line 53: We have a logging action that allows for a colorized output. It has three levels: info, warning, and error. It would be great to add it here. Take a look to one of the `doc-deploy-*` actions. Those use it heavily.

jorgepiloto (dismissed):
I just left some minor changes to fix some conditions and naming. Other that this it looks great!

We will need to be careful when launching v5.1 since this is affecting the code-style action directly.

code-style/action.yml
```diff
@@ -49,3 +63,36 @@ runs:
     - name: "Run pre-commit"
       shell: bash
       run: pre-commit run --all-files --show-diff-on-failure
+
+    # The followings steps are applying docker-style action to allowed working dirs if the exist.
```
MaxJPRey, line 67:
```suggestion
    # The following steps are applying docker-style action to allowed working dirs if they exist.
```

docker-style/action.yml
RobPasMue: I'd agree to use this action as well if possible - the output looks nice!

code-style/action.yml
```diff
@@ -49,3 +63,36 @@ runs:
...
+        directory: docker
+        recursive: ${{ inputs.docker-recursive }}
+        error-level: ${{ inputs.docker-error-level }}
+
+    - name: "Check if .devcontainer directory exists"
+      shell: bash
+      id: checkdevcontainer
+      run: |
+        if [ -d "${{ github.workspace }}/.devcontainer/Dockerfile" ]; then
+          echo "::set-output name=exists::true"
```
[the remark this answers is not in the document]
SMoraisAnsys: ✔️ done

code-style/action.yml
SMoraisAnsys: ✔️ done

docker-style/action.yml
```diff
@@ -0,0 +1,80 @@
...
+          exit 1
+        fi
```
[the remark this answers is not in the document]
SMoraisAnsys: Although I find your proposal interesting, I'm going to keep the current structuring following our discussion on the scope of the action and because it makes logging easier to use.

RobPasMue:
Status update @SMoraisAnsys? What's missing to get this finally reviewed and merged?

SMoraisAnsys:
I finished the modifications and testing yesterday night. I just double checked the tests in https://github.com/SMoraisAnsys/pyansys-geometry/actions/runs/7562523028 and everything is working as expected !

I just need a final commit to update the code-style action as, for testing purposes, I had to make it use ansys/actions/code-style@feat/lint_dockerfile and not the main version. I'll ping you soon

RobPasMue:
https://github.com/ansys/actions/actions/runs/7567175913/job/20605870132?pr=382#step:2:983

Looks like there is an issue in the docs of the action

docker-style/action.yml
```diff
@@ -0,0 +1,138 @@
...
+  Lint Dockerfile
+
+description: >
+  Evaluate the quality of your project Dockerfile(s) by using `hadolint
+  <https://github.com/hadolint/hadolint/>`_. This action is expected to be used within a
+  matrix job to lint Dockerfile(s) from multiple directories. The action uses `hadolint-gh-action
+  <https://github.com/jbergstroem/hadolint-gh-action>`_ behind the scenes. If you want to evaluate multiple
+  Dockerfiles contained in various directories of the provided directory, use the recursive option.
+
+  This action emphasis the fact of having Dockerfile(s) contained inside the 'docker' directory in the
```
RobPasMue, line 11:
```suggestion
  This action emphasizes the fact of having Dockerfile(s) contained inside the 'docker' directory in the
```

docker-style/action.yml
```diff
@@ -0,0 +1,138 @@
...
+  root of the project. Exception can be added as it is the case for the ''.devcontainer' directory.
```
RobPasMue, line 12:
```suggestion
  root of the project. Exceptions can be added as it is the case for the ''.devcontainer' directory.
```

RobPasMue (approved):
LGTM! Great work @SMoraisAnsys

MaxJPRey (approved):
Nice implementation @SMoraisAnsys

SMoraisAnsys:
@MaxJPRey @RobPasMue @jorgepiloto @germa89 Following @greschd comment (potential problems when using the python package `docker`) I've not been able to reproduce something that could be troubling unless I do have a directory named "docker" (as mentioned before), this directory contains a "__init__.py" file and I import docker from the root of my project.

Therefore, I am divided on whether to rename the dedicated "docker" directory to "docker-files". On one hand, since such situation (even if uncommon) might occur, I'd like to avoid creating an incompatibility between the action and one of our repo. On the other hand, naming such directory "docker" is quite standard as a lot of projects use "docker" as directory name for docker related files. What are you thoughts on this ?

RobPasMue:
As much as I hate this... we will have to provide the folder name as an input and default it to ``docker``, IMO. This will allow users like @greschd to adapt it to whatever folder name they choose in that case. However, it will also open the door for people to do as "they wish" and just use this optional input to adapt to their needs. We shouldn't promote using it unless strictly necessary, like in this scenario. Bear in mind that the goal is that all repos have a similar styling, both in folder structure and file. But we also have to accept these situations and adapt 😄 thanks for bringing it to our attention @greschd

SMoraisAnsys:
FYI I'm refactoring the action to follow our previous discussions.
There is still an issue when triggering `hadolint-gh-action` multiple times in the same workflow, I'll wait for https://github.com/jbergstroem/hadolint-gh-action/pull/135 to be merged before going out of draft mode.

SMoraisAnsys:
@RobPasMue the PR is completed, I triggered it on a test repo and it went as expected: https://github.com/SMoraisAnsys/pyansys-geometry/actions/runs/7628030247/job/21173613810?pr=1

I'll update the PR title to be coherent with the refactoring we had to perform.

docker-style/action.yml
```diff
@@ -0,0 +1,133 @@
...
+      with:
+        level: "INFO"
+        message: >
+          Lint directory '${{ github.workspace }}/${{ inputs.directory }}'.
+
+    - name: "Run Hadolint"
+      if: env.DIRECTORY_EXISTS == 'true'
+      uses: jbergstroem/hadolint-gh-action@v1
+      with:
+        dockerfile: >-
```
RobPasMue, line 117: Interesting... why the dash? ``>-``

RobPasMue (approved):

docker-style/action.yml
SMoraisAnsys: It was to use "Folded Style" to create a "long" string in a more readable way by spreading it over several lines.
It might be possible to do it in another way but I didn't know it!

SMoraisAnsys:
@germa89 @jorgepiloto Since you guys provided me with feedback (**thanks again**) do you agree with the changes ?

docker-style/action.yml
```diff
@@ -0,0 +1,133 @@
...
+          If it does not, then a warning is emitted.
+
+    - name: "Check if docker directory exists"
+      shell: bash
+      run: |
+        if [ -d "${{ github.workspace }}/docker" ]; then
+          echo "HAS_DOCKER_DIR=true" >> $GITHUB_ENV
+        else
+          echo "HAS_DOCKER_DIR=false" >> $GITHUB_ENV
+        fi
```
germa89, line 72: This step is always performed... regardless the value of ``directory``. I think this is a bit misleading (aka anti-pattern)

I would do ``directory`` to be a string array, or None. If ``None``, then we default to docker if exists (we rely on the style action to disable the docker lint if we dont want to do docker linting).

germa89 (approved):
Overall it looks good. (I didnt run it ofc).

Just one remark: I would suggest to use string arrays for the input variable.

germa89:
By the way, thank you for the TLDR section in the PR. It is veeery appreciated!

docker-style/action.yml
SMoraisAnsys: If you don't want to lint the docker directory then you can do it. Contrary to the previous approach, which would fail if the input directory was not `"docker"` or `".devcontainer"`, this new approach only triggers a warning saying 

```
No 'docker' directory found. Except in special cases, e.g. '.devcontainer', we emphasize that docker files are stored in the 'docker' directory at the root of the repository.
```

That's our way of reminding users that they should have a docker directory in which they store their docker files. If the message is not to your taste, what about

```
No 'docker' directory found. If possible, please follow the common approach of storing docker files in the 'docker' directory at the root of the repository.
```

Regarding the `directory` being a string array:
1. I think that the current approach already does what you propose since `docker-lint` can be used in a matrix job and defaults to `"docker"` if no input is provided
2. How would you write the action using a string array ? I can only think of using another script file which would add more disadvantages than advantages :(

germa89: Well.. I don't want to discuss the directory structure, rather the need of having a matrix.

It seems that the `hadolint` action support multiple files: https://github.com/jbergstroem/hadolint-gh-action/blob/main/USAGE.md#pass-multiple-dockerfiles-to-process

If that is the case, then you can do something like:

```yml
  steps:
    - name: "Run PyAnsys docker style checks"
      uses: ansys/actions/docker-style@{{ version }}
      with:
        directory: |
            - docker
            - .devcontainer
            - my_other_docker_directory
```
The directory argument will be parsed as one string, replace the `\n *- `(regex) by one single space, and pass it to the `hadolint` action. This way, you avoid having one job per directory, because I presume that action does not render the worker useless, so it can be reused.

SMoraisAnsys: I'll try it out and see how it does render in the end. I liked using matrix since it ends up with separated workflows but I hear your argument on resources / time consumption.
I'll test using multiple files at once and see how it works 👍

SMoraisAnsys: Changing have been made to follow the requirement of handling multiple repo at once without using a github matrix. To use this feature one has to provide white space separated directories, e.g. "docker .devcontainer".

germa89: thank you a lot @SMoraisAnsys :)

docker-style/action.yml
```diff
@@ -0,0 +1,160 @@
...
+        if [[ "${{ inputs.recursive }}" == 'true' ]]; then
+          fmt_directory_array=($fmt_directory)
+          for i in "${!fmt_directory_array[@]}"; do
+            fmt_directory_array[i]="${fmt_directory_array[i]}/**/Dockerfile"
+          done
+          fmt_directory="${fmt_directory_array[*]}"
+        else
+          fmt_directory="${directories}/Dockerfile"
+        fi
+        echo "FORMATED_HADOLINT_DOCKERFILE_INPUT=$fmt_directory" >> $GITHUB_ENV
```
germa89, line 137: What happens if ``inputs.recursive`` is false??  then you only analyse ``"${directories}/Dockerfile"``... where ``directories`` is defined??

SMoraisAnsys: Oh daman you are right, nice catch here !

SMoraisAnsys: @germa89 now everything is working fine. Btw, I also found another bug while fixing your remark :/

RobPasMue (approved):

code-style/action.yml
```diff
@@ -49,3 +80,31 @@ runs:
     - name: "Run pre-commit"
       shell: bash
       run: pre-commit run --all-files --show-diff-on-failure
+
+    # ------------------------------------------------------------------------
+
+    # Trigger docker-style action if docker-lint is set to true.
+    - uses: ansys/actions/_logging@main
+      if: ${{ inputs.docker-lint == 'true' }}
```
jorgepiloto, line 88:
```suggestion
      if: inputs.docker-lint == 'true'
```

jorgepiloto: Not saying that this may not work. Just suggesting the syntax we have in other actions.

code-style/action.yml
```diff
@@ -49,3 +80,31 @@ runs:
...
+      with:
+        level: "INFO"
+        message: >
+          Extend the workflow to lint Dockerfile(s).
+
+    - name: "Run Dockerfile linting in directories 'docker' and '.devcontainer'"
+      if: ${{ inputs.docker-lint == 'true' }}
```
jorgepiloto, line 95:
```suggestion
      if: inputs.docker-lint == 'true'
```

code-style/action.yml
```diff
@@ -5,7 +5,15 @@ description: >
   This action evaluates the code quality of your project by using `pre-commit
   <https://github.com/pre-commit/pre-commit/>`_. The action installs and runs
   ``pre-commit``. It is assumed that your project contains a
-  ``.pre-commit-config.yaml`` file in the root directory.
+  ``.pre-commit-config.yaml`` file in the root directory. The action can also
+  be extended to lint docker files that are contained in docker and
+  .devcontainer.
```
jorgepiloto, line 10:
```suggestion
  be extended to lint docker files that are contained in the docker directory and the .devcontainer directory.
```

doc/source/style-actions/examples/docker-style-basic.yml
```diff
@@ -0,0 +1,10 @@
+docker-style:
+  name: "Docker style"
+  runs-on: ubuntu-latest
+  steps:
+    - name: "Run PyAnsys docker style checks"
+      uses: ansys/actions/docker-style@{{ version }}
```
jorgepiloto, line 6:
```suggestion
    - uses: ansys/actions/docker-style@{{ version }}
```

doc/source/style-actions/index.rst
```diff
@@ -46,6 +46,27 @@ Doc style action
     {% endfor %}
 
 
+Docker style action
+-------------------
+.. jinja:: docker-style
```
jorgepiloto, line 51:
```suggestion
-------------------

.. jinja:: docker-style
```

docker-style/action.yml
```diff
@@ -0,0 +1,157 @@
...
+      'docker' directory in the root of the project.
+
+  .. warning::
+
+      This action only looks for docker files named Dockerfile.
+      A docker file like Dockerfile.linux will not be linted.
+
+inputs:
+
+  # Optional inputs
```
jorgepiloto, line 28: Should we make these required inputs?

jorgepiloto (changes requested):
Thanks for adding this, @SMoraisAnsys. Just left some minor suggestions before merging this.

doc/source/style-actions/index.rst
SMoraisAnsys: I'll apply this to `code-style` and `doc-style`

docker-style/action.yml
SMoraisAnsys: I'm not opposed to it. However, I liked the fact that we pointed to the docker directory by default. It reinforces the idea that this is 'the default place'.

jorgepiloto: Sounds good.

jorgepiloto (approved):

jorgepiloto:
The doc build is failing but just because the action is new and declared for the first time in this pull-request.

PR #388, opened 2024-02-01
fix: update actions/setup-python@v5

jorgepiloto:
As per internal discussion, this pull-request fixes the version of the actions/setup-python.


jorgepiloto:
I wonder why `dependabot` does not open pull-requests for updating the reusable actions. May it be possible that only pays attention to the `.github/` dir... despite being configure with `/` dir...?

RobPasMue:
Looks like!

PR #391, opened 2024-02-12
fix: docker directory check

SMoraisAnsys:
Previous tests performed on `docker-style` were performed in a context where the repo was already cloned on the runner.
However, when that is not the case, the action does not run since there is nothing to lint and warning messages are raised for nothing (see https://github.com/ansys/pyansys-geometry/actions/runs/7869710413/job/21469384363)

This PR adds an extra step as in `doc-style` to trigger project clone if wanted (default is true).

### DIFF
--- a/docker-style/action.yml
+++ b/docker-style/action.yml
@@ -65,7 +65,7 @@ runs:
 
     - name: "Install Git and clone project"
       uses: actions/checkout@v4
-      if: ${{ inputs.checkout == 'true' }}
+      if: inputs.checkout == 'true'
 
     # ------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


jorgepiloto (approved):

docker-style/action.yml
```diff
@@ -51,10 +51,22 @@ inputs:
...
+    required: false
+    type: boolean
+
 runs:
   using: "composite"
   steps:
 
+    - name: "Install Git and clone project"
+      uses: actions/checkout@v4
+      if: ${{ inputs.checkout == 'true' }}
```
jorgepiloto, line 68:
```suggestion
      if: inputs.checkout == 'true'
```

jorgepiloto (approved):

RobPasMue (approved):

PR #393, opened 2024-02-12
feat: hide output log

MaxJPRey:
The idea is to be able to hide the output of the GitHub action log.

For public repos, it is visible by anyone and could expose vulnerabilities.

This solution of redirecting the output toward `devnull` is far from optimal but better than letting full access to the log for this action.

### DIFF
--- a/check-vulnerabilities/action.yml
+++ b/check-vulnerabilities/action.yml
@@ -88,6 +88,15 @@ inputs:
     default: false
     type: boolean
 
+  hide-log:
+    description: >
+      Whether to hide the output log of the GitHub action.
+      By default, the log is available to anyone for public repositories.
+      This would disclose any potential vulnerabilities to anyone.
+    default: true
+    required: false
+    type: boolean
+
   checkout:
     description: >
       Whether to clone the repository in the CI/CD machine. Default value is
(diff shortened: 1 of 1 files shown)


check-vulnerabilities/action.yml
```diff
@@ -453,7 +462,11 @@ runs:
     - name: "Run safety advisory checks"
       shell: bash
       run: |
-        python dependency-check.py
+        if [[ ${{ inputs.hide-log }} == 'true' ]]; then
+          python dependency-check.py > /dev/null 2>&1
```
MaxJPRey, line 466: @RobPasMue @jorgepiloto Maybe we should just redirect `stdout` and not `stderr`for any potential debugging.
But I don't have any certainty of futur content of the `stderr`.

RobPasMue: I'd redirect everything honestly. Don't worry

RobPasMue (approved):

check-vulnerabilities/action.yml
SMoraisAnsys: Since this PR is about hidding output logs, which make total sense, I think we should redirect to `stdout` for the moment. Like you, I'm not sure about the content of `stderr` either. Moreover, the "core" of this action is related to a python script written on the fly. I have the feeling that we could debug manually if necessary.

PR #396, opened 2024-02-15
feat: allow project installation

### LINKED ISSUE #395: Support mypy checks

Currently it's difficult to get mypy static type checking working with the code-style action. There are two ways of getting mypy checking to include dependencies with pre-commit: either add the dependencies to the pre-commit config, or use the 'system' language and install mypy as a development dependency of your project.

Specifying the dependencies in the pre-commit project isn't great, because they could get out of sync with your pyproject.toml. As a result, I've gone with the second option. However, the code-style action doesn't install the project dependencies, it only checks out the code.

This additional step should be added as an argument, so mypy can be used with the 'system' language.

_No response_

_No response_

RobPasMue:
Closes #395 

mypy requires dependencies and project to be installed. By default we will skip the installation of the library but if repos need it they can set

```skip-install: false```

### DIFF
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -79,12 +79,16 @@ runs:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
 
-    - name: "Install project (if required)"
-      if: inputs.skip-install == 'false'
+    - name: "Create virtual environment"
       shell: bash
       run: |
         python -m pip install --upgrade pip
         python -m venv .venv
+
+    - name: "Install project (if required)"
+      if: inputs.skip-install == 'false'
+      shell: bash
+      run: |
         source .venv/bin/activate
         if grep -q 'build-backend = "poetry\.core\.masonry\.api"' "pyproject.toml"; then
           python -m pip install poetry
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by SMoraisAnsys — APPROVED** [review]
LGTM

**Review by Revathyvenugopal162 — APPROVED** [review]

**Comment by Andy-Grigg** [comment]
I've given this a quick test against the jobqueue repo. It looks like it's failing because the `pip install .` line doesn't install the poetry runtime dependencies. I don't really know if this is a solvable problem with this approach: we don't want to have mypy as a main dependency or even an extra, but without using poetry to do the installation you won't pick up the dev dependencies.

**Comment by RobPasMue** [comment]
No worries - let's do a system check in that case

File: code-style/action.yml [hunk 1]
```diff
@@ -76,6 +85,16 @@ runs:
         python -m pip install --upgrade pip pre-commit
         pre-commit install
 
+    - name: "Install project (if required)"
+      if: inputs.skip-install == 'false'
+      shell: bash
+      run: |
+      if grep -q 'build-backend = "poetry\.core\.masonry\.api"' "pyproject.toml"; then
+        poetry install
```
Andy-Grigg: I expect you'll need to install poetry as well in this case?

Andy-Grigg (approved):
Changes to use an explicit virtual environment for poetry look good. Thanks!

RobPasMue:
@ansys/pyansys-core - a kind review?

germa89 (approved):

PR #402, opened 2024-02-20
feat: release workflow

### LINKED ISSUE #399: Automate the release process
🐞 **Problem**

This project follows a rolling release. That means we need to update the main and minor tags every time a new release is done. This translates into deleting these tags in the remote and re-tagging again. All this is manually and prone to error.

💡 **Idea**

Two ideas come to my mind:

- Use `tox.ini` and code the logic there for different platforms. Logic must be: read VERSION file -> delete necessary tags locally and remotely -> generate new tags -> upload new tags

- Create a `release.yml` workflow

RobPasMue:
Closes #399

Now, whenever you create a tag of type ``vX.Y.Z``, a rolling release is created for tags ``vX`` and ``vX.Y``


.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
+name: Rolling release
+on:
+  push:
+    tags:
+      - "v*.*.*"
```
jorgepiloto, line 5:
```suggestion
  push:
    branches:
      - release/*
    tags:
      - "v*.*.*"
```

jorgepiloto (changes requested):
Just ensuring that this only triggers in `release/` branches.

.github/workflows/release.yml
RobPasMue: Yep - makes sense

jorgepiloto (approved):
Awesome!

MaxJPRey (commented):
Thanks for making our deployment more easy and reliable.

PR #404, opened 2024-02-20
feat: os independent venv activation on code-style action

### LINKED ISSUE #400: code-style action doesn't support Windows runners

### 🐞 Description of the bug

It is an exceptionally odd scenario that leads to requiring Windows, but I'll raise this ticket anyway. Feel free to disregard.

In the [openapi-common](https://github.com/ansys/openapi-common) package we depend on the gssapi python package. To install this package on Linux, the `krb5` package must be installed. It isn't installed on ubuntu-latest, and GitHub (understandably) doesn't give us root access, so we can't install this package.

Installing gssapi on Windows doesn't have this requirement.

For testing purposes, we use the python docker container which seemingly does give us root access, and so we can install the necessary package there. See https://github.com/ansys/openapi-common/blob/664a41240e4b7d713c0251faba79621f2c15f359/.github/workflows/ci_cd.yml#L79 if you're interested.

So to summarize, we can't use code-style for openapi common because:

* We want to do mypy checking, which requires installing the venv
* We can't use the ubuntu-latest runner, because it doesn't have krb5, so we can't install the package's dependencies
* We can't use the python docker image (or an ubuntu* or debian image) because it doesn't support installing Python using setup-python
* We can't use the windows-latest runner because this action doesn't support it

In my experience, the easiest way to get cross-platform support is to use the `pwsh` shell (on all github agents) and use the powershell version to activate the venv: `venv/*/Activate.ps1` does the trick.

\* This generates some really weird looking error messages. I can't tell for certain that it's even supported, so I'm just not going to bother exploring.

### 📝 Steps to reproduce

N/A - see above

Windows

N/A - CI only

3.10

### 📦 Installed packages

```shell
N/A
```

RobPasMue:
Closes #400 

Could you give it a try to this branch @Andy-Grigg?

### DIFF
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -85,11 +85,22 @@ runs:
         python -m pip install --upgrade pip
         python -m venv .venv
 
+    - name: "Store venv activation mechanism depending on OS"
+      shell: bash
+      run: |
+        if [[ "$RUNNER_OS" == "Linux" ]]; then
+          echo "ACTIVATE_VENV=$(echo 'source .venv/bin/activate')" >> $GITHUB_ENV
+        elif [[ "$RUNNER_OS" == "macOS" ]]; then
+          echo "ACTIVATE_VENV=$(echo 'source .venv/bin/activate')" >> $GITHUB_ENV
+        elif [[ "$RUNNER_OS" == "Windows" ]]; then
+          echo "ACTIVATE_VENV=$(echo '.venv\Scripts\activate')" >> $GITHUB_ENV
+        fi
+
     - name: "Install project (if required)"
       if: inputs.skip-install == 'false'
       shell: bash
       run: |
-        source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


RobPasMue:
Hey @Andy-Grigg - I gave it a try here: https://github.com/ansys/openapi-common/pull/483/commits/c7ae94d0e09a583490c1be7526f07c7e44613137

Looks like it's working properly now so I will merge this as soon as it gets approved and I will do a patch release right after. This should solve your issues.

SMoraisAnsys:
I'm not sure to understand why these changes are needed after what was performed in #400.

Since the steps use `shell: bash`, shouldn't that do the trick for all OS ?
From what I see with @Andy-Grigg CI this information is not leveraged in the custom code-style workflow. Might give a try to the main version of `ansys/actions` ?

SMoraisAnsys (approved):
LGTM

Issue #284, opened 2023-07-19
Multi-version redirection link is killing pages indexing


jorgepiloto:
🐞 **Problem**
The main `index.html` created by the multi-version documentation looks as follows:

https://github.com/ansys/actions/blob/2070e78a8dc27a01d4f59bafacd4644d887d6fe9/doc-deploy-stable/action.yml#L356-L363

In previous code, the `<title>` tag encloses the following text: `Redirecting to...`. Because this text does not contain a proper title, Google is not properly indexing pages. Also, this index lacks a proper description and keywords. 

Solving this issue is urgent for any project using the multi-version documentation.

RobPasMue:
Agreed - this should be handled ASAP :) let me know if you need any help

PR #406, opened 2024-02-21
fix: complete HTML file for redirect

jorgepiloto:
Some progress on #284. This pull-request ensures that the content of the redirection `index.html` file is a valid HTML file.

### DIFF
--- a/doc-deploy-dev/action.yml
+++ b/doc-deploy-dev/action.yml
@@ -229,15 +229,19 @@ runs:
       shell: bash
       run: |
         echo "<!DOCTYPE html>" > index.html
-        echo "<meta charset=\"utf-8\">" >> index.html
+        echo "<html>" >> index.html
+        echo "  <head>" >> index.html
+        echo "    <meta charset=\"utf-8\">" >> index.html
         if [ -z "${{ inputs.redirection-title }}" ]; then
-          echo "<title>Redirecting to ${{ env.REDIRECTION_URL }}</title>" >> index.html
+          echo "    <title>Redirecting to ${{ env.REDIRECTION_URL }}</title>" >> index.html
         else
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):

Revathyvenugopal162 (approved):

SMoraisAnsys (approved):

MaxJPRey (approved):
LGTM.

doc-deploy-dev/action.yml
```diff
@@ -229,15 +229,19 @@ runs:
...
+          echo "    <title>${{ inputs.redirection-title }}</title>" >> index.html
         fi
-        echo "<meta name=\"description\" content=\"${{ inputs.redirection-description }}\">" >> index.html
-        echo "<meta http-equiv=\"refresh\" content=\"0; URL=${{ env.REDIRECTION_URL }}\">" >> index.html
-        echo "<link rel=\"canonical\" href=\"${{ env.REDIRECTION_URL }}\">" >> index.html
+        echo "    <meta name=\"description\" content=\"${{ inputs.redirection-description }}\">" >> index.html
+        echo "    <meta http-equiv=\"refresh\" content=\"0; URL=${{ env.REDIRECTION_URL }}\">" >> index.html
+        echo "    <link rel=\"canonical\" href=\"${{ env.REDIRECTION_URL }}\">" >> index.html
+        echo "  </head>" >> index.html
+        echo "</html>" >> index.html
```
germa89, line 244: I would have probably write a template HTML file, and then replace the template tags with ``sed``. Just saying, this is a bit difficult to follow.

jorgepiloto: That's a good point. I think we had to go this way because `sed` was not allowing us to use variable substitution inside the template.

PR #407, opened 2024-02-21
fix: replace broken MIT link

klmcadams:
The link to the MIT license, https://opensource.org/licenses/MIT, is broken so I replaced it with https://opensource.org/blog/license/mit

Example of where it broke in a workflow: https://github.com/ansys/actions/actions/runs/7995412316/job/21835720195?pr=357#step:2:900


RobPasMue (approved):
LGTM! Can you present this at the meeting later today?

PR #429, opened 2024-03-19
feat: extend doc-build to be windows compatible

SMoraisAnsys:
WIP to extend doc-build action to work with windows and linux.


SMoraisAnsys:
@RobPasMue Here is a first draft of the modification to allow windows doc build. I need to test it on some repos to ensure that it works correctly but I was able to reach the build command on my fork.

RobPasMue:
Let me try using PyAnsys Geometry with this one =)

SMoraisAnsys:
I'll also try to build pyedb with it using linux and windows.

RobPasMue:
Trying in https://github.com/ansys/pyansys-geometry/pull/1068

_doc-build-linux/action.yml
```diff
@@ -0,0 +1,284 @@
...
+    description: >
+      Whether to use the Python cache for installing previously downloaded
+      libraries. If ``true``, previously downloaded libraries are installed from the
+      Python cache. If ``false``, libraries are downloaded from the PyPI index.
+    required: false
+    default: true
+    type: boolean
+
+  sphinxopts:
+    description: >
```
jorgepiloto, line 60:
```suggestion
    description: |
```

doc-build/action.yml
jorgepiloto, line 1: Could we install Python outside the private actions? This is a common step in both and a cross-platform command.

jorgepiloto (changes requested):

PR #430, opened 2024-03-20
feat: upload vulnerabilities to dashboard

RobPasMue:
As title says. Related to vulnerability tasks.


RobPasMue:
Currently being used in https://github.com/ansys/pyansys-geometry/pull/1071

_pyansys-dashboard-vulnerabilities-upload/action.yml
```diff
@@ -0,0 +1,192 @@
...
+    required: true
+    type: string
+
+  hide-log:
+    description: >
+      Whether to hide the output log of the GitHub action.
+      By default, the log is available to anyone for public repositories.
+      This would disclose any potential vulnerabilities to anyone.
+    required: true
+    type: boolean
```
MaxJPRey, line 83: Can we add a default value and set it to `true`?

RobPasMue: This is a private action. It is only called from the ansys/check-vulnerabilities action. We are passing the value used in that action here.

Following the same approach we have with other private actions, all inputs are compulsory and none have a default value.

RobPasMue: I need to change the description though - that's my bad for copypasting

_pyansys-dashboard-vulnerabilities-upload/action.yml
```diff
@@ -0,0 +1,191 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: "Check whether the repository owner is 'ansys' or 'ansys-internal'"
+      shell: bash
+      run: |
+        if [[ ${{ github.repository_owner }} != "ansys" && ${{ github.repository_owner }} != "ansys-internal" ]]; then
```
jorgepiloto, line 90: This condition should be an "or":

```suggestion
        if [[ ${{ github.repository_owner }} != "ansys" || ${{ github.repository_owner }} != "ansys-internal" ]]; then
```

jorgepiloto (changes requested):

_pyansys-dashboard-vulnerabilities-upload/action.yml
RobPasMue: Imagine you have a repo owned by "numpy"

With the original if...

``if [[ numpy != "ansys" && numpy != "ansys-internal" then ...``

Which equals to ``True && True`` --> And consequently the if condition is triggered

With your change, it would be ``True OR True`` which is also valid...

But what happens if it belongs to the "ansys" org?

``False AND True`` --> False --> It does not trigger the IF condition (expected behavior)
``False OR True`` --> True --> It triggers the IF condition (unexpected behavior)

So... @jorgepiloto, I believe your suggestion is wrong

RobPasMue: Please, verify my thought logic in the previous comment and approve otherwise

jorgepiloto: You are right! I got confused with the negation, my fault. Approving this 👍🏼

jorgepiloto (approved):

check-vulnerabilities/action.yml
```diff
@@ -501,6 +541,7 @@ runs:
 
 
 
+
```
RobPasMue, line 544:
```suggestion

```

PR #445, opened 2024-04-05
refactor: way to retrieve pull request title

### LINKED ISSUE #443: Rework accessing PR title in changelog action
Currently, the doc-changelog action retrieves the title of the PR using `${{ github.event.pull_request.title }}`.
When the title is edited, one may not retrieve the right title unless the workflow is configured to work on PR title edition, for example:
```
pull_request:
    types: [opened, reopened, synchronize, edited]
```
However, specifying such options will trigger the other jobs of the workflow when the PR title is updated.
This can be quite burdensome as merging a PR can preempted for 30+ minutes (or even more) because you have to rerun everything after figuring out that the title is not correct.

SMoraisAnsys:
Retrieving a pull request title was performed through `${{ github.event.pull_request.title }}`. However, this works correctly with edited titles only if you specify that your workflow is triggered on PR edition, e.g.
```
pull_request:
    types: [opened, reopened, synchronize, edited]
```

Since it causes other jobs to be also triggered, I propose another approach to retrieve the PR title. This will allow one to retrigger manually the changelog action after a title change, without rerunning the whole workflow.

Overall I would say that it improves a bit the flexibility of the action. However, if one wants to keep the fact that editing the PR title triggers automatically the job, then one will have to keep using the mentioned scope activation (`edited`). 

Closes #443

### DIFF
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -153,10 +153,27 @@ runs:
           rm $file
         fi
 
+    - name: "Retrieve PR title"
+      env:
+        REPOSITORY_OWNER: ${{ github.repository_owner }}
+        REPOSITORY_NAME: ${{ github.event.repository.name }}
+        PULL_REQUEST_NUMBER: ${{ github.event.number }}
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const response = await github.graphql(`query {
+            repository(owner: "${{ env.REPOSITORY_OWNER }}", name: "${{ env.REPOSITORY_NAME }}") {
+              pullRequest(number: ${{ env.PULL_REQUEST_NUMBER }}) {
+                title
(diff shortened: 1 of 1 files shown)


RobPasMue (changes requested):
There's something I don't get @SMoraisAnsys ...

Why would you want to perform a commit that does not trigger the whole workflow?

The fact that you modify the PR title should trigger the changelog action since the content inside the fragment will be modified. This means that you will have a new commit in your branch and according to the normal way of protecting our repos through CI/CD, the last commit should pass the required stages...

Bottom line - **you will still need to run the full workflow despite the PR title change**

SMoraisAnsys:
Closing since a modifications on the title creates a new commit (which should trigger the whole workflow).

PR #447, opened 2024-04-08
feat: allow upper case in commit-style

### LINKED ISSUE #446: feat: extend commit-style to work with upper case

As title says.

An idea of how to do this could be to leverage the `headerPattern` or `types` options in https://github.com/amannn/action-semantic-pull-request/tree/main

_No response_

_No response_

SMoraisAnsys:
As title says.

Closes #446

Note: option `headerPattern` seems to be related to "how to parse" a commit title to retrieve the `type`, `scope` and `description`. Thus it seems more adequate to use option `types` instead.

### DIFF
--- a/commit-style/action.yml
+++ b/commit-style/action.yml
@@ -38,12 +38,47 @@ inputs:
     required: true
     type: string
 
+  # Optional inputs
+
+  use-upper-case:
+    description: >
+      Use of uppercase letters in the "type" and "scope" fields of the commit.
+      For example, "FIX(SERVER)!: fix server crash issue" would be a valid commit.
+
+      .. note::
+
+          Expected type are upper cases of `conventional commit types
+          <https://github.com/commitizen/conventional-commit-types/blob/master/index.json>`_.
+    required: false
+    default: false
+    type: boolean
+
 runs:
   using: "composite"
   steps:
 
     - name: "Check pull-request title follows conventional commits style"
-      if: ${{ (github.event_name == 'pull_request_target') || (github.event_name == 'pull_request') }}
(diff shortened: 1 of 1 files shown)


commit-style/action.yml
```diff
@@ -38,12 +38,47 @@ inputs:
...
+          CHORE
+          CI
+          DOCS
+          FEAT
+          FIX
+          PERF
+          REFACTOR
+          REVERT
+          STYLE
+          TEST
```
RobPasMue, line 83: Does this accept regex? Otherwise we could just have one case and add ``(?i)``. Just wondering 😄

SMoraisAnsys: To my knowledge it is not possible (see https://github.com/amannn/action-semantic-pull-request/issues/220)

commit-style/action.yml
```diff
@@ -38,12 +38,47 @@ inputs:
...
+  use-upper-case:
+    description: >
+      Use of uppercase letters in the "type" and "scope" fields of the commit.
+      For example, "FIX(SERVER)!: fix server crash issue" would be a valid commit.
+
+      .. note::
+
+          Expected type are upper cases of `conventional commit types
+          <https://github.com/commitizen/conventional-commit-types/blob/master/index.json>`_.
+    required: false
```
RobPasMue, line 52:
```suggestion
          <https://github.com/commitizen/conventional-commit-types/blob/master/index.json>`_.

    required: false
```

RobPasMue (approved):
LGTM - just one minor comment to enhance readability

MaxJPRey (approved):
This is great @SMoraisAnsys .

PR #450, opened 2024-04-08
feat: logging in windows environment

SMoraisAnsys:
This PR ensures that a self-hosted runner recent enough, i.e. using windows 10, created without specific setup is able to use the `_logging` private action. Before it was only working if bash was included in PATH which is not the case for many self hosted runners.

This will allow future actions dedicated to windows could use the _logging action without breaking changes.


SMoraisAnsys:
This worked with linux and windows github runners and a windows self-hosted runner used by pyaedt toolkits (see https://github.com/ansys/actions/actions/runs/8600801687/)

jorgepiloto (approved):

PR #410, opened 2024-02-23
chore: Update default vale version `3.4.1`

dipinknair:
Update the default vale version to `3.4.1`

PR #457, opened 2024-04-17
feat: update to Vale 3

jorgepiloto:
Continuation of #410. Adapts the Vale configuration to use latest version.


doc/styles/Vocab/ANSYS/accept.txt
RobPasMue, line 1: These files need to be added to the vocabularies folder @jorgepiloto

RobPasMue (changes requested):

doc/styles/.gitignore
RobPasMue, line 1: This file should exist @jorgepiloto - other vocabularies are downloaded on runtime locally. We don't want them to appear as changes

jorgepiloto: Thanks! I did not noticed git ignored this files.

doc/styles/.gitignore
```diff
@@ -1,4 +1,4 @@
-*
-!Vocab
-!Vocab/**
+/*
+!**/accept.txt
+!**/reject.txt
```
RobPasMue, line 3:
```suggestion
!**/ANSYS/accept.txt
!**/ANSYS/reject.txt
```

RobPasMue (approved):

PR #458, opened 2024-04-17
fix: regex for fixing links in landing page

jorgepiloto:
Fixes the issues originated when a local link starts with `h`, or `t`, or `p`.


RobPasMue (approved):
LGTM!

PR #464, opened 2024-04-18
docs: Add `code spell` hook changes in vale migration guide

dipinknair:
New vocabularies path has to be modified for code spell hook if it uses `accept.txt`


RobPasMue (approved):
LGTM! Nice catch @dipinknair - doing new patch release

PR #470, opened 2024-04-23
fix: fail doc-style action when wrong Vale config is present

jorgepiloto:
As per a private discussion, we are forcing the `doc-style` action to check the Vale configuration layout according to the selected version.

### DIFF
--- a/doc-style/action.yml
+++ b/doc-style/action.yml
@@ -89,14 +89,14 @@ runs:
       shell: bash
       run: |
         if [ $VALE_MAJOR_VERSION -ge 3 ]; then
-          if [ ! -d ./doc/styles/config/vocabularies/ANSYS ]; then
-            echo "Projects using Vale >=v3 require a doc/styles/config/vocabularies/ANSYS directory."
+          if [ ! -d ./doc/styles/config/vocabularies ]; then
+            echo "Projects using Vale >=v3 require a doc/styles/config/vocabularies directory."
             echo "See migration guide https://actions.docs.ansys.com/version/stable/migrations/docs-style-vale-version-update.html"
             exit 1
           fi
         else
-          if [ ! -d ./doc/styles/Vocab/ANSYS ]
-            echo "Projects using Vale <=v2 require a doc/styles/Vocab/ANSYS directory. Consider updating your Vale version to v3."
+          if [ ! -d ./doc/styles/Vocab ]
+            echo "Projects using Vale <=v2 require a doc/styles/Vocab directory. Consider updating your Vale version to v3."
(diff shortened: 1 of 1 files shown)


doc-style/action.yml
```diff
@@ -71,6 +71,37 @@ runs:
...
+      run: |
+        version=${{ inputs.vale-version }}
+        vale_major_version=$( echo $version | cut -d '.' -f 1)
+        echo "VALE_MAJOR_VERSION=$(echo $major)" >> $GITHUB_ENV
+
+    - name: "Verify Vale's configuration layout according to selected version"
+      shell: bash
+      run: |
+        if [ $VALE_MAJOR_VERSION -ge 3 ]; then
+          if [ ! -d ./doc/styles/config/vocabularies/ANSYS ]; then
```
RobPasMue, line 92: I would just check up to ``vocabularies`` - maybe some other users out there want to benefit from the action?

doc-style/action.yml
```diff
@@ -71,6 +71,37 @@ runs:
...
+            echo "Projects using Vale >=v3 require a doc/styles/config/vocabularies/ANSYS directory."
```
RobPasMue, line 93:
```suggestion
            echo "Projects using Vale >=v3 require a doc/styles/config/vocabularies directory."
```

doc-style/action.yml
RobPasMue, line 92:
```suggestion
          if [ ! -d ./doc/styles/config/vocabularies ]; then
```

doc-style/action.yml
```diff
@@ -71,6 +71,37 @@ runs:
...
+            echo "See migration guide https://actions.docs.ansys.com/version/stable/migrations/docs-style-vale-version-update.html"
+            exit 1
+          fi
+        else
+          if [ ! -d ./doc/styles/Vocab/ANSYS ]
```
RobPasMue, line 98:
```suggestion
          if [ ! -d ./doc/styles/Vocab ]
```

doc-style/action.yml
```diff
@@ -71,6 +71,37 @@ runs:
...
+            echo "Projects using Vale <=v2 require a doc/styles/Vocab/ANSYS directory. Consider updating your Vale version to v3."
```
RobPasMue, line 99:
```suggestion
            echo "Projects using Vale <=v2 require a doc/styles/Vocab directory. Consider updating your Vale version to v3."
```

RobPasMue (changes requested):
Leaving suggestions to adapt properly - let's try and get this merged @jorgepiloto

RobPasMue (approved):

PR #476, opened 2024-05-01
fix: remove unnecessary curly brackets in doc-deploy-changelog docs

klmcadams:
As the title says

### DIFF
--- a/doc/source/migrations/docs-deploy-changelog-setup.rst
+++ b/doc/source/migrations/docs-deploy-changelog-setup.rst
@@ -25,7 +25,7 @@ Once the ``doc-changelog`` action is done being set up, continue with the ``doc-
       steps:
         - uses: ansys/actions/doc-deploy-changelog@{{ version }}
           with:
-            token: ${{ '{{ secrets.PYANSYS_CI_BOT_TOKEN }}' }}
+            token: ${{ secrets.PYANSYS_CI_BOT_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


SMoraisAnsys:
@klmcadams The pattern ${{ '{{ SOMETING }}' }} seems to be used all over the other actions. I'm not aware of any conflict when moving from ${{ '{{ SOMETHING }}' }} to ${{ SOMETHING }} so it might make sense to apply your changes to other actions right ?
@RobPasMue @jorgepiloto Are you aware of any limitations ?

klmcadams:
@SMoraisAnsys I think we have to keep ${{ '{{ SOMETHING }}' }} in the doc/source/*/examples/ yaml files because if they don't have that format there will be an error when making the docs:

In `doc/source/doc-actions/examples/doc-changelog-basic.yml`, if you remove '{{ }}', you get this error: 
jinja2.exceptions.UndefinedError: 'secrets' is undefined

The rst files in the migrations folder still works without the extra '{{ }}' though

RobPasMue (approved):

RobPasMue:
Just what @klmcadams pointed out - it is important that in the "examples" YAML files we follow the double curly braces approach. Jinja resolves the first set of curly braces and we need them as part of the syntax. But in this case, the migration guide is fine to just use them once.

SMoraisAnsys (approved):

PR #478, opened 2024-05-02
feat: vale ignore towncrier directory

### LINKED ISSUE #477: Exclude the changelog fragments from the vale check by default
Changelog fragments are stored as .md files which means vale tries to check them. They frequently contain library names and other things that vale decides aren't correct.

Rather than having every library separately try to work out what regex to use to exclude the things they want as well as all the fragment files, can we add that exclusion here please? I know vale is a complete pain to use, but lets make it slightly less horrid for library developers!

klmcadams:
Gives the option to ignore md files in the towncrier fragment file directory (typically doc/changelog.d). @da1910 ran into this issue where vale flagged `codecov` in the fragment file as an error. By default, the action will check the towncrier directory unless `ignore-changelogd` is specified.

If it makes more sense to ignore the changelog.d directory by default, then we can make the default value true. Let me know what you all think.

changelog.d directory not ignored: https://github.com/ansys/openapi-common/actions/runs/8930323884/job/24530231218?pr=575

changelog.d directory ignored & other rst files in doc are not ignored: https://github.com/ansys/openapi-common/actions/runs/8930356819/job/24530286024?pr=575

Closes #477


SMoraisAnsys (commented):
I would go for ignoring by default. I assume that many people will end up facing this issue at some point and have to ignore the changelogd directory anyway.

da1910 (approved):
Fix looks OK to me, I have no strong feelings about default/not, but the implementation looks reasonable (given my comments before about storing python in workflows)

MaxJPRey:
> I would go for ignoring by default. I assume that many people will end up facing this issue at some point and have to ignore the changelogd directory anyway.

I would also suggest to ignore it by default.

SMoraisAnsys (approved):

RobPasMue (approved):

PR #483, opened 2024-05-07
feat: use GITHUB_ACTION_PATH for vulnerabilities

RobPasMue:
As title says. Exploring the usage of GITHUB_ACTION_PATH...


RobPasMue:
Testing on https://github.com/ansys/pyansys-geometry/pull/1191

doc/source/vulnerability-actions/index.rst
```diff
@@ -10,6 +10,27 @@ and third-party libraries (that is, dependencies).
...
 
+.. note::
+
+   Users can try out the ``ansys/check-vulnerabilities`` action on their local repository
+   by doing the following:
+
+   #. Download the ``check_vulnerabilities.py`` script and the ``requirements.txt`` file from
+      the `ansys/check-vulnerabilities action folder <https://github.com/ansys/actions/tree/main/check-vulnerabilities>`_.
+   #. Move the downloaded files to the root of the repository.
+   #. Create a virtual environment by running ``python -m venv venv``.
```
RobPasMue, line 21:
```suggestion
   #. Create a virtual environment by running ``python -m venv .venv``.
```

RobPasMue:
Can I get a review @ansys/pyansys-core?

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -0,0 +1,348 @@
...
+ERROR_IF_NEW_ADVISORY = (
+    True if os.environ.get("DEPENDENCY_CHECK_ERROR_EXIT", None) else False
+)
+CREATE_ISSUES = (
+    True if os.environ.get("DEPENDENCY_CHECK_CREATE_ISSUES", None) else False
+)
+
+
+def dict_hash(dictionary: Dict[str, Any]) -> str:
+    """MD5 hash of a dictionary."""
```
AlejandroFernandezLuces, line 32: Missing parameter documentation

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -0,0 +1,348 @@
...
+def check_vulnerabilities():
+    """Check library and third-party vulnerabilities."""
+    new_advisory_detected = False
+    # Check that the needed environment variables are provided
+    if any([v is None for v in [TOKEN, REPOSITORY, PACKAGE]]):
+        raise RuntimeError(
+            "Required environment variables are not defined. Enter value for ",
+            "'DEPENDENCY_CHECK_TOKEN', 'DEPENDENCY_CHECK_PACKAGE_NAME', ",
+            "'DEPENDENCY_CHECK_REPOSITORY'.",
+        )
```
AlejandroFernandezLuces, line 50: Why not unravel this into three ifs and raise an error specific for each key? It seems more useful to know which specific key is missing from my point of view.

SMoraisAnsys (approved):
LGTM, cannot wait to try action_path myself !

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -0,0 +1,348 @@
...
+Please update the security advisory status after evaluating. Publish the advisory
+once it has been verified (since it has been created in draft mode).
+
+---
+
+#### Description
+
+{desc}
+"""
+                repo.create_issue(title=summary, body=issue_body, labels=["security"])
```
AlejandroFernandezLuces, line 148: I wonder what happens here if you are using a token with no write permissions, could it the script fail here? I would prefer to check this earlier in the script rather than letting if fail mid script, but not sure how to do it anyway..

AlejandroFernandezLuces (approved):
LGTM, minor comments:

check-vulnerabilities/check_vulnerabilities.py
RobPasMue: Sure, will do

check-vulnerabilities/check_vulnerabilities.py
RobPasMue: Saving time mostly - I can break it down nonetheless

check-vulnerabilities/check_vulnerabilities.py
RobPasMue: This might be trickier. I need to check whether it is even possible

RobPasMue: Yeah this rings a bell... Then let's leave it as is

PR #484, opened 2024-05-07
fix: allow backtick, single, and double quotes for PR titles in the doc-changelog action

### LINKED ISSUE #479: doc-changelog action fails if PR title contains an apostrophe

### 🐞 Description of the bug

Creating a PR with an apostrophe in the title, e.g. `Don't run changelog=fragment for dependabot-initiated PRs` results in the following error: https://github.com/ansys/grantami-bomanalytics/actions/runs/8971689464/job/24637980481#step:2:284

I think the issue is probably here: https://github.com/ansys/actions/blob/92eb6c66edbd44a1e5d9eabed75ebbf4f0f5457b/doc-changelog/action.yml#L175 - the title is included without any escaping.

### 📝 Steps to reproduce

Create a PR with an apostrophe in the title that triggers this action

Linux

Actions v6

3.9

### 📦 Installed packages

```shell
ansys/actions/doc-changelog@v6
```

klmcadams:
Use python to remove extra whitespace from a pull request title and allow backticks, single quotes, and double quotes to be included in the towncrier fragment file. 

[Fragment file in PR with backtick, single and double quotes](https://github.com/ansys-internal/actions-sandbox/pull/52/files#diff-77db1152256d4011c092a8cb1110424352e40892499225279381c2b18edc7dfd)

[Title before extra whitespaces are removed](https://github.com/ansys-internal/actions-sandbox/actions/runs/8990096183/job/24694702524?pr=52#step:2:243)

Closes #479


doc-changelog/action.yml
```diff
@@ -160,23 +160,39 @@ runs:
...
+        PR_TITLE: ${{ github.event.pull_request.title }}
+      shell: python
+      run: |
+        import os
+
+        # Remove extra whitespace
+        clean_title = " ".join("""${{ env.PR_TITLE }}""".split())
+
+        # Add backslash in front of backtick and double quote
+        clean_title = clean_title.replace("`", "\`").replace('"', '\\"')
```
RobPasMue, line 174:
```suggestion
        clean_title = clean_title.replace("`", "\`").replace('"', '\"')
```

Shouldn't it just be a single backslash?

doc-changelog/action.yml
```diff
@@ -160,23 +160,39 @@ runs:
...
 
+    - name: "Clean PR title"
+      env:
+        PR_TITLE: ${{ github.event.pull_request.title }}
+      shell: python
+      run: |
+        import os
+
+        # Remove extra whitespace
+        clean_title = " ".join("""${{ env.PR_TITLE }}""".split())
```
RobPasMue, line 171:
```suggestion
        # Retrieve title
        clean_title = os.getenv('PR_TITLE')

        # Remove extra whitespace
        clean_title = clean_title.strip()
```

RobPasMue (commented):
Some minor comments - other than that, LGTM

doc-changelog/action.yml
klmcadams: @RobPasMue does strip also remove extra whitespace between words, or only remove trailing whitespace?

RobPasMue: Strip works for leading and trailing whitespaces which is what we should focus on... if users add extra whitespaces inside the PR title, this might be intended or it might not... but I don't think we should be "playing" with the "internals" of the PR title.

doc-changelog/action.yml
klmcadams: @RobPasMue I think I had to do two backslashes so that the first backslash would escape the second like this:

```
>>> string = input("string: ")
string: hello "world"
>>> print(string.replace('"', '\"'))
hello "world"
>>> print(string.replace('"', '\\"'))
hello \"world\"
```

doc-changelog/action.yml
RobPasMue: Imagine the following:

```
docs: adding `   ` instead of ` ` inside the API ref landing page
```

This would get translated into:
```
docs: adding ` ` instead of ` ` inside the API ref landing page
```

Which... is definitely wrong. We want it to show up as the first one. My point is, if the user made a mistake with the number of whitespaces inside the PR title itself, then so be it. But we shouldn't be assuming it.

doc-changelog/action.yml
RobPasMue: Oh gotcha. You want the backslash to show up in the string. Then yeah

RobPasMue (approved):
LGTM!

PR #494, opened 2024-05-28
fix: update doc deploy actions to to correct local search path on landing page

Revathyvenugopal162:
This PR addresses a search issue in the deployment script by updating the search action path in the landing page. Previously, the search action was set to `search.html`, causing search functionality to look for the incorrect file. With this fix, the search action path is updated to `version/stable/search.html` for stable versions or `version/dev/search.html` for development versions. This ensures that the search functionality points to the correct location, resolving the search issue effectively.


RobPasMue (commented):
Awesome! I did realize about this behavior but never ended up reporting it. Thanks a lot @Revathyvenugopal162!

Revathyvenugopal162:
Tested successfully in https://templates.ansys.com/

RobPasMue (approved):

MaxJPRey (approved):
LGTM.

PR #497, opened 2024-06-04
feat: allow upper case in doc-deploy-changelog

### LINKED ISSUE #496: Doc-deploy-changelog action does not work well with commit-style action when use-upper-case is set to True
By default `doc-deploy-changelog` creates a PR with title `chore: update CHANGELOG for vX.Y.Z`.
However, if one uses the `commit-style` with `use-upper-case = true` then the PR title does not follow the expected PR type.
A simple fix would be to add `use-upper-case` to the `doc-deploy-changelog` action.

SMoraisAnsys:
Closes #496

### DIFF
--- a/doc-deploy-changelog/action.yml
+++ b/doc-deploy-changelog/action.yml
@@ -85,9 +85,16 @@ inputs:
 
   use-upper-case:
     description: >
-      Use of uppercase letters in the "type" field of the PR.
-      If ``false``, the PR title is "chore: update CHANGELOG for v...".
-      If ``true``, the PR title is "CHORE: update CHANGELOG for v...".
+      Use of uppercase letters in the "type" field of:
+
+      - the PR created into main to add CHANGELOG changes;
+
+      - the commit created in the release branch to add the new
+        section in CHANGELOG.
+
+
+      If ``false``, the title is "chore: ... CHANGELOG for v...".
+      If ``true``, the PR title is "CHORE: ... CHANGELOG for v...".
     required: false
     default: false
     type: boolean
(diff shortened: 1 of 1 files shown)


RobPasMue (approved):

SMoraisAnsys:
Tested it while enabling the use of upper-case and it seems to work well

image

Should I also add a type to the commit added when creating the fragment ? We would have by default, "chore: adding changelog entry..." instead of "Adding changelog entry..."

doc-deploy-changelog/action.yml
```diff
@@ -83,6 +83,22 @@ inputs:
     default: true
     type: boolean
 
+  use-upper-case:
+    description: >
+      Use of uppercase letters in the "type" field of:
+
+      - the PR created into main to add CHANGELOG changes;
```
Revathyvenugopal162, line 90:
```suggestion
      - the PR created into main to add CHANGELOG changes.
```

Revathyvenugopal162 (approved):

klmcadams (approved):

MaxJPRey (approved):
Thanks @SMoraisAnsys . I will be able to proceed with my upper case habits regarding PRs' titles.

doc-deploy-changelog/action.yml
```diff
@@ -83,6 +83,22 @@ inputs:
...
+
+      - the commit created in the release branch to add the new
```
SMoraisAnsys, line 92:
```suggestion
      - The commit created in the release branch to add the new
```

RobPasMue:
Can we merge this?

SMoraisAnsys:
> Can we merge this?

Yes !

PR #503, opened 2024-06-11
fix: only zip and upload if asked

SMoraisAnsys:
Removes zip & upload even when one doesn't ask for it... (add missing if statement in the doc build workflow)


RobPasMue (approved):
Gotcha - LGTM! Thanks for the quick fix

MaxJPRey (approved):
Thanks for the quick fix.

SMoraisAnsys:
Merging since it worked when using PR branch.

PR #505, opened 2024-06-12
fix: ignoring `jinja2` vulnerability ID

clatapie:
`jinja2` is a required dependency for the `check-vulnerability.py` scripts.
Due to `jinja2` [safety result](https://data.safetycli.com/p/pypi/jinja2/97c/), the `check-vulnerabilities` action is failing.


jorgepiloto (approved):
It would be great if we could add links or comments at some point for each number.

SMoraisAnsys:
@jorgepiloto Yeah, I do agree (that would also help to know when new releases patch the issues)
For the one added : https://data.safetycli.com/v/70612/97c/ and it affects versions >= 0
image

clatapie:
@jorgepiloto, as it's a `.txt` file, it should not be a problem indeed

RobPasMue:
Links are available in the rendered docs: https://actions.docs.ansys.com/version/dev/vulnerability-actions/index.html
Search for "Accepted safety vulnerabilities"

Comments would be fine too for explaining why we are ignoring it

PR #506, opened 2024-06-12
docs: adding a warning for `vulnerability_checks` with private repositories

clatapie:


check-vulnerabilities/ignored-safety.txt
```diff
@@ -1,4 +1,4 @@
-52495
-62044
-67599
-70612
\ No newline at end of file
+52495 # `setuptools` vulnerability - for setuptools versions < 65.5.1
+62044 # `pip` vulnerability - installing a package from a Mercurial VCS URL
+67599 # `pip` vulnerability - using --extra-index-url option
+70612 # `jinja2` vulnerability
```
SMoraisAnsys, line 4:
```suggestion
70612 # `jinja2` vulnerability - affected versions >=0 (current latest version is 3.1.4)
```

SMoraisAnsys (dismissed):
LGTM
I left a minor comment on jinja vulnerability to add visibility on the range of affected versions.

check-vulnerabilities/ignored-safety.txt
RobPasMue, line 1: This is failing, and it will affect all our logic. It is not as simple as adding comments.

Docs build is failing because of this
https://github.com/ansys/actions/blob/6ec01e70893be9ea6ec6b8d540e1964877a21e22/check-vulnerabilities/action.yml#L52

Action will also fail now.

I suggest you revert these comments or figure out the way to do them properly in both places.

RobPasMue (changes requested):
As it is, it is not going to work. Also, the purpose of this PR is not adding the comments. Please do it in a separate PR

check-vulnerabilities/ignored-safety.txt
RobPasMue: All this information is in the links... See https://actions.docs.ansys.com/version/dev/vulnerability-actions/index.html

image

I still don't understand why you all want to duplicate the information @jorgepiloto @SMoraisAnsys @clatapie

SMoraisAnsys: @RobPasMue The main purpose is to avoid indirection when one wants to retrieve information. If you are reading the source code it simply avoids having to look at the online documentation. Also, I had network issues yesterday and it would have saved me some time haha.
To fix the file, I assume that putting the comment before each ID should be enough (currently it's at the end of the line) and the docs could be fixed with a simple `{% if v_id != '' and not v_id.startswith('#') %}` right ?

@clatapie I do agree with Roberto's statement on the title. I didn't read the title (my bad...) and though that the purpose of this PR was to extend the text file used to ignore vulnerability with short explanation on each vulnerability.

RobPasMue: Fine with the requirement - we can add the comment then for "offline" and quick check conditions. I do prefer the comments as they are right now rather than before the vulnerability ID. @clatapie you would have to do some ``strip`` operations IMO.

Also, bare in mind that we are also using this TXT file here:

https://github.com/ansys/actions/blob/7419c6df33dbbab62318f8ac304a6a481776476c/check-vulnerabilities/action.yml#L210-L219

My point is that it is not as simple as it seems 😄 - it's doable, it's just not as simple as adding some comments in the txt file. And yeah, better in a different PR so that this one does not get blocked

RobPasMue (approved):

PR #507, opened 2024-06-17
fix: only compress content inside html folder

RobPasMue:
Current Windows implementation was zipping the entire "doc/build..." folder structure. We are only interested in the contents inside the "html" folder

### DIFF
--- a/_doc-build-windows/action.yml
+++ b/_doc-build-windows/action.yml
@@ -369,7 +369,8 @@ runs:
       if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       shell: powershell
       run: |
-        zip -r documentation-html.zip ${{ env.EXPECTED_BUILD_DIR }}\html
+        cd ${{ env.EXPECTED_BUILD_DIR }}\html; zip -r documentation-html.zip *; cd ..\..\..
+        mv ${{ env.EXPECTED_BUILD_DIR }}\html\documentation-html.zip ${{ env.EXPECTED_DOWNLOAD_DIR }}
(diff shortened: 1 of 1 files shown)


RobPasMue:
Requesting specific review from @SMoraisAnsys

SMoraisAnsys (approved):
LGTM, thanks for noticing this behavior !
I think we could also use `Compress-Archive -Path ${{ env.EXPECTED_BUILD_DIR }}\html\* -DestinationPath documentation-html.zip` but you solutions works fine !

RobPasMue:
> LGTM, thanks for noticing this behavior ! I think we could also use `Compress-Archive -Path ${{ env.EXPECTED_BUILD_DIR }}\html\* -DestinationPath documentation-html.zip` but you solutions works fine !

I tried to be as little invasive as possible with my change - I figured that if you used "zip" directly it was for a reason. Honestly, I'd even delegate all this to the zip action we typically use https://github.com/vimtor/action-zip

SMoraisAnsys:
> > LGTM, thanks for noticing this behavior ! I think we could also use `Compress-Archive -Path ${{ env.EXPECTED_BUILD_DIR }}\html\* -DestinationPath documentation-html.zip` but you solutions works fine !
> 
> I tried to be as little invasive as possible with my change - I figured that if you used "zip" directly it was for a reason. Honestly, I'd even delegate all this to the zip action we typically use https://github.com/vimtor/action-zip

I saw `zip` being used in other actions and/or repos so I went for it. I would also have used `https://github.com/vimtor/action-zip` otherwise :')

Issue #459, opened 2024-04-17
Python Organzation


da1910:
There are now several actions that have dependences and extensive python code as multiline literals in a workflow file. These can't be tested, and won't be properly covered by dependabot etc for their dependencies.

Can we please get these into a repo, built into a proper package and distributed via pip correctly? Something like `ansys.actions.scripts` or similar?

jorgepiloto:
Thanks for raising this discussion, @da1910. I think this is a great approach. In addition, I would like to keep such package in this same repository.

This would also allow us to take advantage of tools like [semver](https://python-semver.readthedocs.io/en/latest/advanced/convert-pypi-to-semver.html) to solve old issues like #217.

jorgepiloto:
@pyansys-automation never close issue

PR #515, opened 2024-07-08
feat: doc-changelog conventional commits

klmcadams:
By default, the doc-changelog action will check if there is a conventional commit type in the PR title (both lowercase and uppercase types are allowed). If you want to use labels to create the changelog fragment, you have to add the `use-labels: true` line to the doc-changelog job.

This will break repositories until the pyproject.toml files are updated to include the additional sections listed in the doc-changelog documentation

Conventional commit tests:

[Fails when there is no conventional commit type in the title](https://github.com/ansys-internal/actions-sandbox/actions/runs/9840920384/job/27166328031?pr=56)
[Passes when there is an all caps conventional commit type in the title](https://github.com/ansys-internal/actions-sandbox/actions/runs/9841008696/job/27166646415?pr=57#step:2:377)
[Passes when there is an all lowercase conventional commit type in the title](https://github.com/ansys-internal/actions-sandbox/actions/runs/9841021730/job/27166695270?pr=58#step:2:377)

Labels tests:

[Documentation label](https://github.com/ansys-internal/actions-sandbox/actions/runs/9841269350/job/27167526398#step:2:333)
[Test label](https://github.com/ansys-internal/actions-sandbox/actions/runs/9841547875/job/27168436603#step:2:333)

### DIFF
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -140,8 +140,8 @@ runs:
       shell: bash
       run: |
         pr_title="${{ env.PR_TITLE }}"
-        echo "${pr_title%%:*}"
-        echo CC_TYPE='"${pr_title%%:*}"' >> $GITHUB_ENV
+        cc_type="${pr_title%%:*}"
+        echo CC_TYPE=$cc_type >> $GITHUB_ENV
 
     - name: "Get labels in the pull request"
       if: ${{ inputs.use-labels == 'true' }}
@@ -166,7 +166,7 @@ runs:
         import os
 
         # Get conventional commit type from env variable
-        cc_type = "${{ env.CC_TYPE }}".lower()
+        cc_type = ${{ env.CC_TYPE }}.lower()
 
         print(cc_type)
 
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -273,7 +273,9 @@ runs:
         use_labels = "${{ inputs.use-labels }}"
 
         # Capitalize first letter of string, so it becomes True or False
-        use_labels = use_labels[0].upper() + use_labels[1:]
+        use_labels = bool(use_labels[0].upper() + use_labels[1:])
+
+        print(use_labels)
 
         # If not using label, remove conventional commit type from title
         if not use_labels:
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -105,7 +105,16 @@ runs:
     #   with:
     #     token: ${{ inputs.token }}
 
+    - name: "Check pull-request title follows conventional commits style"
+      id: lowercase
+      if: ${{ inputs.use-labels == 'false' && ((github.event_name == 'pull_request_target') || (github.event_name == 'pull_request')) }}
+      uses: amannn/action-semantic-pull-request@v5
+      env:
+        GITHUB_TOKEN: ${{ inputs.token }}
+      continue-on-error: true
+
     - name: "Check pull-request title follows conventional commits style with upper case"
+      id: caps
       if: ${{ inputs.use-labels == 'false' && ((github.event_name == 'pull_request_target') || (github.event_name == 'pull_request')) }}
       uses: amannn/action-semantic-pull-request@v5
       env:
@@ -125,11 +134,9 @@ runs:
           TEST
       continue-on-error: true
 
-    - name: "Check pull-request title follows conventional commits style"
-      if: ${{ inputs.use-labels == 'false' && ((github.event_name == 'pull_request_target') || (github.event_name == 'pull_request')) }}
-      uses: amannn/action-semantic-pull-request@v5
-      env:
-        GITHUB_TOKEN: ${{ inputs.token }}
+    - name: "Check failures"
+      if: steps.lowercase.outcome == 'failure' && steps.caps.outcome == 'failure'
+      run: exit 1
 
     # - name: "Get labels in the pull request"
     #   if: ${{ inputs.use-labels == 'true' }}
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -137,10 +137,20 @@ runs:
       if: ${{ inputs.use-labels == 'false' }}
       env:
         PR_TITLE: ${{ github.event.pull_request.title }}
-      shell: bash
+      shell: python
       run: |
-        cc_type=$(echo ${{ env.PR_TITLE }} | cut -d: -f1)
-        echo CC_TYPE='"'$cc_type'"' >> $GITHUB_ENV
+        import os
+
+        pr_title = f"""{os.environ.get('PR_TITLE')}"""
+        colon_index = pr_title.index(":")
+        cc_type = pr_title[:4]
+        print(cc_type)
+
+        # Get the GITHUB_ENV variable
+        github_env = os.getenv('GITHUB_ENV')
+
+        with open(github_env, "a") as f:
+            f.write(f"CC_TYPE={cc_type}")
 
     - name: "Get labels in the pull request"
       if: ${{ inputs.use-labels == 'true' }}
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -139,7 +139,7 @@ runs:
         PR_TITLE: ${{ github.event.pull_request.title }}
       shell: bash
       run: |
-        pr_title=${{ env.PR_TITLE }}
+        pr_title="${{ env.PR_TITLE }}"
         echo "${pr_title%%:*}"
         echo CC_TYPE='"${pr_title%%:*}"' >> $GITHUB_ENV
 
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -141,7 +141,7 @@ runs:
       run: |
         pr_title="${{ env.PR_TITLE }}"
         cc_type="${pr_title%%:*}"
-        echo CC_TYPE=$cc_type >> $GITHUB_ENV
+        echo CC_TYPE='"$cc_type"' >> $GITHUB_ENV
 
     - name: "Get labels in the pull request"
       if: ${{ inputs.use-labels == 'true' }}
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -140,8 +140,7 @@ runs:
       shell: bash
       run: |
         pr_title=${{ env.PR_TITLE }}
-        cc_type="${ ${{ env.PR_TITLE }}%%:* }"
-        # "${pr_title%%:*}"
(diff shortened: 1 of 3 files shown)


doc-changelog/action.yml
```diff
@@ -91,7 +99,60 @@ runs:
...
+          CI
+          DOCS
+          FEAT
+          FIX
+          PERF
+          REFACTOR
+          REVERT
+          STYLE
+          TEST
+      continue-on-error: true
```
SMoraisAnsys, line 129: Here, I'm wondering what would be best between running both jobs (like now) or running the most coherent one - e.g. determining if the first letter is lower case or not and run a single job accordingly. This would add complexity to the workflow but diminish resource consumption.

doc-changelog/action.yml
```diff
@@ -91,7 +99,60 @@ runs:
...
+
+        pr_title = f"""{os.environ.get('PR_TITLE')}"""
+        colon_index = pr_title.index(":")
+        cc_type = '"'+pr_title[:colon_index]+'"'
+
+        # Get the GITHUB_ENV variable
+        github_env = os.getenv('GITHUB_ENV')
+
+        with open(github_env, "a") as f:
+            f.write(f'CC_TYPE={cc_type}')
```
SMoraisAnsys, line 152: Can you rewrite this into a python file and run it like in `doc-build/parse_doc_conf.py`.
If possible, do the same for other python parts of this action !
This would be another step to #459 and we might want to leverage this in other actions.

SMoraisAnsys (changes requested):
Great job, I left a minor comment on the python code associated to this PR

doc-changelog/action.yml
klmcadams: Ohh yeah, great point. I think it makes more sense to check the first letter and run the test based on which it is

doc-changelog/action.yml
```diff
@@ -91,7 +99,49 @@ runs:
...
+        PR_TITLE: ${{ github.event.pull_request.title }}
+      shell: python
+      run: |
+        import sys
+        sys.path.insert(1, '${{ github.action_path }}/../doc-changelog/')
+
+        from parse_pr_title import get_first_letter_case
+
+        pr_title = """${{ env.PR_TITLE }}"""
+        get_first_letter_case(pr_title)
```
klmcadams, line 114: I did this instead of using a bash shell and running `python ${{ github.action_path }}/../doc-changelog/parse_pr_title.py` because of the PR_TITLE. I have to put triple quotes around the env.PR_TITLE in case the title contains double quotes, single quotes, and/or backticks

SMoraisAnsys: Good job in thinking about quotes and backticks.

Wouldn't something like 
```python
pr_title = f"""{os.getenv("PR_TITLE")}"""
```
do the job though ? If I don't understand the problem, can you tell me what I'm missing here ?

klmcadams: @SMoraisAnsys That would work too, but I think it would do the same thing as line 113. If env.PR_TITLE is 

docs: pr's title with "double quotes",   `backticks`, and ``double backticks``!     

Then I have to wrap it in triple quotes when I set it equal to the pr_title variable. 

In bash, I wasn't able to pass in the env.PR_TITLE as an argument to this command `python ${{ github.action_path }}/../doc-changelog/parse_pr_title.py --get-first-letter-case --pr_title ${{ env.PR_TITLE }}` because I couldn't find a way to wrap the env.PR_TITLE in the triple quotes equivalent for bash (if that's even a thing)

klmcadams: so I think the issue down the line would be, I make the code above into a python script and run the command in bash, but then we run into the issue of how to pass the env.PR_TITLE value with special characters into the argument

doc-changelog/action.yml
```diff
@@ -91,7 +99,49 @@ runs:
...
+      uses: ansys/actions/commit-style@v6
+      with:
+        token: ${{ inputs.token }}
+
+    - name: "Check pull-request title follows conventional commits style with upper case"
+      if: ${{ env.FIRST_LETTER == 'uppercase' }}
+      uses: ansys/actions/commit-style@v6
+      with:
+        token: ${{ inputs.token }}
+        use-upper-case: true
```
RobPasMue, line 127: If ``use-labels == 'true'`` we should skip all these steps, right?

RobPasMue: Or the same thing the other way around... these steps should only execute if ``use-labels == 'false'``

doc-changelog/action.yml
```diff
@@ -106,44 +156,29 @@ runs:
...
-        # Get the GITHUB_ENV variable
-        github_env = os.getenv('GITHUB_ENV')
-
-        # Append the PR_LABEL with its value to GITHUB_ENV
-        # For example, PR_LABEL="added" if the PR had an "enhancement" label
-        with open(github_env, "a") as f:
-            f.write(f"PR_LABEL={get_changelog_section(pr_labels, existing_labels)}")
+        import sys
+        sys.path.insert(1, '${{ github.action_path }}/../doc-changelog/')
+
```
RobPasMue, line 177: I really don't like ``sys.path.insert`` but I see no other way around it honestly in this case 😄 - except when we move all these Python scripts to a separate package (see #459)

doc/source/doc-actions/examples/doc-changelog-basic.yml
```diff
@@ -20,3 +20,4 @@ changelog-fragment:
     - uses: ansys/actions/doc-changelog@{{ version }}
       with:
         token: ${{ '{{ secrets.PYANSYS_CI_BOT_TOKEN }}' }}
+        # use-labels: true  # uncomment this line to use labels instead of conventional commits
```
RobPasMue, line 23: I think we should allow for backwards compatibility until next version and change the logic... We should implement both ``use-conventional-commits`` and ``use-labels`` and ``use-labels`` should default to true, whereas ``use-conventional-commits`` should default to false - temporarily.

Next major release we change the behavior. But this way we can live with a minor release and check how this works without breaking everybody's workflows (in case they don't use conventional commits)

doc/source/migrations/docs-changelog-setup.rst
```diff
@@ -257,15 +262,20 @@ Also, replace ``ansys.<product>.<library>`` with the name under ``tool.flit.modu
...
     [[tool.towncrier.type]]
     directory = "miscellaneous"
     name = "Miscellaneous"
     showcontent = true
 
+    [[tool.towncrier.type]]
+    directory = "test"
+    name = "Test"
+    showcontent = true
+
```
RobPasMue, line 278: For conventional commits we should template all this section and inject it on runtime IMO. That way we prevent users having to define it on every single repo.

RobPasMue (changes requested):

doc-changelog/parse_pr_title.py
```diff
@@ -0,0 +1,204 @@
...
+    # Remove extra whitespace
+    clean_title = clean_title.strip()
+
+    # Add backslash in front of backtick and double quote
+    clean_title = clean_title.replace("`", "\\`").replace('"', '\\"')
+
+    save_env_variable("CLEAN_TITLE", clean_title)
+
+
+def add_towncrier_config(org_name, repo_name, web_release_notes):
```
klmcadams, line 123: @RobPasMue, is this what you were thinking for the template? The web_release_notes is a bool whether or not the person is using or wants to use the release note configuration in documentation vs. the CHANGELOG.md file. If the [tool.towncrier] section doesn't exist, it'll add it based on the web_release_notes bool

doc-changelog/action.yml
SMoraisAnsys: I think using multiple python files would solve this problem as well as the one with sys path setup. However, they will likely contain copy pasted line of codes and it will be a pain to maintain. This will change if we ever extract all of this into a single python package !

doc-changelog/parse_pr_title.py
RobPasMue: Yep exactly but you could simplify it. Just accept a boolean to either include the entire towncrier section or not. You don't have to do all the conditional checks I believe. Either you add the entire section or you don't add it based on the users input ``use_pyansys_towncrier_default_config`` or something like that

doc-changelog/parse_pr_title.py
```diff
@@ -0,0 +1,345 @@
...
+            if module != ("DNE" or ""):
+                name = module.get("name", "DNE")
+                # If [tool.flit.module] name exists, create the package string
+                if name != ("DNE" and ""):
+                    towncrier_config_sections["package"] = f'"{name}"'
+
+        if default_config:
+            # If there is no towncrier configuration information or if [[tool.towncrier.type]]
+            # is the only towncrier information in the pyproject.toml file
+            if towncrier == "DNE" or len(towncrier) == 1:
```
klmcadams, line 261: I have to check this because there can't be two [tool.towncrier] sections. If for some reason someone only has the [[tool.towncrier.type]] sections in their pyproject.toml, then len(towncrier) is 1

doc-changelog/parse_pr_title.py
```diff
@@ -0,0 +1,345 @@
...
+                # Write the [tool.towncrier] section
+                write_towncrier_config_section(file, towncrier_config_sections, True)
+
+        if towncrier != "DNE":
+            # Get the existing [[tool.towncrier.type]] sections
+            types = towncrier.get("type", "DNE")
+            if types != "DNE":
+                remove_existing_types(types, changelog_sections)
```
klmcadams, line 269: I can only get the [[tool.towncrier.type]] sections if towncrier is not "DNE". Once we know the existing [[tool.towncrier.type]] types, then we can remove them from the changelog_sections list so that we're only appending the missing [[tool.towncrier.type]] sections in line 272

doc-changelog/parse_pr_title.py
```diff
@@ -0,0 +1,345 @@
...
+        # List containing changelog sections under each release
+        changelog_sections = [
+            "added",
+            "dependencies",
+            "documentation",
+            "fixed",
+            "maintenance",
+            "miscellaneous",
+            "test",
+        ]
```
klmcadams, line 226: These changelog sections all have to correspond to a [[tool.towncrier.type]] section, so even if someone says they don't want to use the default towncrier configuration, we still have to add the missing [[tool.towncrier.type]] sections so it doesn't fail to create the fragment.

If we want people to be able to define their own [[tool.towncrier.type]] sections, then I could read in these sections from their pyproject.toml and generate the changelog_sections list from that information. This would change the logic later in the script a little bit. Lmk what you think

RobPasMue (approved):
LGTM!

PR #517, opened 2024-07-11
refactor: remove non needed inputs

SMoraisAnsys:
Thanks @dipinknair for noticing that some inputs in the private actions were not needed.

This comes from the fact that two steps are performed before running any private action (checkout and setup-python) Since both private actions (windows and linux) are not involved in python setup and git install + clone, this PR removes the following inputs from the private actions:
- python-version
- use-python-cache
- checkout

### DIFF
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -42,20 +42,6 @@ inputs:
 
   # Required inputs
 
-  python-version:
-    description: >
-        Python version used for installing and running ``Sphinx``.
-    required: true
-    type: string
-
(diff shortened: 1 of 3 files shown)


RobPasMue (approved):
Yeah good catch! We delegate it to the main action now

RobPasMue:
Branch name not valid @SMoraisAnsys - please resubmit PR

MaxJPRey (approved):
LGTM.

SMoraisAnsys:
Closing for #519

Issue #524, opened 2024-07-24
Add index_name as a variable to release-pypi-private


dnwillia-work:
### 📝 Description of the feature

Right now the index name is hard coded to:

https://github.com/ansys/actions/blob/7d192c03875a47bd0eda6e3f7512bcbce4142e8c/release-pypi-private/action.yml#L110C8-L110C12

but I'd like to set that to something else.  The current setting should be the default.

### 💡 Steps for implementing the feature

Add a variable to the action that I can configure when using it.

PR #525, opened 2024-07-25
fix: Upload to various private indexes.

MaxJPRey:
Fix: #524 .

Previously, it was only possible to upload the release to the private Ansys pypi.
Now it should be possible to release to a different index.


RobPasMue (approved):
LGTM! Let me do a patch release

PR #529, opened 2024-07-31
feat: doc-changelog allow breaking changes symbol in conventional commit type

klmcadams:
Adjusted the logic to get the changelog section from the cc_type_changelog_dict when there's an exclamation point in the conventional commit type string from the PR title. It looks like amannn/action-semantic-pull-request [allows the breaking changes symbol (!)](https://github.com/amannn/action-semantic-pull-request/blob/main/README.md#examples)

### DIFF
--- a/doc-changelog/parse_pr_title.py
+++ b/doc-changelog/parse_pr_title.py
@@ -312,7 +312,7 @@ def write_towncrier_config_section(
             file.write(f"{key} = {value}\n")
 
 
-def remove_existing_types(types, changelog_sections: list):
+def remove_existing_types(types: list, changelog_sections: list):
     """Remove the existing [[tool.towncrier.types]] from the changelog_sections list.
 
     Parameters
@@ -325,8 +325,9 @@ def remove_existing_types(types, changelog_sections: list):
     for group in types:
         # Remove changelog section if it exists under [[tool.towncrier.type]] so that
         # only missing sections are appended to the pyproject.toml file
-        if group["directory"] in changelog_sections:
-            changelog_sections.remove(group["directory"])
+        section = group["directory"]
+        if section in changelog_sections:
+            changelog_sections.remove(section)
 
 
 def write_missing_types(changelog_sections: list, file):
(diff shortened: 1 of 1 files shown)


doc-changelog/parse_pr_title.py
```diff
@@ -321,7 +323,9 @@ def remove_existing_types(types, changelog_sections: list):
     for group in types:
         # Remove changelog section if it exists under [[tool.towncrier.type]] so that
         # only missing sections are appended to the pyproject.toml file
-        changelog_sections.remove(group["directory"])
+        section = group["directory"]
```
RobPasMue, line 326:
```suggestion
        section = group.get("directory")
```
I'm guessing it's a dict. If for whatever reason dictionary is not defined, this will immediately crash. Better to use get. If you are sure "directory" is always going to exist, you can disregard this comment.

RobPasMue (approved):
Minor comment - otherwise LGTM

doc-changelog/parse_pr_title.py
```diff
@@ -101,8 +101,10 @@ def changelog_category_cc(cc_type: str):
         "ci": "maintenance",
     }
 
-    # Get the changelog section based on the conventional commit type
-    changelog_section = cc_type_changelog_dict[cc_type]
+    for key, value in cc_type_changelog_dict.items():
+        if key in cc_type:
+            # Get the changelog section based on the conventional commit type
+            changelog_section = cc_type_changelog_dict[key]
```
SMoraisAnsys, line 107:
```suggestion
    for key, value in cc_type_changelog_dict.items():
        if key in cc_type:
            # Get the changelog section based on the conventional commit type
            changelog_section = value
            break
```

SMoraisAnsys: Just leveraging `value` and breaking when a key matches.

SMoraisAnsys (changes requested):
Just a minor comment, LGTM otherwise.
Great enhancement !

doc-changelog/parse_pr_title.py
RobPasMue: Ah true, we already have the value. Good catch

RobPasMue: Completely overseen on my side 😄

SMoraisAnsys (approved):

PR #533, opened 2024-08-07
fix: changelog actions docs missing

### LINKED ISSUE #531: Missing ``doc-*`` actions in the documentation
If you go here https://actions.docs.ansys.com/version/stable/doc-actions/index.html#

Only 4 doc actions appear. We should include the remaining ones.

Also, it would be worth it to make a reusable template for the jinja statements we use. They are basically the same everywhere

RobPasMue:
Closes #531 

Also adding proper formatting to all inputs.

### DIFF
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -54,14 +54,14 @@ inputs:
     type: string
 
   dependencies:
-    description: >
+    description: |
       String of system dependencies to be installed before building the
       documentation of the project.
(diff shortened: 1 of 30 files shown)


RobPasMue:
Once this gets merged we should do a new major release. The changelog actions have been modified extensively

jorgepiloto (approved):
Many thanks!

Revathyvenugopal162 (approved):

PR #536, opened 2024-08-08
fix: links redirect to search `js` for the landing page

Revathyvenugopal162:
related with the latest ansys-sphinx-theme update.

PR #539, opened 2024-08-09
fix: indentation with docs and redirection to dev for static js file in docs deploy action

Revathyvenugopal162:
fix issue with redirection to dev in #536 
fix indentation of admonitions

### DIFF
--- a/doc-deploy-stable/action.yml
+++ b/doc-deploy-stable/action.yml
@@ -447,11 +447,13 @@ runs:
           sed -i 's/href="\([^:"]*\)"/href="version\/stable\/\1"/g' index.html
           sed -i 's/src="\([^:"]*\)"/src="version\/stable\/\1"/g' index.html
           sed -i 's/action="search.html"/action="version\/stable\/search.html"/g' index.html
+          sed -i 's|script\.src = ".*_static/js/meilisearch_theme_wrap\.js"|script.src = "version/stable/_static/js/meilisearch_theme_wrap.js"|g' index.html
         elif [[ -f 'version/dev/index.html' ]]; then
           cp version/dev/index.html index.html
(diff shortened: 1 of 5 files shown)


doc-deploy-stable/action.yml
```diff
@@ -453,7 +453,7 @@ runs:
           sed -i 's/href="\([^:"]*\)"/href="version\/dev\/\1"/g' index.html
           sed -i 's/src="\([^:"]*\)"/src="version\/dev\/\1"/g' index.html
           sed -i 's/action="search.html"/action="version\/dev\/search.html"/g' index.html
-          sed -i 's|script\.src = ".*_static/js/meilisearch_theme_wrap\.js"|script.src = "version/stable/_static/js/meilisearch_theme_wrap.js"|g' index.html
+          sed -i 's|script\.src = ".*_static/js/meilisearch_theme_wrap\.js"|script.src = "version/dev/_static/js/meilisearch_theme_wrap.js"|g' index.html
```
RobPasMue, line 456: Why should it be ``dev`` here instead of ``stable``? This is the doc-deploy-stable action

RobPasMue (changes requested):

doc-deploy-stable/action.yml
Revathyvenugopal162: We were using the same script in both the stable and development deployments, which I believe was added during the testing phase. I've since removed the script in [2d665bd](https://github.com/ansys/actions/pull/539/commits/2d665bde7c97f350feffe73a7129561d14ea3178)

RobPasMue (approved):

doc-deploy-stable/action.yml
RobPasMue: Okay makes sense now, thanks @Revathyvenugopal162

PR #553, opened 2024-09-02
fix: needs-quarto is required parameter

SMoraisAnsys:
While updating #543 conflict, I found out that the parameter `needs-quarto` was not correctly defined in private actions.


RobPasMue (approved):
LGTM! Good catch @SMoraisAnsys

PR #555, opened 2024-09-04
fix: add [dependabot-skip] to doc-changelog commit message

### LINKED ISSUE #485: Changelog action does not play nicely with dependabot
I don't know that there is actually a nice solution for this, but its worth flagging up.

Dependabot will only update its own PRs if nobody else has touched them (for example if a later version of a package becomes available). This means that if you use the docs/changelog action (which is marvellous and we should definitely use), then dependabot will see that someone (the pyansys-ci-bot) has edited its PR and refuse to maintain it.

This isn't a dealbreaker most of the time, as it's not especially common for a package to receive multiple releases in a short time, but it is annoying. Additionally it's worth thinking about whether it makes sense to document every separate bump to a package version. Or does it make more sense to add the package bumps in an aggregated form at release time?

klmcadams:
Add [dependabot-skip] to the commit message in doc-changelog so dependabot ignores the commit. 

Closes #485


doc-changelog/action.yml
```diff
@@ -266,6 +266,6 @@ runs:
         if [ ! -z "$modified" ]; then
           echo "modified: $modified"
           # Commit and push fragment
-          git commit -m "chore: adding changelog file $fragment"
+          git commit -m "chore: adding changelog file $fragment [dependabot-skip]"
```
RobPasMue, line 269: Doesn't it have to go at the beginning? Maybe I'm confused

klmcadams: I think it can be anywhere, but I can move it to the front if it looks better

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#allowing-dependabot-to-rebase-and-force-push-over-extra-commits

RobPasMue: No worries, let's leave it there :)

RobPasMue (approved):

PR #557, opened 2024-09-05
fix: command context when no pyproject.toml

### LINKED ISSUE #544: Problem in determining the build backend in doc-build actions when no pyproject.toml file exist
At the moment there is no check on the existence of the `pyproject.toml` file in our doc build actions. However, this file is leveraged to decide which backend should be used in our CICD, see https://github.com/ansys/actions/blob/8643c5a882d18be06255fa8cc6e32bd9cc7e8cf1/_doc-build-linux/action.yml#L188 and https://github.com/ansys/actions/blob/8643c5a882d18be06255fa8cc6e32bd9cc7e8cf1/_doc-build-linux/action.yml#L198.

If this is a requirement then we should put it somewhere in the documentation. If that's not the case then we should fix the current action to add this check. Note that this would help projects which have no `pyproject.toml` file, e.g. projects mainly composed of examples. Currently, I'm not sure that they are able to use the `doc-build` action.

**Note**: speaking of projects without `pyproject.toml` file, we could use `pip` as default backend in those cases. What do you think ?

SMoraisAnsys:
This PR make the `doc-build` action compatible with repos without `pyproject.toml` file. This is typically the case for repositories dedicated to example, e.g. [pymapdl-examples](https://github.com/ansys/pymapdl-examples) or  [pymechanical-examples](https://github.com/ansys/pymechanical-examples).

Close #544

### DIFF
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -219,7 +219,7 @@ runs:
     - name: "Determine make command context"
       shell: bash
       run: |
-        if grep -q 'build-backend = "poetry\.core\.masonry\.api"' "pyproject.toml"; then
+        if [[ -f "pyproject.toml" ]] && grep -q 'build-backend = "poetry\.core\.masonry\.api"' "pyproject.toml"; then
           echo "SPHINX_BUILD_MAKE=$(echo 'poetry run -- make')" >> $GITHUB_ENV
         else
           echo "SPHINX_BUILD_MAKE=$(echo 'make')" >> $GITHUB_ENV
@@ -285,6 +285,7 @@ runs:
           Environment variable PDF_FILENAME set to ${{ env.PDF_FILENAME }}.
 
     # ------------------------------------------------------------------------
+
     - uses: ansys/actions/_logging@main
       if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       with:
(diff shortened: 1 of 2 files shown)


germa89 (approved):

Revathyvenugopal162:
We have ``skip-install`` for the libraries that don't have pyproject.toml. Example our dev guide https://github.com/ansys/pyansys-dev-guide/blob/ae61cd29db323dbfa6b62dd5dd36af18082d502a/.github/workflows/ci_cd.yml#L54. Isn't it enough for the repos without pyproject.toml file?

RobPasMue (approved):
Sounds good to me!

SMoraisAnsys:
> We have `skip-install` for the libraries that don't have pyproject.toml. Example our dev guide https://github.com/ansys/pyansys-dev-guide/blob/ae61cd29db323dbfa6b62dd5dd36af18082d502a/.github/workflows/ci_cd.yml#L54. Isn't it enough for the repos without pyproject.toml file? or perhaps we could combine both if they serve the same purpose?

It is not a problem when you install the doc dependencies. It is a problem that arises when the action tries to determine which command should be used to launch the documentation build (using `poetry` or not). I checked the repo `pyansys-dev-guide``and it seems that you have an error but it is not failing the workflow as you are running in Linux, see below

image

However, when I try to use it in a Windows environment here is what I get

image

Revathyvenugopal162 (approved):

PR #558, opened 2024-09-06
fix: use powershell in quarto check (windows)

SMoraisAnsys:
In private action `_doc-build-windows`, set step "Check Quarto Version" to use `shell: powershell` instead of `shell: bash`.


RobPasMue (approved):
Yep - some Windows system do not come with it. Thanks @SMoraisAnsys!

Revathyvenugopal162 (approved):

Issue #559, opened 2024-09-06
Errors when working in a VM


SMoraisAnsys:
Some action's behavior seem to not work well when using VMs. In fact, most actions leverage `actions/setup-python` and then use the installed environment to perform multiple things. While this work well on Github runners, it's not the case for our self-hosted runners (running in our VMs) as we may already have:
- multiple python version installed which make some step do not go as planned, see https://github.com/ansys-internal/pyaedt-examples/pull/230#issuecomment-2331507013;
- caching directories already existing when running an action;
- ...

A simple work around would be to use virtual environments to ensure that every time an action is called, all of its steps are performed inside of a box that will not impact future CICD execution.

jorgepiloto:
I know @RobPasMue implemented some action or repository that fixes this. Maybe we could think about adding that within the `setup-python` action.

RobPasMue:
Yeah this is something different. My problem was with the installation of Python on self-hosted runners (Windows) when using VMs.

What @SMoraisAnsys is showing is something different. In order to properly use Python environments inside VMs it is necessary to activate the virtual environment in every single step. That's what he is working on in #560 I believe. It's a mess... on GH runners this doesn't happen because the runners are ephemeral and you don't care about having a venv. But on self-hosted ones you need to make sure you are cleaning up and using a proper environment. Otherwise you enter into caching issues (cached by the machine itself) and so on.

SMoraisAnsys:
Yeah that's right. It's even more common that the VMs are shared among multiple projects, each having their own way of setting their environment. On top of adding the environment activation at each step, we also share the same behavior for `poetry` and `pip`. I had urgent things to handle so I didn't finished the PR but that should be ok in the coming days. Sry for the delay and thanks both of you for your attention on this issue !

SMoraisAnsys:
The PR #560 is ready.
@jorgepiloto On top of the previous changes previously mentioned, the PR is also "more compatible" with poetry. In a previous meeting @ludovicsteinbach mentioned that even for packages with `poetry` as build backend, the default approach of some actions (modified in this PR) was to use `pip` to install the package. On top of being not aligned with the build backend, this requires one to define two tokens instead of one when one wants to use a private pypi.

SMoraisAnsys:
Fixed by #560

PR #560, opened 2024-09-06
refactor: use virtual environment to be consistent

SMoraisAnsys:
Following the work on #543, it seems that there was a need for three things:
- better handling of unexpected behavior when working with VMs, see https://github.com/ansys/actions/issues/559;
- better coherence in how we deal with `poetry` and `pip` build backends;
- better handling of the corner case where poetry and the project have shared dependencies with different versions.

This PR aims at tackling those three points by enforcing the use of a virtual environment, even when working with `poetry`.

On top of those changes, the modified actions are now "more compatible" with `poetry`. Before, the default behavior was to use `python -m pip install .` while now we detect the build backend and use `poetry` if appropriate.

> **Note**: following @Andy-Grigg comments, we use a "hack" to ensure that poetry and the project are not installed in the same environment and thus don't clash with each other if they share common dependencies with different version. The idea is to use `pipx` to install poetry in a temp folder cleaned up before and after running each job. The only side effect is for VMs as `pipx` will still be here afterward.

### DIFF
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -272,11 +272,6 @@ runs:
           python -m pip install .[doc]
         fi
 
-    - name: "Move into doc foled"
-      shell: bash
-      run: |
-        cd doc
-
     # ------------------------------------------------------------------------
 
     - uses: ansys/actions/_logging@main
@@ -290,15 +285,15 @@ runs:
       shell: bash
       run: |
         source .venv/bin/activate
-        ${{ env.SPHINX_BUILD_MAKE }} html SPHINXOPTS="${{ inputs.sphinxopts }}"
-        ${{ env.SPHINX_BUILD_MAKE }} pdf
+        ${{ env.SPHINX_BUILD_MAKE }} -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
+        ${{ env.SPHINX_BUILD_MAKE }} -C doc pdf
         if [[ ${{ inputs.check-links }} == 'true' ]];
         then
-          ${{ env.SPHINX_BUILD_MAKE }} linkcheck SPHINXOPTS="${{ inputs.sphinxopts }}"
+          ${{ env.SPHINX_BUILD_MAKE }} -C doc linkcheck SPHINXOPTS="${{ inputs.sphinxopts }}"
         fi
         if [[ ${{ inputs.skip-json-build }} == 'false' ]];
         then
-          ${{ env.SPHINX_BUILD_MAKE }} json SPHINXOPTS="${{ inputs.sphinxopts }}"
+          ${{ env.SPHINX_BUILD_MAKE }} -C doc json SPHINXOPTS="${{ inputs.sphinxopts }}"
         fi
 
     - name: "Build HTML, PDF, and JSON documentation using xvfb"
@@ -306,15 +301,16 @@ runs:
       shell: bash
       run: |
         source .venv/bin/activate
-        xvfb-run ${{ env.SPHINX_BUILD_MAKE }} html SPHINXOPTS="${{ inputs.sphinxopts }}"
-        xvfb-run ${{ env.SPHINX_BUILD_MAKE }} pdf
+        cd doc
+        xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
+        xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc pdf
         if [[ ${{ inputs.check-links }} == 'true' ]];
         then
-          xvfb-run ${{ env.SPHINX_BUILD_MAKE }} linkcheck SPHINXOPTS="${{ inputs.sphinxopts }}"
+          xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc linkcheck SPHINXOPTS="${{ inputs.sphinxopts }}"
         fi
         if [[ ${{ inputs.skip-json-build }} == 'false' ]];
         then
-          xvfb-run ${{ env.SPHINX_BUILD_MAKE }} json SPHINXOPTS="${{ inputs.sphinxopts }}"
+          xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc json SPHINXOPTS="${{ inputs.sphinxopts }}"
         fi
 
     # ------------------------------------------------------------------------
@@ -360,7 +356,7 @@ runs:
       if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       shell: bash
       run: |
-        echo "EXPECTED_BUILD_DIR=_build" >> $GITHUB_ENV
+        echo "EXPECTED_BUILD_DIR=doc/_build" >> $GITHUB_ENV
 
     - name: Check expected build directory
       if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
@@ -436,14 +432,14 @@ runs:
       uses: actions/upload-artifact@v4
       with:
         name: documentation-html
-        path: _build/html
+        path: doc/_build/html
         retention-days: 7
 
     - name: "Upload PDF documentation artifact"
       uses: actions/upload-artifact@v4
       with:
         name: documentation-pdf
-        path: _build/latex/*.pdf
+        path: doc/_build/latex/*.pdf
         retention-days: 7
 
     - name: "Upload JSON documentation artifact"
@@ -451,5 +447,5 @@ runs:
       if: inputs.skip-json-build == 'false'
       with:
         name: documentation-json
-        path: _build/json
+        path: doc/_build/json
         retention-days: 7
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -179,56 +179,111 @@ runs:
         sudo apt-get update
         sudo apt-get install -y poppler-utils
 
+    # ------------------------------------------------------------------------
+
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Determine context.
+
+    - name: "Determine Github environnement variables"
+      shell: bash
+      run: |
+        if [[ -f "pyproject.toml" ]] && grep -q 'build-backend = "poetry\.core\.masonry\.api"' "pyproject.toml"; then
+          echo "BUILD_BACKEND=$(echo 'poetry')" >> $GITHUB_ENV
+          echo "SPHINX_BUILD_MAKE=$(echo 'poetry run -- make')" >> $GITHUB_ENV
+        else
+          echo "BUILD_BACKEND=$(echo 'pip')" >> $GITHUB_ENV
+          echo "SPHINX_BUILD_MAKE=$(echo 'make')" >> $GITHUB_ENV
+
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Build backend: ${{ env.BUILD_BACKEND }}
+          Sphinx build make: ${{ SPHINX_BUILD_MAKE }}
+
+    # ------------------------------------------------------------------------
+
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Set up python to build the documentation.
+
+    - name: "Set poetry environment variable(s) (if required)"
+      shell: bash
+      if: env.BUILD_BACKEND == 'poetry'
+      run: |
+        # For projects using poetry, do not use a virtual environment.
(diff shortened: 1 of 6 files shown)


_doc-build-linux/action.yml
```diff
@@ -179,56 +179,111 @@ runs:
...
-    - name: "Check if requirements.txt file exists"
+    # ------------------------------------------------------------------------
+
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Determine context.
+
+    - name: "Determine Github environnement variables"
```
MaxJPRey, line 190:
```suggestion
    - name: "Determine GitHub environnement variables"
```

_doc-build-windows/action.yml
```diff
@@ -211,47 +211,95 @@ runs:
...
+
+    - name: "Determine Github environnement variables"
+      shell: powershell
+      run: |
+        if ((Test-Path "pyproject.toml") -and (Get-Content "pyproject.toml" | Select-String -Pattern 'build-backend = "poetry\.core\.masonry\.api"')) {
+          echo "BUILD_BACKEND=$(echo 'poetry')" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          echo "SPHINX_BUILD_MAKE=$(echo 'poetry run -- doc\make.bat')" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        } else {
+          echo "BUILD_BACKEND=$(echo 'pip')" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          echo "SPHINX_BUILD_MAKE=$(echo 'doc\make.bat')" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
RobPasMue, line 228: Why the change of location for Windows? On Linux you were assuming the make commands to be ran at "doc" level. I would expect the same behavior on both actions.

RobPasMue (approved):
Overall LGTM - just a minor comment

_doc-build-windows/action.yml
```diff
@@ -211,47 +211,95 @@ runs:
 
     # ------------------------------------------------------------------------
 
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Determine context.
+
+    - name: "Determine Github environnement variables"
```
SMoraisAnsys, line 220:
```suggestion
    - name: "Determine GitHub environnement variables"
```

_doc-build-windows/action.yml
```diff
@@ -211,47 +211,95 @@ runs:
 
     # ------------------------------------------------------------------------
 
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Determine context.
+
+    - name: "Determine GitHub environnement variables"
```
SMoraisAnsys, line 220:
```suggestion
    - name: "Determine GitHub environment variables"
```

_doc-build-windows/action.yml
SMoraisAnsys: The thing is that on Linux we leverage `make` with option `-C` to "change directory before doing anything". We can't do the same with Windows as we are directly calling the `make.bat` file.
If you want to have the same behavior for both actions, I can add a "move" step before building the documentation. Would you prefer that ?

RobPasMue: In my experience... context matters when running these batch script files. Sometimes they are not properly adapted to be ran from a different directory (which is what you are doing on the Windows version). I'd recommend changing directories, yeah.

SMoraisAnsys: Fine by me, in that case I'll avoid using `-C` with linux and also move to the `doc` folder. If you have any other feedback, let me now !

RobPasMue: ``make -C`` is effectively changing directory for you. so that's fine. You do not need to do it yourself on Linux. See https://www.man7.org/linux/man-pages/man1/make.1.html

make.bat is a batch script with which you are trying to mimic the Make GNU program. So, that's why the context is important in that case

SMoraisAnsys: While starting the modifications, I noticed that the steps associated to assets would be impacted by the changes.
Currently, this steps define `EXPECTED_BUILD_DIR=doc/_build` for Linux and `EXPECTED_BUILD_DIR=doc\_build` for Windows. While I do agree with your previous comment, I'd rather perform a manual folder move with `cd doc` in both Windows and Linux. Otherwise we would end up with "different" environment variable which might become confusing later on. @RobPasMue Is that fine with you ?

RobPasMue: Sure, go ahead

doc-build/action.yml
```diff
@@ -220,7 +225,7 @@ runs:
 
     - name: Documentation build (Windows)
       if: ${{ runner.os == 'Windows' }}
-      uses: ansys/actions/_doc-build-windows@main
+      uses: ansys/actions/_doc-build-windows@fix/inconsistent-behavior
```
SMoraisAnsys, line 228:
```suggestion
      uses: ansys/actions/_doc-build-windows@main
```

tests-pytest/action.yml
```diff
@@ -101,31 +101,72 @@ runs:
...
+        echo "POETRY_VIRTUALENVS_CREATE=false" >> $GITHUB_ENV
+
     - name: "Set up Python"
       uses: ansys/actions/_setup-python@main
       with:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
 
-    - name: "Upgrade pip and disable virtual environment (for poetry only)"
+    - name: Create virtual environment
```
SMoraisAnsys, line 153:
```suggestion
    - name: "Create virtual environment"
```

_doc-build-windows/action.yml
SMoraisAnsys: After testing I'll keep `make -C` for Linux (previous behavior). I though that it might be confusing with `EXPECTED_BUILD_DIR` but I forgot the fact that every step of the job starts at the root level. Even if in a previous step I did `cd doc` :)

SMoraisAnsys:
This worked as expected for:
- https://github.com/ansys/pyansys-sound/actions/runs/10937085524
- https://github.com/ansys/pyconceptev/actions/runs/10936996838?pr=122
- currently running for https://github.com/ansys-internal/pyaedt-examples/actions/runs/10917092741?pr=230

RobPasMue:
You sure had some fun in this PR @SMoraisAnsys 😄

_doc-build-linux/action.yml
```diff
@@ -185,56 +185,112 @@ runs:
...
+      with:
+        level: "INFO"
+        message: >
+          Determine context.
+
+    - name: "Determine GitHub environment variables"
       shell: bash
       run: |
-        echo "EXISTS_DOC_REQUIREMENTS=$(if [ -f ${{ inputs.requirements-file }} ]; then echo 'true'; else echo 'false'; fi)" >> $GITHUB_ENV
+        if [[ -f "pyproject.toml" ]] && grep -q 'build-backend = "poetry\.core\.masonry\.api"' "pyproject.toml"; then
```
RobPasMue, line 199: Glad you also took into consideration setup.py projects (or even no-package doc projects using raw reqs files).

_doc-build-linux/action.yml
```diff
@@ -250,6 +306,8 @@ runs:
       if: inputs.requires-xvfb == 'true'
       shell: bash
       run: |
+        source .venv/bin/activate
+        cd doc
         xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
         xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc pdf
```
RobPasMue, line 312:
```suggestion
        xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
        xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc pdf
```
If you are navigating to the "doc" folder you don't need ``-C doc``. Or viceversa. I took the second approach (not navigating) but up to you :)

RobPasMue (approved):
LGTM! Some minor comments

_doc-build-linux/action.yml
```diff
@@ -185,56 +185,112 @@ runs:
...
       run: |
+        source .venv/bin/activate
         python -m pip install -r ${{ inputs.requirements-file }}
 
+    - name: "Install poetry (if required)"
+      shell: bash
+      if: env.BUILD_BACKEND == 'poetry'
+      run: |
+        source .venv/bin/activate
+        python -m pip install poetry
```
Andy-Grigg, line 257: Installing poetry into the same virtual environment that the project is installed into isn't recommended. If poetry and the running package have shared dependencies with different versions, poetry tries to upgrade/downgrade its dependencies while it's running, which often fails, especially on Windows.

The workaround we use is to use `pipx` to install poetry. Something like:

```yaml
      - name: "Add pipx/bin dir to path"
        run: echo "${{ runner.temp }}/pipx/bin" >> $GITHUB_PATH

      - name: "Install poetry and create empty virtual environment"
        run: |
          python -m pip install pipx --user
          python -m pipx install poetry
        env:
          PIPX_BIN_DIR: ${{ runner.temp }}/pipx/bin
          PIPX_HOME : ${{ runner.temp }}/pipx/home
```

Installing pipx into the `${{ runner.temp }}` directory means that the pipx installation will automatically be cleaned up between jobs on a VM. Adding the bin directory to the path first means pipx shouldn't complain that the poetry executable isn't on path.

This way, poetry is kept separate to the project, and they are free to use different versions of the same shared dependencies.

_doc-build-windows/action.yml
```diff
@@ -212,47 +212,95 @@ runs:
...
       if: ${{ env.EXISTS_DOC_REQUIREMENTS == 'true' }}
       run: |
+        .venv\Scripts\Activate.ps1
         python -m pip install -r ${{ inputs.requirements-file }}
 
+    - name: "Install poetry (if required)"
+      shell: powershell
+      if: env.BUILD_BACKEND == 'poetry'
+      run: |
+        .venv\Scripts\Activate.ps1
```
Andy-Grigg, line 281: Same here. I won't call out every location poetry is installed, but it should always be installed in a separate environment. pipx and runner.temp is a convenient way of doing that, especially on persistent infrastructure.

Andy-Grigg (changes requested):
Poetry should be installed in a separate virtual environment to the project that's being tested/built, to avoid situations in which both poetry and the project have dependencies on different versions of the same package. I've included an approach that uses pipx to get around this.

@ludovicsteinbach I know you originally raised this topic about poetry support across all actions. Can you take a look as well please?

_doc-build-linux/action.yml
Andy-Grigg: This _does_ mean that pipx is installed outside of any virtual environment, but I think that's only an issue if you really care about using a specific version of pipx between different jobs, which I imagine is a pretty rare requirement.

Andy-Grigg (approved):
The changes look good to me. Thanks!

SMoraisAnsys:
> The changes look good to me. Thanks!

I need to double check with some testing though :) Thanks again for the feedback !

SMoraisAnsys:
Test closing this PR due to workflow problem

code-style/action.yml
```diff
@@ -130,12 +195,14 @@ runs:
...
       shell: bash
       run: |
+        source .venv/bin/activate
         python -m pip install pre-commit
         pre-commit install
 
     - name: "Run pre-commit"
       shell: bash
       run: |
+        source .venv/bin/activate
```
jorgepiloto, line 205: Do we need to activate the venv every time? Just curious if this is a consequence when using Poetry.

jorgepiloto (approved):
Thanks for the big effort here, @SMoraisAnsys. Approving. Just left one comment.

code-style/action.yml
SMoraisAnsys: We have to do it because we don't want to pollute / use the system installation. Otherwise, we might end up installing in the wrong place which leads to the issues in #559.
Since we work with a virtual environment, you have to activate it at each step you require it because each step is like a fresh shell. This means that not doing it in a step `B` leads to working with the system instead of your virtual environment, even if you activated it in the previous step `A`.

Andy-Grigg: Alternatively, you can run the python binary directly in `./venv/bin/python` or `./venv/scripts/python` in Windows. I find this a bit awkward though, and you can end up with very long lines if you're calling pytest or sphinx with a whole load of arguments, for example.

PR #565, opened 2024-09-18
ci: fix dependabot glob pattern

jorgepiloto:
The issue with the current glob pattern for `dependabot` is that these patterns can not be in conflict between different ecosystems. That is, if two ecosystems are checking for the same file, dependabot fails, see https://github.com/ansys/actions/network/updates/20639306/jobs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,7 +13,7 @@ updates:
 
   - package-ecosystem: "github-actions"
     directories:
-      - "![requirements]**/*"
+      - "!(requirements)/**/*"
     schedule:
       interval: "daily"
     commit-message:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,7 +13,7 @@ updates:
 
   - package-ecosystem: "github-actions"
     directories:
-      - "**/*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -13,7 +13,7 @@ updates:
 
   - package-ecosystem: "github-actions"
     directories:
-      - "**/*"
+      - "![requirements]**/*"
```
jorgepiloto, line 16:
```suggestion
      - "!(requirements)/**/*"
```

.github/dependabot.yml
```diff
@@ -13,7 +13,7 @@ updates:
 
   - package-ecosystem: "github-actions"
     directories:
-      - "**/*"
+      - "!(requirements)/**/*"
```
jorgepiloto, line 16: I am not even sure if the `!` negate is supported by the glob implemented by GitHub. We can merge and manually trigger dependabot to see how it behaves. Right now, it is broken with the status in main.

RobPasMue: Sounds good!

RobPasMue (approved):

PR #567, opened 2024-09-19
chore: use trusted publisher in ci_cd

klmcadams:
Use a trusted publisher instead of twine for releases and add the bot user & email secrets

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -138,6 +138,10 @@ jobs:
     name: "Tests"
     runs-on: ubuntu-latest
     needs: code-style
+    environment: release
+    permissions:
+      id-token: write
+      contents: write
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -174,8 +178,7 @@ jobs:
       uses: ansys/actions/release-pypi-test@main
       with:
         library-name: ${{ env.test-library-name }}
-        twine-username: "__token__"
-        twine-token: ${{ secrets.PYANSYS_PYPI_TEST_PAT }}
+        use-trusted-publisher: true
```
jorgepiloto, line 181: I just added trusted publishing support also for the PyPI project. Could we update those lines too?

jorgepiloto (approved):

PR #569, opened 2024-09-20
fix: skip release to PyPI test if version exists

jorgepiloto:
Fixes the issues raised in https://github.com/ansys/actions/actions/runs/10960208443, where the version being pushed was already present in the Test PyPI.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -178,8 +178,17 @@ jobs:
       uses: ansys/actions/release-pypi-test@main
       with:
         library-name: ${{ env.test-library-name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -179,6 +179,7 @@ jobs:
       with:
         library-name: ${{ env.test-library-name }}
         use-trusted-publisher: true
+        skip-existing: true
```
RobPasMue, line 182: The default was ``true`` already, right?

jorgepiloto: Good catch! Just enabled this input parameter.

PR #570, opened 2024-09-23
ci: fix test to pypi

SMoraisAnsys:
WIP to fix the current issue of sharing the same environment for testing pypi and releasing to the official pypi.

PR #572, opened 2024-09-24
ci: remove environment from test PyPI

SMoraisAnsys:
Follow up of #570 which fixed the environment used for test PyPI.
The problem with the fix is that we still end up with messages referring to the deployment in the PR discussion.

image
 
This PR consists in removing the environment used in test PyPI. Note that:
- the current environment used (`tests`) is an empty shell;
- the previous one (`release`) shouldn't have been used for testing if we want to keep the protection rules.


jorgepiloto (approved):
Tanks a lot, @SMoraisAnsys

PR #582, opened 2024-09-25
feat: remove support for doc deploy index

Revathyvenugopal162:
BREAKING CHANGE: depreciate meilisearch and use the latest stable search

WIP: should be merge after reelase of `ansys sphinx theme 1.1`


jorgepiloto (approved):
Yeah!

Revathyvenugopal162:
> I think we can merge this in main, but we need to make sure that before releasing `v8` we have released `ansys-sphinx-theme==1.1`, is that right, @Revathyvenugopal162?

yes that is right.

requirements/requirements_doc.txt
```diff
@@ -1,6 +1,5 @@
 ansys-sphinx-theme==1.0.11
 jupyter==1.1.1
-nbformat==5.10.4
```
jorgepiloto, line 3: This is no longer required?

jorgepiloto (approved):

PR #590, opened 2024-10-01
fix: `index.rst` and migration guide title

Revathyvenugopal162:


doc/source/index.rst
```diff
@@ -1,3 +1,7 @@
+
+Ansys actions
```
jorgepiloto, line 2: This is already included from the `README.rst`, right?

jorgepiloto: Oh! I see it is no longer present. Thanks for adding this.

Revathyvenugopal162: https://github.com/ansys/actions/pull/588 with this PR, the directive starts after the title, so there is no title for the action in the stable docs
image

jorgepiloto, line 2:
```suggestion
Ansys actions
```

jorgepiloto (approved):
Thanks! Merging and cherry-picking.

PR #593, opened 2024-10-02
feat: verify docs have been generated properly on Windows

### LINKED ISSUE #583: Enforce `doc-build` with a check of the success in building HTML files
I want to use the `doc-build` action with `pyaedt-examples`. Our workflow is such that running all examples takes ~3-4 hours. However, from time to time, we have errors appearing in an example as we make the API evolve in `pyaedt`. In such cases building the HTML files stops after an arbitrary amount of time, e.g. 2 hours. However, this doesn't stop the action and it then tries to run the examples to create the PDF files. This is a problem because they will fail in the same way as the previous step, e.g. after 2 hours.

Could we had a simple check after the build of HTML files to ensure it succeeded ?

I gave it a try in the past and here was the workaround for windows. If someone has a better approach, that would be great. Note that this should be taking into account the fact that one uses `nbsphinx_allow_errors = False` in the doc settings.

```
      # NOTE: Ugly HTML checking but redirecting errors with our docs either leads
      # to errors in powershell or workflows getting stuck in Github CICD.
      - name: Check HTML doc creation
        run : |
          if (Test-Path "doc\_build\html\index.html" ) {
            Write-Output "File index.html found, HTML build successful."
          } else {
            Write-Error "Sphinx build failed, check if there is no cell execution error."
            exit 1            
          }
```

RobPasMue:
As title says. Closes #583

### DIFF
--- a/_doc-build-windows/action.yml
+++ b/_doc-build-windows/action.yml
@@ -431,6 +431,24 @@ runs:
 
     # ------------------------------------------------------------------------
 
+    - name: Verify the docs have been generated properly
+      shell: powershell
+      run: |
+        # Check if the HTML files have been generated
+        if (Test-Path -Path "${{ env.EXPECTED_BUILD_DIR }}\html\index.html") {
+          Write-Host "Docs have been generated successfully"
+        } else {
+          Write-Error "Docs have not been generated properly"
+        }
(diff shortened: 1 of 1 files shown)


MaxJPRey (approved):
LGTM.

PR #596, opened 2024-10-03
fix: solving issues in ``actions/tests-pytest`` action

RobPasMue:
Standarizing virtual environment creation based on OS and decoupling problems from observed behavior on Windows GitHub runners

See run: https://github.com/ansys-internal/allie-clients-python/actions/runs/11156816834

### DIFF
--- a/tests-pytest/action.yml
+++ b/tests-pytest/action.yml
@@ -176,6 +176,10 @@ runs:
       shell: bash
       run: |
         python -m venv .venv
+
+    - name: "Set up virtual environment activation command"
+      shell: bash
+      run: |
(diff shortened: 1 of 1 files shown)


jorgepiloto (approved):
Thanks for the hot-fix, @RobPasMue

SMoraisAnsys (approved):
LGTM.
I'm wondering if we shouldn't split the tests action like the doc build since some VMs might not be able to use `shell: bash`. While this is not required for other actions where the OS used is not relevant to the purpose of the action, I think that's not the case for documentation build and testing.

PR #599, opened 2024-10-04
feat: Add an input flag to configure diff output in code style checks

viseshrp:
Hello!
Most of the time, [our use case](https://github.com/ansys/pydynamicreporting/blob/main/.github/workflows/ci_cd.yml#L32) does not require the --show-diff-on-failure flag which when turned on, it produces a large diff of unwanted output that pollutes the action logs. Thereby, I'd like to propose a flag to turn it off but keep it on by default.

### DIFF
--- a/code-style/action.yml
+++ b/code-style/action.yml
@@ -89,6 +89,13 @@ inputs:
     default: true
     type: boolean
 
+  show-diff-on-failure:
+    description: |
+      Whether to show the diff when a pre-commit hook fails.
+    required: false
+    default: true
+    type: boolean
+
 runs:
   using: "composite"
   steps:
@@ -164,7 +171,7 @@ runs:
         python -m venv .venv
       env:
         PIPX_BIN_DIR: ${{ runner.temp }}/pipx/bin
-        PIPX_HOME : ${{ runner.temp }}/pipx/home
+        PIPX_HOME: ${{ runner.temp }}/pipx/home
 
     - name: "Create a virtual environment"
       if: env.BUILD_BACKEND == 'pip'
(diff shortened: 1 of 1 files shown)


RobPasMue (approved):
Sounds like a great idea! Thanks @viseshrp !

viseshrp:
> Sounds like a great idea! Thanks @viseshrp !

Thank you! How often do you guys make releases of ansys/actions? Thanks.

RobPasMue:
> > Sounds like a great idea! Thanks @viseshrp !
> 
> Thank you! How often do you guys make releases of ansys/actions? Thanks.

In this case, as soon as it gets merged (and I fix another coming up PR) I will do a patch release :) v8 will be updated today

RobPasMue:
Waiting on https://github.com/ansys/actions/pull/600 to release

RobPasMue:
New version released! If you consume ``v8`` you should get your changes @viseshrp

viseshrp:
> New version released! If you consume `v8` you should get your changes @viseshrp

Awesome. Thank you

viseshrp:
@RobPasMue Sorry for the late response, but I cannot get this to work:

image

I still see the diff on failure.
I am not sure what other changes are required or if I'm doing this the wrong way. Any thoughts?

RobPasMue:
Can you share a CI/CD log where this is happening so that I can pull the thread and investigate? =)

viseshrp:
@RobPasMue https://github.com/ansys/pydynamicreporting/actions/runs/11335692082/job/31524324550
I think this may be my misunderstanding and it may actually work.
The check here maybe incorrect even though it works. Should the input value mentioned here be in quotes to make sure the comparison is correct? It's comparing boolean against a string, if I'm right?

image

viseshrp:
If that is indeed the problem, one of your existing steps do the same too, so I guess it isn't?

image

viseshrp:
@RobPasMue I am sorry, I think bash somehow casts the bool to a string. Everything looks and works fine. Please ignore my comments.

PR #604, opened 2024-10-08
fix: standardize venv activation in `actions/code-style`

ludovicsteinbach:
Similarly to the issues resolved by https://github.com/ansys/actions/pull/596, the v8 `code-style` action fails on windows runners. See https://github.com/ansys/openapi-common/actions/runs/11208266030/job/31226892848 for example.

This PR applies fixes similar to https://github.com/ansys/actions/pull/596, however that still was failing https://github.com/ansys/openapi-common/actions/runs/11234704576/job/31231061286, in a way that suggests that the virtual envrionment wasn't activated at all.

I've had to change the command from `.venv\\Scripts\\activate` to `source .venv/scripts/activate` for it to actually activate the env and get a successful run https://github.com/ansys/openapi-common/actions/runs/11234704576

### DIFF
--- a/check-vulnerabilities/action.yml
+++ b/check-vulnerabilities/action.yml
@@ -262,24 +262,33 @@ runs:
       run: |
         python -m venv .venv
 
+    - name: "Set up virtual environment activation command"
+      shell: bash
+      run: |
+        if [[ ${{ runner.os }} == 'Windows' ]]; then
+          echo "ACTIVATE_VENV=$(echo 'source .venv/scripts/activate')" >> $GITHUB_ENV
+        else
+          echo "ACTIVATE_VENV=$(echo 'source .venv/bin/activate')" >> $GITHUB_ENV
+        fi
+
     # ------------------------------------------------------------------------
 
     - name: "Update pip"
       shell: bash
       run: |
-        source .venv/bin/activate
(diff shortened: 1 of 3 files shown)


RobPasMue (approved):
LGTM! Looks like this might have to be applied to the other actions as well whenever we use a bash shell.

ludovicsteinbach:
I can't comment on what is meant to be supported, but we've been using the code-style action on windows for a while, and I can't see anything in the docs that suggests it should be limited to Linux.
Happy to have a look at the check-vulnerabilities and apply similar changes

RobPasMue:
Actions should be as OS-compatible as possible. @ludovicsteinbach could you also adapt the ``check-vulnerabilities`` logic as well? Thanks in advance!

RobPasMue:
Thanks @ludovicsteinbach! I also realized ``tests-pytest`` would have to be adapted. Everything should work now. Approving!

RobPasMue (approved):

SMoraisAnsys (approved):
LGTM
Since the shell was `bash` I always though it was implicitly a way to state that it was dedicated to Linux. Nvm a glad to know that it's not the case :)

RobPasMue:
This change has been released in ``v8``.

PR #606, opened 2024-10-10
fix: exit 1 when logging error messages

### LINKED ISSUE #603: Username and email inputs not throwing error
Although the error gets reported in the log

image

The action is not failing... I'd recommend throwing out an error. Assigning to @klmcadams. This is a bug.

klmcadams:
Add check to see if bot user or email inputs are empty. If so, log an error message & exit 1. Exiting 1 when logging an error message was applied to the _logging action.

[Exit 1 on error](https://github.com/ansys-internal/actions-sandbox/actions/runs/11278326057/job/31366557792)
[Bot secrets exist, so no error](https://github.com/ansys-internal/actions-sandbox/actions/runs/11278177040/job/31366367645)

Closes #603

### DIFF
--- a/_logging/action.yml
+++ b/_logging/action.yml
@@ -57,6 +57,7 @@ runs:
       if: ${{ inputs.level == 'ERROR' && runner.os == 'Linux' }}
       run: |
         echo -e "\033[1;91m[ERROR]: ${{ inputs.message }}\033[0m"
+        exit 1
 
     - name: "Report warning message"
       shell: bash
@@ -77,6 +78,7 @@ runs:
       if: ${{ inputs.level == 'ERROR' && runner.os == 'Windows' }}
       run: |
         Write-Host "$([char]27)[91m[ERROR]: ${{ inputs.message }}$([char]27)[0m"
+        Exit 1
 
     - name: "Report warning message"
       shell: powershell
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -123,7 +123,7 @@ runs:
 
     # TODO: remove this deprecation in ansys/actions@v9
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@fix/exit-error-logging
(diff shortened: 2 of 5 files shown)


germa89 (approved):
I think doing `exit 1` when we are loggin an error makes sense... however have you checked that you are not logging anywhere else with ``ERROR`` level? 
Alternatively, you could do that exit optionally using another argument (``force_exit`` maybe?)

Other than that, it looks good to me.

klmcadams:
It looks like the other locations logging with ERROR are [_doc-build-windows](https://github.com/ansys/actions/blob/9e5cc68dd3d411d8318ccc51baae1468cab8d964/_doc-build-windows/action.yml#L368) and [_doc-build-linux](https://github.com/ansys/actions/blob/9e5cc68dd3d411d8318ccc51baae1468cab8d964/_doc-build-linux/action.yml#L362) when it's unable to determine the PDF filename using the conf.py file. If being unable to determine the PDF filename isn't a dealbreaker & the action can still pass without it, I think the logging level for those cases should change to WARNING instead

SMoraisAnsys (approved):
LGTM For the `doc-build` actions I think it's great to have this early exit because if the error logs are triggered, it means that the job using the action will fail fore sure.

SMoraisAnsys:
> It looks like the other locations logging with ERROR are [_doc-build-windows](https://github.com/ansys/actions/blob/9e5cc68dd3d411d8318ccc51baae1468cab8d964/_doc-build-windows/action.yml#L368) and [_doc-build-linux](https://github.com/ansys/actions/blob/9e5cc68dd3d411d8318ccc51baae1468cab8d964/_doc-build-linux/action.yml#L362) when it's unable to determine the PDF filename using the conf.py file. If being unable to determine the PDF filename isn't a dealbreaker & the action can still pass without it, I think the logging level for those cases should change to WARNING instead

It would break further steps in the action if one triggers any of the error log.

RobPasMue (approved):
LGTM!

RobPasMue:
Releasing this change

PR #614, opened 2024-10-16
fix: install additional deps on doc-style action for rST support

RobPasMue:
As title says. This was modified in https://github.com/errata-ai/vale-action/commit/1262efadd6fde1040255db25589fe57804d785d2 inside the vale-action so now we need to install it manually.

### DIFF
--- a/doc-style/action.yml
+++ b/doc-style/action.yml
@@ -134,6 +134,11 @@ runs:
       run: |
         python -m pip install --upgrade pip toml==${{ inputs.toml-version }}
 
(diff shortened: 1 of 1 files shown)


RobPasMue:
Validated here: https://github.com/ansys/pyansys-geometry/pull/1483

RobPasMue:
Merging and releasing since this is blocking for the entire ecosystem

RobPasMue:
Failing status is due to the fact that the PR uses the version in main

PR #615, opened 2024-10-16
fix: unreachable link

RobPasMue:
As title says... the doc build has started to fail. My feeling is that they must have implemented some protection against bots.


SMoraisAnsys (approved):
LGTM
@RobPasMue Did you tried defining an agent-user ? I know this might lead to the same issues we are having with corporate pages - just wanting to know if that could help.

RobPasMue:
> LGTM @RobPasMue Did you tried defining an agent-user ? I know this might lead to the same issues we are having with corporate pages - just wanting to know if that could help.

I did not but I can give it a try - one sec

RobPasMue:
We are brutally ignored even if we pass in the user-agent 😄 

https://github.com/ansys/actions/actions/runs/11360456614/job/31598472212#step:2:3868

Reverting

PR #618, opened 2024-10-22
feat: add install target to test-pytest

### LINKED ISSUE #617: Add install target input for tests-pytest

I would like to be able to specify the install target to use when running the tests with action `tests-pytest`. Currently, the action expects a `tests` target and installs the associated dependencies. However, if I want to tests another set of dependencies, I'm not able to do it with the current action.

Adding the install target would allow one to test different backends or to test with fewer dependencies (decreasing CI duration). An example for the second benefit would be to have an `unittests` install target composed of only `pytest` and `mock` instead of having to install `matplotlib`, `numpy`, `pyvista`, ...

Add a new input and make it default to `tests` to keep previous behavior.

_No response_

SMoraisAnsys:
As title says. Allow users to provide an install target (pip) or dependency group (poetry) to install before running the tests.

Close #617

---

**Edit** As mentioned in #617, the scope which was previously mentioned as been extended to include requirements file.


RobPasMue (approved):
LGTM!

tests-pytest/action.yml
```diff
@@ -228,17 +240,17 @@ runs:
...
+        python -m pip install -r requirements/requirements_${{ inputs.optional-dependencies-name }}.txt
 
     - name: "Install test dependencies from pyproject.toml"
       shell: bash
       if: env.EXISTS_TESTS_REQUIREMENTS == 'false'
       run: |
         ${{ env.ACTIVATE_VENV }}
         if [[ ${{ env.BUILD_BACKEND }} == 'poetry' ]]; then
-          poetry install --with tests
+          poetry install --with ${{ inputs.optional-dependencies-name }}
```
vgelbgras, line 251: Maybe it could be nice to add --all-extras. It does not fail if no extras are defined and if some extras are defined, it might be required to run the tests

RobPasMue: This would be a poetry-only feature. As such, we should document it. I'm fine with enabling this argument @SMoraisAnsys - final decision up to you since you are the author of the PR. However, if you open this door, I would do the same on the docs-build action

SMoraisAnsys: I just checked and it seems that `all-extras` installs all dependencies defined in your pyproject.toml file.

Could you describe why would you want this behavior ? Wouldn't having a specific group dependency make more sense as it would allow you to install only what you require for testing and nothing else ? Otherwise you'll probably end up installing dependencies for dev, documentation, ... which could increase the CICD duration.

vgelbgras: --all-extras will install all extras, not all dependencies group.

extras can be defined so the users of the packages can pip install a package with additional packages, e.g.: : pip install package[pyansys]

SMoraisAnsys: I missed that subtilty, thanks for pointing it out !

The current approach we are using for multiple actions is to leverage dependencies group when a project uses `poetry`. After reading a bit more about the difference between `extra` and `dependency group` I'd stick to `dependency group`. From my understanding, `dependency group` separates specific environments for dev, test, docs, ... while extras seem to be more suited for selecting optional functionalities of the project. As we are running that in CICD, I think it makes more sense to keep the current approach.

As for the `--all-extras` flag, my previous point is still valid as it leads to installing all extras right ?

Revathyvenugopal162 (approved):

RobPasMue (approved):

MaxJPRey (approved):

tests-pytest/action.yml
SMoraisAnsys: @vgelbgras I'll close this conversation for the purpose of merging this PR.
If you have additional information to provide or would like to raise a discussion over the `extra` and `--all-extras` subject, please open a discussion and ping us :)

PR #621, opened 2024-10-24
fix: doc build without python cache

SMoraisAnsys:
Follow up of #543 which led to other PR with conflicting changes being integrated in the main branch (see #650 for example)

This PR mainly does: 
- ensure that `use-python-cache` is propagated to private doc build actions;
- add `--no-cache-dir` flag for `pip` build backend and `--no-cache` flag for poetry backend if the input asks for no python cache;

> Note: a commit starting with `tbr` is here for testing purpose, it will be reverted afterward

---

Tested with `pip` through this project and it work, see [action run](https://github.com/ansys/actions/actions/runs/11500989589)

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -45,16 +45,6 @@ jobs:
           use-python-cache: true
           needs-quarto: true
 
-      - name: "Build documentation with cache"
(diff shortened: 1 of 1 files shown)


jorgepiloto:
Delaying this to the next major version.

SMoraisAnsys:
Closing since the PR diverged that much. If we ever want to rework on that topic, this could be used as future ref.

PR #623, opened 2024-10-24
fix: check step in _doc-build-windows

### LINKED ISSUE #622: Bug located in doc-build (windows)

### 🐞 Description of the bug

When one uses the latest version of `ansys/actions/doc-build` in windows, there is an additional step that was introduced in #593. However the implementation leverage an environment variable that does not exist if one doesn't activate the upload assets input.

I'll open the floor for others things that could lead to other issues / devs:

1. Are we sure that the current workflows avoids pdf build from being performed ? AFAIK, the changes of #593 do not prevent to build the pdf (and consume CICD time), they only check before uploading. If we agree with that, I'll propose some changes in the action.
2. Couldn't this checking step be added to _doc-build-linux ?

### 📝 Steps to reproduce

See the logs of https://github.com/ansys/pyansys-sound/actions/runs/11289857973/job/31666254238

Windows

out of topic

3.12

### 📦 Installed packages

```shell
out of topic
```

SMoraisAnsys:
Ensures that `EXPECTED_BUILD_DIR` is defined even if `add-pdf-html-docs-as-assets` is `false`.
Only trigger checking if the expected doc build directory is used, aka `doc/_build`.

Close #622

### DIFF
--- a/_doc-build-windows/action.yml
+++ b/_doc-build-windows/action.yml
@@ -379,20 +379,17 @@ runs:
     # ------------------------------------------------------------------------
 
     - uses: ansys/actions/_logging@main
-      if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       with:
         level: "INFO"
         message: >
           Check if expected directories exist.
 
     - name: Set expected build directory
-      if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       shell: powershell
       run: |
         echo "EXPECTED_BUILD_DIR=$(echo 'doc\_build')" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
     - name: Check expected build directory
-      if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       shell: powershell
       run: |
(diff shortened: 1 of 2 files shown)


_doc-build-windows/action.yml
```diff
@@ -379,20 +379,17 @@ runs:
     # ------------------------------------------------------------------------
 
     - uses: ansys/actions/_logging@main
-      if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
```
MaxJPRey, line 382: Could you remind me why did we put this condition check in the first place?

SMoraisAnsys: Previously we didn't have `EXPECTED_BUILD_DIR` or `EXISTS_EXPECTED_BUILD_DIR` environment variables.
They were introduced when the feature to add assets was added through `add-pdf-html-docs-as-assets`.

Since some projects can provide a different build directory, the environment variables scope of work was associated to the asset feature which requires a standard location (or would have to ask the user for the location). Since we try to share common design, this was a lever for adoption :)

a-bouth (approved):
Looks good, thanks a lot !!

_doc-build-windows/action.yml
MaxJPRey: OK, now I remember it. Thanks

PR #624, opened 2024-10-24
feat: use the CHANGELOG file in the `release-github` action

### LINKED ISSUE #460: Allow release-github action to use the changelog output
We should try to combine both actions: changelog build and release-github. That way when we release, we can include the changelog of the current version inside the release notes on GitHub. ``softprops/action-gh-release`` allows for it so it should be feasible

klmcadams:
Use the CHANGELOG file in the `release-github` action
- Search pyproject.toml file for towncrier filename (the path to the changelog file)
- If the filename does not exist, then release to GitHub without using the changelog file
- If the changelog file exists, release to GitHub using the changelog file

Note: this will be a breaking change due to the new required "token" input

Closes #460


release-github/action.yml
```diff
@@ -173,12 +190,56 @@ runs:
...
+    - name: "Release to GitHub"
+      uses: softprops/action-gh-release@v1
+      env:
+        REPO_NAME: ${{ github.event.repository.name }}
+      permissions:
+        contents: write
+      if: inputs.dry-run == 'false' && env.CHANGELOG_LOC != ""
+      with:
+        fail_on_unmatched_files: false
+        generate_release_notes: true
```
klmcadams, line 218: Should `generate_release_notes` be false when using the changelog file? This is the description of the input:

Whether to automatically generate the name and body for this release. If name is specified, the specified name will be used; otherwise, a name will be automatically generated. If body is specified, the body will be pre-pended to the automatically generated notes. See the [GitHub docs for this feature](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes) for more information

I think it should be true for the sake of the name being automatically generated?

release-github/action.yml
```diff
@@ -173,12 +204,54 @@ runs:
...
+    - name: "Install toml"
+      shell: bash
+      run: |
+        python -m pip install --upgrade pip toml==${{ inputs.toml-version }}
+
+    - name: "Get CHANGELOG file location"
+      shell: python
+      run: |
+        import sys
+        sys.path.insert(1, '${{ github.action_path }}/../doc-changelog/')
```
jorgepiloto, line 216: This is a very interesting approach. I know that, any file inside a `composite-action/` gets copied at CI/CD runtime. Maybe, this is one step for solving #459. We could have a composite actions that just contain Python files utilities.

release-github/action.yml
SebastienMorais: Agreeing

release-github/action.yml
```diff
@@ -173,12 +204,54 @@ runs:
...
+
+        changelog_loc = get_towncrier_config_value("filename")
+        save_env_variable("CHANGELOG_LOC", changelog_loc)
+
+    - name: "Release to GitHub"
+      uses: softprops/action-gh-release@v1
+      env:
+        REPO_NAME: ${{ github.event.repository.name }}
+      if: inputs.dry-run == 'false' && env.CHANGELOG_LOC != ''
+      with:
```
RobPasMue, line 228: Can we avoid duplication of the stage using conditional inputs for the arguments and assigning defaults whenever we don't use them?

klmcadams:
After testing, this is what the release notes look like:

[Release notes with changelog.rst](https://github.com/ansys-internal/actions-sandbox/releases)
[Release notes with CHANGELOG.md](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.10.1)

The RST changelog doesn't render well since the release notes are in markdown. Also, the entire changelog is copied into the release notes, not just the section for that release.

Some ideas:
- Translate RST changelog files to MD in the action
- Use regex (?) to parse the CHANGELOG file & retrieve only the section of the changelog for the corresponding tag
    - Potentially put the tagged section in a temporary file and use that file in the softprops/action-gh-release

Thoughts/suggestions? There could be more efficient ways to do this that I'm not thinking of

RobPasMue:
Yeah... the rST file version looks horrible. That one would need some conversion to MD for sure.

> Some ideas:
> 
> * Translate RST changelog files to MD in the action
> * Use regex (?) to parse the CHANGELOG file & retrieve only the section of the changelog for the corresponding tag
>   * Potentially put the tagged section in a temporary file and use that file in the softprops/action-gh-release

The ideas seem fine to me. Another option that occurred to me is to check the git history of the changelog and extract the last commit to it. If things are done how they should, this section will contain the release notes for the current release. If they are in RST we can convert them, to MD. If it's already in MD you just need to embed it.

klmcadams:
Test results:

[Only generated notes are there when changelog-release-notes is false](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.11.17)

CHANGELOG.md tests
- [More than one section](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.11.13)
- [One section](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.11.14)
- [No sections](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.11.15)
	
Test changelog.rst
- [More than one section](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.11.5)
- [One section](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.11.8)
- [No sections](https://github.com/ansys-internal/actions-sandbox/releases/tag/v0.11.9)

germa89 (approved):
LGTM, thank you for this!

python-utils/parse_pr_title.py
```diff
@@ -18,7 +19,12 @@ def save_env_variable(env_var_name: str, env_var_value: str):
 
     # Save environment variable with its value
     with open(github_env, "a") as file:
-        file.write(f"{env_var_name}={env_var_value}")
+        if "\n" in env_var_value or "\r\n" in env_var_value:
```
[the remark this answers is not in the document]
klmcadams: Hmm that's true. Should I also check for the `\r` newline value (Mac) too? If not, I'll change it to only check for `\n`

release-github/action.yml
```diff
@@ -188,12 +236,42 @@ runs:
       shell: bash
       run: ls -R dist/
 
+    - name: "Install toml"
```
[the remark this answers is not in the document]
klmcadams: Adding that would be very helpful! Thanks for pointing it out

python-utils/parse_pr_title.py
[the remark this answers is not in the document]
SMoraisAnsys: I think that `\r` was the old way to state a newline in Mac and that nowadays `\n` should be the value we encounter. However, better be safe than sorry. Please add the check on `\r`

MaxJPRey (approved):
Great work Kerry

python-utils/release_github_utils.py
```diff
@@ -0,0 +1,216 @@
+import re
+from pathlib import Path
+
+from parse_pr_title import get_towncrier_config_value, save_env_variable
+
+"""Semantic version regex as found on semver.org:
```
MaxJPRey, line 6: Thanks for adding the reference that explains the REGEX for futur maintenance.

RobPasMue (commented):
LGTM! Thanks @klmcadams

PR #627, opened 2024-10-25
ci: fix rolling release

SMoraisAnsys:
Following the failure of release `v8.1.0`, this PR should fix the problem of performing the rolling release workflow when one didn't create the minor/major release associated. In the case of `v8.1.0`, `v8` was already existing thanks to previous workflows but that was not the case for `v8.1`. The same problem can happen with major release.

> Note: There are two commits here and they propose two approaches to handle that problem. The latest comes from @jacobevansgit and is performed in `ansys-internal` actions. The idea is to add `|| true` to the tag deletion command which ensures that the workflow won't stop even if the tag does not exist. The other approach checks tags and either perform deletion or logs an info, depending on tag existence.

Let me know which approach you prefer

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
     - name: "Checkout repository"
       uses: actions/checkout@v4
+      with:
+        fetch-tags: true
 
     - name: "Configure git"
       run: |
@@ -44,8 +46,16 @@ jobs:
 
     - name: "Remove old tags"
       run: |
+        if git tag -l | grep -q "^${{ env.X }}$"; then 
           git push --delete origin ${{ env.X }}
+        else
+          echo -e "\033[1;92m[INFO]: Tag ${{ env.X }} does not exist, skipping deletion.\033[0m"
+        fi
+        if git tag -l | grep -q "^${{ env.X }}$"; then 
           git push --delete origin ${{ env.X }}.${{ env.Y }}
+        else
+          echo -e "\033[1;92m[INFO]: Tag ${{ env.X }}.${{ env.Y }} does not exist, skipping deletion.\033[0m"
+        fi
 
     - name: "Create new tags"
       run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
     - name: "Checkout repository"
       uses: actions/checkout@v4
+      with:
+        fetch-tags: true
 
     - name: "Configure git"
       run: |
@@ -44,8 +46,16 @@ jobs:
 
     - name: "Remove old tags"
       run: |
-        git push --delete origin ${{ env.X }} || true
-        git push --delete origin ${{ env.X }}.${{ env.Y }} || true
+        if git tag -l | grep -q "^${{ env.X }}$"; then 
+          git push --delete origin ${{ env.X }}
+        else
+          echo -e "\033[1;92m[INFO]: Tag ${{ env.X }} does not exist, skipping deletion.\033[0m"
+        fi
+        if git tag -l | grep -q "^${{ env.X }}$"; then 
+          git push --delete origin ${{ env.X }}.${{ env.Y }}
+        else
+          echo -e "\033[1;92m[INFO]: Tag ${{ env.X }}.${{ env.Y }} does not exist, skipping deletion.\033[0m"
+        fi
 
     - name: "Create new tags"
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -44,8 +44,8 @@ jobs:
 
     - name: "Remove old tags"
       run: |
-          git push --delete origin ${{ env.X }}
-          git push --delete origin ${{ env.X }}.${{ env.Y }}
+        git push --delete origin ${{ env.X }} || true
+        git push --delete origin ${{ env.X }}.${{ env.Y }} || true
```
RobPasMue, line 49: This could mask other issues IMO. I would do a more active checking - verify if the tag exists or not, and if it exists, delete it.

For example, if the action is enabled in a repo with some specific tag protection... we could be hitting an error that gets masked because the `` || true`` section swallows anything. For the sake of our end users, I would do the active checking

RobPasMue (changes requested):

.github/workflows/release.yml
```diff
@@ -44,8 +46,16 @@ jobs:
...
+        if git tag -l | grep -q "^${{ env.X }}$"; then 
           git push --delete origin ${{ env.X }}
+        else
+          echo -e "\033[1;92m[INFO]: Tag ${{ env.X }} does not exist, skipping deletion.\033[0m"
+        fi
+        if git tag -l | grep -q "^${{ env.X }}$"; then 
           git push --delete origin ${{ env.X }}.${{ env.Y }}
+        else
+          echo -e "\033[1;92m[INFO]: Tag ${{ env.X }}.${{ env.Y }} does not exist, skipping deletion.\033[0m"
+        fi
```
RobPasMue: This one I mean... Much safer IMO.

.github/workflows/release.yml
jacobevansgit: > protected tags

☝ 
I'm gonna change my implementation on the if chance

.github/workflows/release.yml
```diff
@@ -44,8 +46,16 @@ jobs:
 
     - name: "Remove old tags"
       run: |
+        if git tag -l | grep -q "^${{ env.X }}$"; then 
           git push --delete origin ${{ env.X }}
+        else
+          echo -e "\033[1;92m[INFO]: Tag ${{ env.X }} does not exist, skipping deletion.\033[0m"
+        fi
+        if git tag -l | grep -q "^${{ env.X }}$"; then
```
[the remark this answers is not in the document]
RobPasMue: Correct, good catch @jacobevansgit!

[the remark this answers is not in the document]
SMoraisAnsys: @jacobevansgit Thanks for the feedback, I also missed the `\` before the `.` making it a total different check :D
Glad you had a look !

RobPasMue (approved):
LGTM!

RobPasMue (approved):

jorgepiloto (approved):
Thanks for adding this, @SMoraisAnsys

MaxJPRey (approved):
LGTM.
Should help many of us in the futur. 😁

.github/workflows/release.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     steps:
     - name: "Checkout repository"
       uses: actions/checkout@v4
+      with:
+        fetch-tags: true
```
SMoraisAnsys, line 18:
```suggestion
      with:
        fetch-tags: true
        ref: ${{ github.ref }}
```

SMoraisAnsys:
Didn't had the time to test this on a dummy repo and when doing so I found an issue with how `action/checkout` works.
Below was the error I faced in my sandbox

image

Seems like there is a problem when working with a commit associated to a tag and using input `fetch-tags: true`. The solution I found to that is to force the checkout on the tag itself through input `ref: ${{ github.ref }}`. This avoids the conflict as we are now working with `+refs/tags/v3.0.0:refs/tags/v3.0.0` instead of `2c8f0122cd9d29813c194ae3713e1ee09eef2676:refs/tags/v3.0.0`.

Final result seem to be working as expected. Workflow https://github.com/SMoraisAnsys/hadolint-gh-action/actions/runs/11609641550
image

RobPasMue:
Aha... good catch @SMoraisAnsys. This is very good to know!

RobPasMue:
Are we merging this?

SMoraisAnsys:
Sure, just wanted to let some time for everyone to notice the latest changes. I'm merging it right now :)

Issue #587, opened 2024-09-30
Migrate README to markdown and use GitHub mermaid


jorgepiloto:
Right now, the `README` file is an `rST` one. If using `Markdown`, it would be possible to use [mermaid] and have dynamic workflows. Main drawback is that we can not reuse the file within our docs but so far, we are only reusing a couple of lines.

```mermaid
graph LR
    A[Doc style] --> B[Doc build]
    B --> C[Build library]
    
    D[Code style] --> E[Tests]
    E --> C[Build library]
```

MaxJPRey:
Will look really nice for sure.

PR #644, opened 2024-11-19
fix: demo library README is still rst

RobPasMue:
Follow up to #587


RobPasMue:
CI/CD was failing - demo library is still in rst format, we shouldn't have changed it there.

PR #648, opened 2024-11-21
feat: skip label step if coming from a forked repository

### LINKED ISSUE #645: Ignore labeler in case an outside collaborator is submitting a request
Following #643 we should consider ignoring write operations on GitHub in case a user does not have write access to a repo. This happens whenever a user submits a PR coming from a forked repo, for example. We could simply skip it under those circumstances. This should be backported to all repos since we all (or most of us) suffer from it. We should present it to the larger dev team.

RobPasMue:
Closes #645 

This will skip the label insertion if the change is coming from a forked repository... requires testing but should work.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -36,6 +36,8 @@ jobs:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
      - name: "Label pull-request"
+       # Skip if user does not have write permissions - PR coming from a fork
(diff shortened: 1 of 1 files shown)


RobPasMue:
@ansys/pyansys-core care to review? :)

Revathyvenugopal162 (approved):

AlejandroFernandezLuces (approved):

jorgepiloto (commented):
Thanks a lot for adding this, @RobPasMue

jorgepiloto (approved):

PR #646, opened 2024-11-20
feat: allow for extra targets in check-vulnerabilities action

RobPasMue:
Coming from an internal request by @greschd - we should allow the option to pass in extra targets to our check-vulnerabilities action.

PR #650, opened 2024-11-22
feat: change installation order of deps in check-vulnerabilities action

RobPasMue:
Following #646 - @greschd noticed that when using poetry and ``extra-targets`` there is a small problem... ``--with ...`` uninstalls any other deps found and thus causes setuptools not to be present, which is needed by our action to run.

By inverting the installation order, we should be able to fix this.

### DIFF
--- a/check-vulnerabilities/action.yml
+++ b/check-vulnerabilities/action.yml
@@ -294,12 +294,6 @@ runs:
         ${{ env.ACTIVATE_VENV }}
         python -m pip install -U pip
 
-    - name: "Install requirements"
-      shell: bash
-      run: |
-        ${{ env.ACTIVATE_VENV }}
-        pip install -r ${{ github.action_path }}/requirements.txt
-
     - name: "Install library"
       shell: bash
       # Note: we need to use an array to store the extra targets, otherwise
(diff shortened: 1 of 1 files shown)


greschd:
>  --with ... uninstalls any other deps found

Small correction: the dependencies in other _dependency groups_ (which the `--with` syntax refers to) are not uninstalled. It's the ones in _extras_ which are not explicitly installed. Or (probably, haven't tested) packages which aren't in the poetry dependencies at all.

But yeah, the fix makes sense.

greschd (approved):

RobPasMue:
> Small correction: the dependencies in other _dependency groups_ (which the `--with` syntax refers to) are not uninstalled. It's the ones in _extras_ which are not explicitly installed. Or (probably, haven't tested) packages which aren't in the poetry dependencies at all.

Thanks for writing down the clarification @greschd !

check-vulnerabilities/action.yml
```diff
@@ -325,6 +319,12 @@ runs:
           python -m pip install ."${extra_targets[@]}"
         fi
 
+    - name: "Install action requirements"
+      shell: bash
+      run: |
+        ${{ env.ACTIVATE_VENV }}
+        pip install -r ${{ github.action_path }}/requirements.txt
```
jorgepiloto, line 326: I know this was already, but it would be nice to add here:

```suggestion
        python -m pip install -r ${{ github.action_path }}/requirements.txt

```

jorgepiloto (approved):

RobPasMue:
I'll do a follow up patch release with this - thank you @greschd @jorgepiloto for reviewing and approving

PR #653, opened 2024-11-29
fix: allow for missing patch number in tag when deploying docs

RobPasMue:
Following discussion on dev meeting - tags missing patch number should also be accepted for deploying stable docs since it isn't necessary to have them


doc-deploy-stable/action.yml
```diff
@@ -191,10 +191,11 @@ runs:
         tag=${{ env.REF_NAME }}
         version=$( echo $tag | cut -d 'v' -f 2)
 
-        if [[ $version =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$ ]]; then
+        if [[ $version =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(\.(0|[1-9][0-9]*))?$ ]]; then
```
RobPasMue, line 194:
```suggestion
        if [[ $version =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(\..*)?$ ]]; then
```
Another option is to align this with the next regex... up to you @ansys/pyansys-core - I just wanted to remain as less intrusive as possible

jorgepiloto: I would stick to the first one just to avoid any unknown issues caused by suffix tags. Just wondering how the logic would behave.

check-vulnerabilities/ignored-safety.txt
```diff
@@ -1,4 +1,3 @@
 52495
 62044
 67599
-70612
```
jorgepiloto, line 4: Is this intended?

SMoraisAnsys:
> Following discussion on dev meeting - tags missing patch number should also be accepted for deploying stable docs since it isn't necessary to have them

That might be a dumb question but why are we allowing versions that don't follow X.Y.Z template ?
Are maintainers following that approach only releasing minor releases or do they expect to also release patch releases ? If it's the latter, I'm a bit confused as of how this will end up in pypi or github releases.

check-vulnerabilities/ignored-safety.txt
RobPasMue: It is - this Safety advisory ahs been removed

RobPasMue:
> > Following discussion on dev meeting - tags missing patch number should also be accepted for deploying stable docs since it isn't necessary to have them
> 
> That might be a dumb question but why are we allowing versions that don't follow X.Y.Z template ? Are maintainers following that approach only releasing minor releases or do they expect to also release patch releases ? If it's the latter, I'm a bit confused as of how this will end up in pypi or github releases.

According to our actions, we were allowing them previously. We can remove this behavior altogether as well, but there's nothing preventing a user from creating a tag "v0.8" and pushing it to PyPI without a patch number. Not ideal, but nothing preventing you from doing it either.

We can discuss this further on Thursday if you prefer.

doc-deploy-stable/action.yml
MaxJPRey: +1

SMoraisAnsys:
> > > Following discussion on dev meeting - tags missing patch number should also be accepted for deploying stable docs since it isn't necessary to have them
> > 
> > 
> > That might be a dumb question but why are we allowing versions that don't follow X.Y.Z template ? Are maintainers following that approach only releasing minor releases or do they expect to also release patch releases ? If it's the latter, I'm a bit confused as of how this will end up in pypi or github releases.
> 
> According to our actions, we were allowing them previously. We can remove this behavior altogether as well, but there's nothing preventing a user from creating a tag "v0.8" and pushing it to PyPI without a patch number. Not ideal, but nothing preventing you from doing it either.
> 
> We can discuss this further on Thursday if you prefer.

I don't think I'll be able to join on Thursday (I'm getting back from travel) but that's a subject that should be clarified yeah. I'd rather have no X.Y version and only X.Y.Z versions.

check-vulnerabilities/ignored-safety.txt
jorgepiloto: Nice!

jorgepiloto (approved):

SMoraisAnsys (changes requested):
Taking the liberty to "block" this PR as long as the dicussion on releasing minor versions is not clarified

RobPasMue:
Was this discussed in the end? I missed the meeting in that case...

SMoraisAnsys:
> Was this discussed in the end? I missed the meeting in that case...

No, I'm about to mention that on thursday meeting

RobPasMue:
> > Was this discussed in the end? I missed the meeting in that case...
> 
> No, I'm about to mention that on thursday meeting

 I'm skipping that one too (on holidays). Feel free to close the PR if you end up deciding it's not worth exposing it.

SMoraisAnsys:
Passing by (in day off) we should wait à bit before merging or not. Further discussions must be performed

RobPasMue:
> Passing by (in day off) we should wait à bit before merging or not. Further discussions must be performed

Sounds good - thanks for the update @SMoraisAnsys

RobPasMue:
What was the final result here? Should we abandon this PR?

SMoraisAnsys:
> What was the final result here? Should we abandon this PR?

For the moment no, please leave me a few days and I'll comment more

jorgepiloto:
Closing after a private discussion with the @ansys/pyansys-core team.

PR #652, opened 2024-11-27
feat: new action for automerging dependabot and pre-commit.ci PRs

RobPasMue:
WIP... needs testing.

PR #654, opened 2024-12-02
fix: use actions/cache@v4

jorgepiloto:
Use the latest version for the [actions/cache] action, see this notice of upcoming deprecation: https://github.blog/changelog/2024-09-16-notice-of-upcoming-deprecations-and-changes-in-github-actions-services/


RobPasMue (approved):
Ah good catch! Is this the only ref?

RobPasMue:
The failing section is due to a Safety advisory that has been removed... see other PRs like #653 or #652 for the fix

MaxJPRey (approved):

PR #666, opened 2025-01-07
refactor: replacing-check-with-scan

### LINKED ISSUE #641: Use scan command in favor of check one for bandit
Currently, [bandit] uses the [check] command, see:

https://github.com/ansys/actions/blob/126c7eb0d84802c87341694b7b22a66a02f1430d/check-vulnerabilities/action.yml#L315

However, I started to get this notification:

```console
DEPRECATED: this command (`check`) has been DEPRECATED, and will be unsupported beyond 01 June 2024.

We highly encourage switching to the new `scan` command which is easier to use, more powerful, and can be set up to mimic the deprecated command if required.
```

We should migrate to the new command.

germa89:
As the title, using `scan` instead of check.

Close #641 

### Notes

I have keep the `--output screen` because... I guess otherwise it is not printed? Anyways, I'm not 100% sure on this. Feel free to remove it and keep the output to json only.
The `||` added to avoid failing on error.

```terminal
safety scan --output screen --save-as json info_safety.json || echo "Running 'safety' failed."
```

### DIFF
--- a/check-vulnerabilities/default_safety_policy.yaml
+++ b/check-vulnerabilities/default_safety_policy.yaml
@@ -17,13 +17,13 @@ report:
       cvss-severity: []
       vulnerabilities:
         52495:
-          reason: We are not considering this for the moment.
+          reason: We are not considering this vulnerability for the moment.
           expires: '2026-02-01'
         62044:
-          reason: We are not considering this for the moment.
+          reason: We are not considering this vulnerability for the moment.
           expires: '2026-02-01'
         67599:
-          reason: We are not considering this for the moment.
+          reason: We are not considering this vulnerability for the moment.
           expires: '2026-02-01'
 
 fail-scan-with-exit-code:


RobPasMue (changes requested):
It's not as simple as it looks...

The output is different and thus affects A LOT the python files used to postprocess the data from bandit. I would prevent doing this change for now. I have a local branch where I was doing this... we can talk offline.

germa89:
> It's not as simple as it looks...
> 
> The output is different and thus affects A LOT the python files used to postprocess the data from bandit. I would prevent doing this change for now. I have a local branch where I was doing this... we can talk offline.

Interesting... Yeah. I'm happy to help! Ping me tomorrow when available :)

moe-ad:
@RobPasMue and @germa89, please have there been progress on this?
Just wanted to know so we can plan v10.1 release better.

RobPasMue:
No progress here - and for now we might not want to change it I think. This requires major changes to the action

PR #667, opened 2025-01-07
feat: adding `skip-pdf-build` to documentation build action

### LINKED ISSUE #638: Add an option skip-pdf-build

Add an option ``skip-pdf-build``, similar to ``skip-json-build``, to prevent the build and deployment of pdf documentation.

Rationale: The errors in pdf documentation build require expertise to be fixed, not always available. In the meantime, these errors prevent ci to complete.

The workaround consisting of duplicating the ``doc-build`` action is not reasonable due to its complexity.

_No response_

_No response_

germa89:
As the title.

Close #638

### Notes

I'm not very skilled in powershell so it will be nice to have someone double checking the syntax and stuff. I guess we do not have any CICD tests for this yet.

### DIFF
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -322,7 +322,7 @@ runs:
         source .venv/bin/activate
         ${{ env.SPHINX_BUILD_MAKE }} -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
 
-        if [[ ${{ inputs.skip-pdf-build }} != 'true' ]];
+        if [[ ${{ inputs.skip-pdf-build }} == 'false' ]];
         then
           ${{ env.SPHINX_BUILD_MAKE }} -C doc pdf
         else
@@ -345,7 +345,7 @@ runs:
         source .venv/bin/activate
         xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
 
-        if [[ ${{ inputs.skip-pdf-build }} != 'true' ]];
+        if [[ ${{ inputs.skip-pdf-build }} == 'false' ]];
         then
           xvfb-run ${{ env.SPHINX_BUILD_MAKE }} -C doc pdf
         else
@@ -377,7 +377,7 @@ runs:
         python ${{ github.action_path }}/../python-utils/parse_doc_conf.py
 
     - uses: ansys/actions/_logging@main
-      if: ${{ ((inputs.add-pdf-html-docs-as-assets == 'true' ) && (env.PDF_FILENAME == '') ) && inputs.skip-pdf-build != 'true' }}
+      if: ${{ (inputs.skip-pdf-build == 'false') && (inputs.add-pdf-html-docs-as-assets == 'true' ) && (env.PDF_FILENAME == '') }}
       with:
         level: "ERROR"
         message: >
@@ -466,13 +466,9 @@ runs:
       run: |
         mv documentation-html.zip ${{ env.EXPECTED_DOWNLOAD_DIR }}
 
-        if [[ ${{ inputs.skip-pdf-build }} == 'true' ]];
-        then
+        if [[ ${{ inputs.skip-pdf-build }} == 'true' ]]; then
           echo "Not attaching PDF since its generation has not being requested."
-          exit 0;
-        fi
-
-        if [ -e  "${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }}" ]; then
+        elif [ -e  "${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }}" ]; then
             cp ${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }} ${{ env.EXPECTED_DOWNLOAD_DIR }}
             echo "File ${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }} has been copied."
         else
@@ -491,7 +487,7 @@ runs:
 
     - name: "Upload PDF documentation artifact"
       uses: actions/upload-artifact@v4
-      if: inputs.skip-pdf-build != 'true'
+      if: inputs.skip-pdf-build == 'false'
       with:
         name: documentation-pdf
         path: doc/_build/latex/*.pdf
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -96,6 +96,12 @@ inputs:
     required: true
     type: boolean
 
+  skip-pdf-build:
+    description: |
+      Whether to skip the generation of PDF documentation.
+    required: true
+    type: boolean
+
   check-links:
     description: |
       Whether to perform external link checks during the generation of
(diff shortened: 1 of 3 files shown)


_doc-build-linux/action.yml
```diff
@@ -445,6 +465,13 @@ runs:
...
         mv documentation-html.zip ${{ env.EXPECTED_DOWNLOAD_DIR }}
+
+        if [[ ${{ inputs.skip-pdf-build }} == 'true' ]];
+        then
+          echo "Not attaching PDF since its generation has not being requested."
+          exit 0;
+        fi
+
         if [ -e  "${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }}" ]; then
             cp ${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }} ${{ env.EXPECTED_DOWNLOAD_DIR }}
```
RobPasMue, line 476:
```suggestion

        if [[ ${{ inputs.skip-pdf-build }} == 'true' ]]; then
          echo "Not attaching PDF since its generation has not being requested."
        elif [ -e  "${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }}" ]; then
            cp ${{ env.EXPECTED_BUILD_DIR }}/latex/${{ env.PDF_FILENAME }} ${{ env.EXPECTED_DOWNLOAD_DIR }}
```

_doc-build-linux/action.yml
```diff
@@ -464,6 +491,7 @@ runs:
 
     - name: "Upload PDF documentation artifact"
       uses: actions/upload-artifact@v4
+      if: inputs.skip-pdf-build != 'true'
```
RobPasMue, line 494:
```suggestion
      if: inputs.skip-pdf-build == 'false'
```

RobPasMue (changes requested):
Left some comments - I mostly aligned the equalities so that they were in line with ``skip-json-build``. This will make it easier to maintain if both are checking the conditions in the same way. I also reformulated some of the conditionals. Apart from that the implementation looks good.

However, my main concern is whether we want to open the door to this or not. We've been opposing to it for some time because this way we know for a fact that libraries are generating their docs in PDF version. With this option... they can easily bypass that request. Pinging @ansys/pyansys-core for discussion...

I'll leave it as "request for changes" for now to prevent automerge. Once we decide I'll approve. Thanks @germa89!

_doc-build-linux/action.yml
germa89: I have asumed here than `exit 0` will just exit this step (code snippet). Am I right?

germa89:
I personally believe that because of the nature of the option "skip" and since we do not want to encourage skipping; I think evaluate ``!= 'true'`` is more strict than ``=='false'``, any mispelling and you will be skiping. 

Other than that, I am happy with the suggestions. I implemented them all.

_doc-build-linux/action.yml
RobPasMue: Since it is now a multiple-if statment, you do not need to worry

RobPasMue:
> evaluate `!= 'true'` is more strict than `=='false'`, any mispelling and you will be skiping.

When you consider them as strings, you are totally right. However, they are identified as booleans in the input type... meaning that you can just use one or the other (otherwise, GH complains)... consequently being equally strict.

MaxJPRey:
> Left some comments - I mostly aligned the equalities so that they were in line with `skip-json-build`. This will make it easier to maintain if both are checking the conditions in the same way. I also reformulated some of the conditionals. Apart from that the implementation looks good.
> 
> However, my main concern is whether we want to open the door to this or not. We've been opposing to it for some time because this way we know for a fact that libraries are generating their docs in PDF version. With this option... they can easily bypass that request. Pinging @ansys/pyansys-core for discussion...
> 
> I'll leave it as "request for changes" for now to prevent automerge. Once we decide I'll approve. Thanks @germa89!

OK, I have started to reflect on it. Thanks @germa89 and @RobPasMue. I will come back on it by tomorrow.

doc-build/action.yml
```diff
@@ -142,6 +142,13 @@ inputs:
     required: false
     type: boolean
 
+  skip-pdf-build:
+    description: |
+      Whether to skip the generation of PDF documentation.
```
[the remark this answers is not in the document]
RobPasMue: Agreed, yeah... if we decide to merge it, we should consider adding this warning.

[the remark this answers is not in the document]
germa89: Good catch @SMoraisAnsys !

SMoraisDev:
> Left some comments - I mostly aligned the equalities so that they were in line with `skip-json-build`. This will make it easier to maintain if both are checking the conditions in the same way. I also reformulated some of the conditionals. Apart from that the implementation looks good.
> 
> However, my main concern is whether we want to open the door to this or not. We've been opposing to it for some time because this way we know for a fact that libraries are generating their docs in PDF version. With this option... they can easily bypass that request. Pinging @ansys/pyansys-core for discussion...
> 
> I'll leave it as "request for changes" for now to prevent automerge. Once we decide I'll approve. Thanks @germa89!

I reflected a bit more on this and I'm really divided:
- The main reason why I would lean toward allowing PDF skip is that we are already allowing people to make a new github release without documentation. Since it's not mandatory (though I thought it was before), I don't see a reason why not bypassing PDF generation.
- On the other hand:
  - That could be the start to bypassing HTML and only creating PDF in the doc-build action (**I don't think we should allow that**)
  - Although I'd like to think that if someone disables PDF generation because of a temporary problem, he will find the time to fix it later. I can easily imagine cases where that later won't happen :)

Note that one can currently use our actions to perform a Github release while uploading PDF file and not HTML files (with `additional-artifacts`).

RobPasMue:
> Note that one can currently use our actions to perform a Github release while uploading PDF file and not HTML files (with `additional-artifacts`).

That's kind of a hack really...

https://github.com/ansys/actions/blob/9831f6ebee1e6bc8f8e140f68f136a841ae02c49/release-github/action.yml#L169-L176

If you are doing an ``only-code`` release, we are not uploading any kind of docs. Of course, if you upload your PDF as an additional artifact... that's it, you hacked it. But I am more worried about the logic/sense of the decision rather than the technical possibility. I am not worried about whether it can be done or not -  I'm sure you could hack it. 

IMO - it's fine if we open the door. But HTML would be the last one standing there on "default" scenarios and we should not avoid short-circuiting its generation for libraries... (only when you want to deliver code it is acceptable)

germa89:
We might be reflecting a bit too much on this... The fact we are not reaching a consensus among us is (for me) reason enought to not change anything. `ansys/actions` are meant to help to whoever wants to follow the standard (or recommended road). If the standard road is to have a PDF along with the HTML documentation... then we should NOT give the option for it.

If we were all agreing on this, I would see no problem on doing either way (PDF yes/no). But because there is no agreement/consensus and the recommended approach is to generate the PDF, then we should not have this feature included.

If you are having issues with the PDF, fix it. If you do not want to generate the PDF, well... figure out your own CICD :) #NotTryingToBeHarshHere #JustScopingPyAnsysCoreWork #WeCannotConsiderAndMantainAllCases #StopTalkingOnHashtags #Yeezzzzzz

germa89:
Soooo... we are *NOT* doing this....

PR #679, opened 2025-02-15
feat: add provenance attestation

SMoraisAnsys:
This PR consists in:
- allowing to add build provenance attestations for some workflow artifacts: sdist, wheel and wheelhouse;
- extending a release note with instruction on how to verify the provenance with [Github's CLI tool](https://cli.github.com/);
- allowing users to chose whether they want to have `generate_release_notes` for the release-github action or not.

To use it, one needs to add `attest-provenance: true` (default value is false) to the CICD job.

> [!NOTE]
The last bullet point is important because in some cases one could end up with a release note containing both pre generated releases notes AND notes generated using the changelog content. The behavior of `generate_release_notes` inside of the https://github.com/softprops/action-gh-release that we are leveraging is shown below:

image

> [!NOTE]
I decided to not attest documentation build since they are not critical (and also because HTML includes a lot of files).

> [!WARNING]
On top of the extra input, one needs to add permissions as below to the job. Otherwise, the action will fail.
```
permissions:
  id-token: write
  attestations: write
```


SMoraisAnsys:
Here is the result I just got from a dummy repo
image

I'll try it on a more elaborate projects to see how things are behaving.

RobPasMue (approved):
LGTM! Thanks @SMoraisAnsys

SMoraisAnsys:
Converting to draft to test removing the "What's changed" section in the release's body.

.github/workflows/ci_cd.yml
```diff
@@ -117,7 +117,8 @@ jobs:
     - name: "Install build and twine"
       shell: bash
       run: |
-        python -m pip install build twine
+        python -m pip install --upgrade build twine
+        python -m pip list
```
SMoraisAnsys, line 121:
```suggestion
        python -m pip install build twine
```

PR #698, opened 2025-02-26
fix: use ansys/pip-licenses to handle PEP 639

SMoraisAnsys:


SMoraisAnsys:
The action seems to work correctly with PyAEDT, see https://github.com/ansys/pyaedt/pull/5696
Note that since ansys/pip-licenses is python>=3.10, the PyAEDT license-check is disabled in python 3.8 and 3.9

check-licenses/action.yml
```diff
@@ -128,10 +128,22 @@ runs:
...
+      uses: actions/checkout@v4
+      with:
+        repository: ansys/pip-licenses
+        path: "external/pip-licenses"
+        ref: "main"
+
+    - name: Install ansys/pip-licenses main branch
+      shell: bash
+      run: |
+        pip install --no-cache-dir external/pip-licenses
```
RobPasMue, line 146: Why don't we release it to PyPI?

RobPasMue (commented):
I'm guessing my comment is just because "we are not there yet" and you are still validating things. If that's the reason, then fine =)

SMoraisAnsys:
@jorgepiloto We might want to release this fix in v9 because it has a breaking change (projects using the action to check python version <3.10 will fail).

check-licenses/action.yml
jorgepiloto, line 146:
```suggestion
        python -m pip install --no-cache-dir external/pip-licenses
```

check-licenses/action.yml
```diff
@@ -128,10 +128,22 @@ runs:
...
         use-cache: ${{ inputs.use-python-cache }}
 
-    - name: "Update pip and install the pip-licenses library"
+    - name: "Update pip"
       shell: bash
       run: |
-        python -m pip install -U pip pip-licenses
+        python -m pip install -U pip
+
+    - name: Clone ansys/pip-licenses on main branch
```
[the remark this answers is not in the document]
SMoraisAnsys: I don't think there is a need as the repo itself clearly state that the project is a fork from the original
image

check-licenses/action.yml
```diff
@@ -118,6 +118,25 @@ inputs:
...
+
+    - name: Break if Python version is too old
+      uses: ansys/actions/_logging@main
+      if: env.PYTHON_TOO_OLD == 'true'
+      with:
+        level: "ERROR"
+        message: >
+          Check licenses leverages Ansys fork of pip-licenses.
+          This package requires Python 3.10 or higher.
+
```
RobPasMue, line 139: Awesome, thanks a lot for the error handling!

build-wheelhouse/action.yml
```diff
@@ -120,6 +120,25 @@ runs:
...
+
+    - name: Break if Python version is too old
+      if: ${{ inputs.check-licenses == 'true' && env.PYTHON_TOO_OLD == 'true' }}
+      uses: ansys/actions/_logging@main
+      with:
+        level: "ERROR"
+        message: >
+          Check licenses leverages Ansys fork of pip-licenses.
+          This package requires Python 3.10 or higher.
+
```
RobPasMue, line 141:
```suggestion
```

Why not just handle this error handling at the check/licenses action? I know we would "lose time" by running the entire wheel action but I prefer that rather than duplicating code honestly. The license-check action can be run on standalone mode -- specific logic to it shouldn't appear in other actions IMO

RobPasMue (changes requested):

build-wheelhouse/action.yml
SMoraisAnsys: Fine by me, I was focusing on avoiding using runner time but your point is also valid.
I'll do that associated changes.

SMoraisAnsys:
Here is the result I just got 

image

SMoraisAnsys:
After the duplicate removal

image

RobPasMue (approved):
Awesome, thanks a lot @SMoraisAnsys 🚀

check-licenses/action.yml
```diff
@@ -118,6 +118,25 @@ inputs:
...
+          echo "PYTHON_TOO_OLD=true" >> "$GITHUB_ENV"
+        fi
+
+    - name: Break if Python version is too old
+      uses: ansys/actions/_logging@main
+      if: env.PYTHON_TOO_OLD == 'true'
+      with:
+        level: "ERROR"
+        message: >
+          Check licenses leverages Ansys fork of pip-licenses.
```
SMoraisAnsys, line 137:
```suggestion
          The check-licenses action leverages Ansys fork of pip-licenses.
```

PR #700, opened 2025-02-27
fix: whitespace control in changelog template

Revathyvenugopal162:
in favor of https://github.com/ansys/pre-commit-hooks/pull/294/files


jorgepiloto (approved):
Thanks for fixing this, @Revathyvenugopal162. It is always challenging to catch up with blank spaces set or removed by Jinja.

PR #701, opened 2025-02-27
ci: split workflows

jorgepiloto:
This pull-request splits the workflow file into four. Each workflow targets:

- Pull-requests
- Main branch
- Releases
- Nightly build

Main reason for splitting the workflow is to ease its maintenance for each scenario. Even if we use conditionals, we end up having too much of them.


.github/workflows/ci_cd_main.yml
```diff
@@ -0,0 +1,41 @@
...
+env:
+  MAIN_PYTHON_VERSION: '3.12'
+  DOCUMENTATION_CNAME: 'actions.docs.ansys.com'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
```
RobPasMue, line 16: This one will need contents: write if I am not mistaken @jorgepiloto -- you need to publish to gh-pages

.github/workflows/ci_cd_night.yml
```diff
@@ -0,0 +1,86 @@
...
+    - cron: '0 0 * * *'
+
+env:
+  MAIN_PYTHON_VERSION: '3.12'
+  LIBRARY_NAME: 'ansys-actions'
+
+permissions:
+  contents: read
+  pull-requests: read
+
```
RobPasMue, line 14: Same as before, might need contents: write

.github/workflows/ci_cd_night.yml
```diff
@@ -0,0 +1,86 @@
+name: pull-request
```
RobPasMue, line 1: Is this a nightly workflow? If so adapt the name please

RobPasMue (changes requested):

.github/workflows/ci_cd_night.yml
SMoraisAnsys: @jorgepiloto Just an open question but do we require a nightly workflow ?

I think it makes sense to have it as it will help us catching issues that are happening outside of the changes of a PR. For example, that would have helped with the recent `twine` issue. However, it increase our consumption of runners :(
Maybe there might be other use cases that I'm not thinking about and we can leave my comment to the garbage :D

jorgepiloto: Thanks for raising this concern, @SMoraisAnsys. However, this project is public and CI/CD minutes are not billed into our organization.

jorgepiloto:
Fixed permissions and names in https://github.com/ansys/actions/pull/701/commits/2ff1dbe3aa32c81c5bc8796c49fbfc582b9b8242, @RobPasMue

RobPasMue (approved):
LGTM!

PR #704, opened 2025-02-27
feat: add logging into workflow summary

SMoraisAnsys:
Adding the logging of warning and error into the check summary. This could be a great enhancement in the way we share future deprecation of features (as it might be the case with the pypi-release actions). Also, the logging was refactored to use fewer number of steps (by deciding shell value with conditional).

Here is the result when testing in [1d0eb0c](https://github.com/ansys/actions/pull/704/commits/1d0eb0c325a124bfe5691c4f9846cf7a38264c73)

The previous INFO logging level is kept untouched as I think that it improves reading the workflow logs.

image

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -47,7 +47,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: labeler
     steps:
-      - uses: ansys/actions/check-pr-title@feat/add-log-in-workflow-summary
+      - uses: ansys/actions/check-pr-title@main
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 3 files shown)


RobPasMue (approved):

jorgepiloto (approved):

MaxJPRey (approved):
LGTM.

RobPasMue:
I would remove them yeah, the built-in feature using the ``::<type>::`` seems like the right approach :)

jorgepiloto:
> I would remove them yeah, the built-in feature using the `::<type>::` seems like the right approach :)

Agreed.

SMoraisAnsys:
The refactoring of the shell value is working, see the shell value below (when run on a PyAEDT windows self-hosted runner)

image

jorgepiloto (approved):
Excellent!

PR #706, opened 2025-02-28
fix: if statement condition in logging

SMoraisAnsys:
I left a typo in #704, this should fix it.


jorgepiloto (approved):
Thanks for fixing this. My bad when approving and merging to fast #704.

SMoraisAnsys:
> Thanks for fixing this. My bad when approving and merging to fast #704.

That's on me, I posted a screen done before doing the cleanup.... I should have clean things morecarefully

PR #707, opened 2025-02-28
refactor: warn about release pypi deprecation and encourage to use trusted publisher

SMoraisAnsys:
This PR contains:
- a notification to motivate users to use the trusted publisher approach;
- a notification on the pypi-release future deprecation.

Note that we might want tot keep the release-pypi actions and reduce their scope to only work with twine token.
In the case of someone using them with trusted publisher, we could redirect to the documentation page on how to update the CICD workflow.

---

**OPEN TO DISCUSSION**

Should we enforce trusted published approach and totally remove the release-pypi action or restrict the scop as suggested above ? I would vote for a restriction of the scope even if I think it would be wise for all our products to leverage the trusted publisher.


_release-pypi/action.yml
```diff
@@ -100,6 +100,12 @@ runs:
   using: "composite"
   steps:
 
+    - name: "Notify to use trusted publisher"
+      shell: ${{ runner.os == 'Windows' && 'powershell' || 'bash' }}
```
jorgepiloto, line 104: Nice! This is super. I didn't know this was valid for setting the shell.

doc/source/release-actions/index.rst
```diff
@@ -31,3 +31,44 @@ Release GitHub action
...
+        with:
+            name: ${{ env.PACKAGE_NAME }}-artifacts
+            path: ${{ env.PACKAGE_NAME }}-artifacts
+
+        - name: "Display the structure of downloaded files"
+        shell: bash
+        run: ls -R
+
+        - name: "Upload artifacts to PyPI using Trusted Publisher"
+        uses: pypa/gh-action-pypi-publish@v1.10.1
```
jorgepiloto, line 68: Suggesting to add here the latest version of the action:

```suggestion
        uses: pypa/gh-action-pypi-publish@v1.12.4
```

jorgepiloto (commented):
Approved on my side. Since this includes some documentation, it would be great if @PipKat could give this a look. Thanks, @SMoraisAnsys.

doc/source/release-actions/index.rst
SMoraisAnsys: Nice catch !

_release-pypi/action.yml
SMoraisAnsys: I gave it a first try in the log refactoring and it worked like a charm :)

doc/source/release-actions/index.rst
```diff
@@ -31,3 +31,53 @@ Release GitHub action
...
+    runs-on: ubuntu-latest
+    needs: [build-library]
+    # Specifying a GitHub environment is optional, but strongly encouraged
+    environment: release
+    permissions:
+        # IMPORTANT: this permission is mandatory for trusted publishing
+        id-token: write
+    if: github.event_name == 'push' && contains(github.ref, 'refs/tags')
+    steps:
+        - name: "Download the library artifacts from build-library step"
```
RobPasMue, line 55:
```suggestion
      - name: "Download the library artifacts from build-library step"
```

doc/source/release-actions/index.rst
```diff
@@ -31,3 +31,53 @@ Release GitHub action
...
+        uses: actions/download-artifact@v4
+        with:
+            name: ${{ env.PACKAGE_NAME }}-artifacts
+            path: ${{ env.PACKAGE_NAME }}-artifacts
+
+        - name: "Display the structure of downloaded files"
```
RobPasMue, line 61:
```suggestion
      - name: "Display the structure of downloaded files"
```

doc/source/release-actions/index.rst
```diff
@@ -31,3 +31,53 @@ Release GitHub action
...
+        shell: bash
+        run: ls -R
+
+        - name: "Upload artifacts to PyPI using Trusted Publisher"
```
RobPasMue, line 65:
```suggestion
      - name: "Upload artifacts to PyPI using Trusted Publisher"
```

release-pypi-private/action.yml
```diff
@@ -110,6 +110,18 @@ runs:
...
+        level: "WARNING"
+        message: >
+          As the python ecosystem is moving to project metadata version 2.4,
+          we need to use a more recent version of pypa/gh-action-pypi-publish.
+          However, this action is not compatible with composite actions.
+          Therefore, the `ansys/release-pypi-private` action is destined to
+          disappear in a future version. We invite all users to define their
+          release job manually. As an example, see
+          https://actions.docs.ansys.com/version/stable/release-actions/index.html#release-pypi-template
+
```
RobPasMue, line 124: I don't think the private PyPI action will ever disappear in fact. This is the one used to publish to our private PyPI... which we will still need I guess because the Trusted Publisher approach doesn't work there

doc/source/release-actions/index.rst
```diff
@@ -31,3 +31,53 @@ Release GitHub action
 .. jinja:: release-github
     :file: _templates/action.rst.jinja
 
+Release PyPI template
+---------------------
+
+Below, it is assumed that a job `build-library` already created library sources
+and wheel artifacts using the environment variable `PACKAGE_NAME`. Also, the job
+is defined to not upload files if one already exists thanks to the
+`skip-existing` input.
```
RobPasMue, line 40: I think we need to explain a bit more why we need to change to having it separated in the workflow... currently this doesn't explain the reasons, right?

RobPasMue (changes requested):
Left some comments =)

doc/source/release-actions/index.rst
SMoraisAnsys: You're right, a quick explanation is triggered in the pypi-release actions but it could be added here and extended.

PipKat, line 40:
```suggestion
The following code assumes that a `build-library` job has already created library sources
and wheel artifacts using the `PACKAGE_NAME` environment variable. Also, the job's 
`skip-existing` input indicates that files that already exist are not uploaded again.
```
It's clearer when the noun describing the code entity follows the code entity. Also, we should avoid placement indicators like above and below and use preceding and following instead. This is because a PDF of the documentation might have placed content on a preceding or following page.

_release-pypi/action.yml
```diff
@@ -100,6 +100,12 @@ runs:
...
+      if: inputs.use-trusted-publisher == 'false'
+      run: |
+        echo "::notice::We recommend using Trusted Publishers to securely publish packages on PyPI. Please consider moving to this approach."
```
PipKat, line 107:
```suggestion
        echo "::notice::We recommend using trusted publishers to securely publish packages on PyPI. You should consider moving to this approach."
```
It doesn't appear that Trusted Publishers should be treated as a proper noun. Avoid "please" in technical documentation (although this does seem to be changing as it is so avoid used in writing prompts for AI)!

doc/source/release-actions/index.rst
```diff
@@ -31,3 +31,53 @@ Release GitHub action
...
+
+.. code::yaml
+
+    release-pypi:
+    name: "Release to PyPI with trusted publisher approach"
+    runs-on: ubuntu-latest
+    needs: [build-library]
+    # Specifying a GitHub environment is optional, but strongly encouraged
```
PipKat, line 48:
```suggestion
    # Specifying a GitHub environment is optional but strongly encouraged
```

doc/source/release-actions/index.rst
PipKat: Replace "Trusted Publisher" with "trusted pulisher approach"?

doc/source/release-actions/index.rst
```diff
@@ -31,3 +31,53 @@ Release GitHub action
...
+        - name: "Upload artifacts to PyPI using Trusted Publisher"
+        uses: pypa/gh-action-pypi-publish@v1.12.4
+        with:
+            repository-url: "https://upload.pypi.org/legacy/"
+            print-hash: true
+            packages-dir: ${{ env.PACKAGE_NAME }}-artifacts
+            skip-existing: false
+
+When used to test the release process, you can update the repository URL as
+follow:
```
PipKat, line 74:
```suggestion
When testing the release process, you can update the repository URL as
follows:
```
I hope I didn't change the meaning, but the subject "you" isn't related to "When used to test the release process, ...." For the existing construction, you would need to have something like this: "When used to test the release process, the following action updates the repository URL."

doc/source/release-actions/index.rst
jorgepiloto: Just added your feedback to the suggestion.

doc/source/release-actions/index.rst
jorgepiloto: I moved this to the migration guide. We have a dedicated page for it. It was created when introducing the trusted publishers for the first time.

Also, I added warnings in the description of every release action. Now, those are included in the docs too.

doc/source/migrations/release-pypi-trusted-publisher.rst
```diff
@@ -1,47 +1,51 @@
...
-          - name: Release to the public PyPI repository
-            uses: ansys/actions/release-pypi-public@v6
+
+          - name: "Download the library artifacts from build-library step"
+            uses: actions/download-artifact@v4
             with:
-              library-name: ${{ env.PACKAGE_NAME }}
-              use-trusted-publisher: true
+                name: ${{ env.PACKAGE_NAME }}-artifacts
+                path: ${{ env.PACKAGE_NAME }}-artifacts
```
jorgepiloto, line 38:
```suggestion
              name: ${{ env.PACKAGE_NAME }}-artifacts
              path: ${{ env.PACKAGE_NAME }}-artifacts
```

release-pypi-private/action.yml
```diff
@@ -30,6 +30,13 @@ description: |
   file is expected to contain a wheel and a source distribution file for the
   desired library.
 
+  .. attention::
+
+     Starting with version ``ansys-actions@v9``, the ``use-trusted-publisher``
+     input parameter is deprecated. Refer to the
+     :ref:`release_pypi_trusted_publisher` if you wish to use this security
+     feature.
```
jorgepiloto, line 38:
```suggestion
      Starting with version ``ansys-actions@v9``, the ``use-trusted-publisher``
      input parameter is deprecated. Refer to the
      :ref:`release_pypi_trusted_publisher` if you wish to use this security
      feature.
```

release-pypi-public/action.yml
```diff
@@ -29,6 +29,13 @@ description: |
   named ``<library-name>-artifacts``. This file is expected to contain a wheel file
   and a source distribution file for the desired library.
 
+  .. attention::
+
+     Starting with version ``ansys-actions@v9``, the ``use-trusted-publisher``
+     input parameter is deprecated. Refer to the
+     :ref:`release_pypi_trusted_publisher` if you wish to use this security
+     feature.
```
jorgepiloto, line 37:
```suggestion
      Starting with version ``ansys-actions@v9``, the ``use-trusted-publisher``
      input parameter is deprecated. Refer to the
      :ref:`release_pypi_trusted_publisher` if you wish to use this security
      feature.
```

release-pypi-test/action.yml
```diff
@@ -30,6 +30,13 @@ description: |
   file is expected to contain a wheel file and a source distribution file for
   the desired library.
 
+  .. attention::
+
+     Starting with version ``ansys-actions@v9``, the ``use-trusted-publisher``
+     input parameter is deprecated. Refer to the
+     :ref:`release_pypi_trusted_publisher` if you wish to use this security
+     feature.
```
jorgepiloto, line 38:
```suggestion
      Starting with version ``ansys-actions@v9``, the ``use-trusted-publisher``
      input parameter is deprecated. Refer to the
      :ref:`release_pypi_trusted_publisher` if you wish to use this security
      feature.
```

release-pypi-test/action.yml
```diff
@@ -103,6 +110,18 @@ runs:
   using: "composite"
   steps:
 
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "WARNING"
+        message: >
+          As the python ecosystem is moving to project metadata version 2.4,
```
jorgepiloto, line 117:
```suggestion
          As the Python ecosystem is moving to project metadata version 2.4,
```

jorgepiloto (approved):

PR #714, opened 2025-03-04
fix(ci): syntax

jorgepiloto:
Fixes the issues raised in https://github.com/ansys/actions/actions/runs/13656832593/job/38178046621.


moe-ad (approved):
LGTM

PR #723, opened 2025-03-06
fix: install for Python libraries

jorgepiloto:
As pointed out by @SMoraisAnsys, https://github.com/ansys/actions/pull/723 I removed the installation of libraries thinking this was not required. However, Towncrier requires Python projects to be installed to retrieve their current version.

### DIFF
--- a/doc-deploy-changelog/action.yml
+++ b/doc-deploy-changelog/action.yml
@@ -133,6 +133,15 @@ inputs:
     default: true
     type: boolean
 
+  skip-install:
+    description: |
+      Whether to skip the installation of the project or not. If ``true``, the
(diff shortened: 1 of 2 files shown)


SMoraisAnsys (approved):

SMoraisAnsys:
For reference, this is the run that failed due to the missing installation https://github.com/ansys/pyspeos/actions/runs/13700064250/job/38311335066
It seems that you need to install the project you want to use `towncrier build`. It's probably due to the fact that we are passing information to `towncrier` through the pyproject.toml file. From the documentation, I wasn't able to find a way to avoid this installation requirement.

PR #725, opened 2025-03-06
feat: check actions security action

moe-ad:


.github/workflows/ci_cd_pr.yml
```diff
@@ -27,6 +27,8 @@ jobs:
 
      - name: "Checkout project"
        uses: actions/checkout@v4
+       with:
+        persist-credentials: false
```
jorgepiloto, line 31: Why do we need this?

check-actions-security/action.yml
```diff
@@ -0,0 +1,112 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+name: |
+  Check actions security action
```
jorgepiloto, line 24:
```suggestion
  Check actions security action.
```

check-actions-security/action.yml
```diff
@@ -0,0 +1,112 @@
...
+
+description: |
+  Detects common security issues in GitHub Actions yml files using `zizmor <https://woodruffw.github.io/zizmor/>`_.
+
+inputs:
+
+  token:
+    description: |
+      Use the PYANSYS_CI_BOT_TOKEN to allow online audits by zizmor
```
jorgepiloto, line 33:
```suggestion
      Use the PYANSYS_CI_BOT_TOKEN to allow online audits by zizmor.
```

check-actions-security/action.yml
```diff
@@ -0,0 +1,112 @@
...
+      shell: bash
+      run: |
+        uv tool install zizmor
+        zizmor --version
+
+    - name: "Optionally set GH_TOKEN"
+      if: ${{ inputs.token != '' }}
+      shell: bash
+      run: |
+        echo 'GH_TOKEN=${{ inputs.token }}' >> "$GITHUB_ENV"
```
jorgepiloto, line 73: I do not think secrets are inherited in composite actions unless specified.

check-actions-security/action.yml
```diff
@@ -0,0 +1,112 @@
...
+
+    - name: "Output a summary of detected vulnerabilities"
+      if: ${{ inputs.generate-summary == 'true' }}
```
jorgepiloto, line 76: GitHub Actions syntax is very tricky. I think this should be:

```suggestion
      if: inputs.generate-summary == 'true'
```

jorgepiloto (commented):
Just left some minor comments. Thanks for the work in here, @moe-ad.

.github/workflows/ci_cd_pr.yml
moe-ad: I have started resolving the raised vulnerability issues on this same PR. This was flagged as well.
Should we address resolving the raised issues on a separate PR?

check-actions-security/action.yml
moe-ad: My mistake. `token` in the input list is `GH_TOKEN`, not `PYANSYS_CI_BOT_TOKEN`. So this is expected to be passed as an optional input when using the action.

check-actions-security/action.yml
moe-ad: This syntax and the syntax you have suggested are both correct according to the [documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#about-expressions):

> [!NOTE]  
> The exception to this rule is when you are using expressions in an if clause, where, optionally, **you can usually omit ${{ and }}**. For more information about if conditionals, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif).

I just decided to explicitly use `${{ <expression> }}` since it seems like the best practice. But I can drop it if you want us to stick to the suggested form instead. I bet that is how it is other action files within the repo.

jorgepiloto: Thanks for adding here that documentation reference! Then, let's stick to your original commit.

moe-ad:
There are situations where we definitely still need to use $GITHUB_ENV e.g. `ansys/actions/check_vulnerabilities` where environment variables are needed by the `check-vulnerabilities.py` script. The only way to ignore audit rules on composite actions is via [inline comments](https://woodruffw.github.io/zizmor/configuration/#precedence).

> [!WARNING] 
> [Zizmor configuration file isn't supported](https://woodruffw.github.io/zizmor/configuration/#precedence)
> image

However, it seems there is a bug impeding proper behavior of this feature within composite actions. At least, I couldn't get `# zizmor ignore[github-env]` to work. We might need to open an issue on the `zizmor` repo itself.

moe-ad:
Issue opened: https://github.com/woodruffw/zizmor/issues/612

moe-ad:
@jorgepiloto I finished correcting the raised zizmor issues. You can start reviewing this PR.

The auditing level in `ci_cd_pr.yml` is currently set to `normal`. When `high` auditing level is used, the issues raised relate to using commit SHA instead of tags, similar to what @SMoraisAnsys mentioned yesterday. Using `strict` auditing level raised no issues. As @SMoraisAnsys already pointed out [on this PR](https://github.com/ansys/actions/pull/739), we don't need warnings about commit SHA vs tags on our internal actions and once his PR is merged, external actions will be using commit SHA. So I left the auditing level within `ci_cd_pr.yml` as `normal`

However, I have set the default auditing level within `actions/check-actions-security` itself to `high` as it makes sense for pyansys repos using the action to have that option as the default. Should a pyansys repo choose to go with this default but want to trust `ansys/actions` (so they still can use the tag approach when we make new releases), which I guess they will, they will just have to add `# zizmor: ignore[unpinned-uses]`  in front of any ansys action being used to mark them for zizmor to ignore.

Let me know what you think about these decisions.

SMoraisAnsys:
Seems like we could leverage https://woodruffw.github.io/zizmor/configuration/#rulesidignore but I don't think there is a simple way to filter our actions

moe-ad:
@SMoraisAnsys to buttress your point, I also can't see what works for us currently wrt the current methods of ignoring things in zizmor.  What do you think about parsing the zizmor's output and overriding with a 0 exit code if the ffl are satisfied:
- All raised issues are due to `unpinned-uses` audit rule &&
- The `unpinned-uses` issues are due to `ansys/actions`

Otherwise we might need to request a feature which allows fine-grained ignore rules controls, the maintainer seems very responsive.

SMoraisAnsys:
> @SMoraisAnsys to buttress your point, I also can't see what works for us currently wrt the current methods of ignoring things in zizmor. What do you think about parsing the zizmor's output and overriding with a 0 exit code if the ffl are satisfied:
> 
> * All raised issues are due to `unpinned-uses` audit rule &&
> * The `unpinned-uses` issues are due to `ansys/actions`
> 
> Otherwise we might need to request a feature which allows fine-grained ignore rules controls, the maintainer seems very responsive.

I think contacting the maintainer is a better option since it would also benefit other organization implementing their own composite actions. Moreover, in my experience, parsing the output of an app can be a real pain if the output / format changes even just a bit. I know that people in the open source world are more aware of breaking changes but I would avoid it if possible.

moe-ad:
Feature requested:https://github.com/woodruffw/zizmor/issues/626

_doc-build-windows/action.yml
```diff
@@ -334,16 +349,21 @@ runs:
...
-          ${{ env.SPHINX_BUILD_MAKE }} linkcheck SPHINXOPTS="${{ inputs.sphinxopts }}"
+        $env:SPHINX_BUILD_MAKE html SPHINXOPTS="$env:SPHINXOPTS"
+        $env:SPHINX_BUILD_MAKE pdf
+        if ("$env:CHECK_LINKS" -eq 'true' ) {
+          $env:SPHINX_BUILD_MAKE linkcheck SPHINXOPTS="$env:SPHINXOPTS"
         }
-        if ("${{ inputs.skip-json-build }}" -eq 'false' ) {
-          ${{ env.SPHINX_BUILD_MAKE }} json SPHINXOPTS="${{ inputs.sphinxopts }}"
+        if ("$env:SKIP_JSON_BUILD" -eq 'false' ) {
+          $env:SPHINX_BUILD_MAKE json SPHINXOPTS="$env:SPHINXOPTS"
```
moe-ad:
```suggestion
        if ("$env:SKIP_JSON_BUILD" -eq 'false' ) {
          & $env:SPHINX_BUILD_MAKE json SPHINXOPTS="$env:SPHINXOPTS"
```

doc-style/action.yml
```diff
@@ -101,23 +101,28 @@ runs:
...
+      id: vale-major-version
       shell: bash
+      env:
+        VALE_VERSION: ${{ inputs.vale-version }}
       run: |
-        version=${{ inputs.vale-version }}
+        version="$VALE_VERSION"
         vale_major_version=$( echo $version | cut -d '.' -f 1)
-        echo "VALE_MAJOR_VERSION=$(echo $vale_major_version)" >> $GITHUB_ENV
+        echo "VALE_MAJOR_VERSION=$(echo $vale_major_version)" >> $GITHUB_OUTPUT
```
jorgepiloto, line 119:
```suggestion
        echo "VALE_MAJOR_VERSION=$(echo $vale_major_version)" >> ${GITHUB_OUTPUT}
```

tests-pytest/action.yml
```diff
@@ -277,13 +299,25 @@ runs:
...
+       PYTEST_MARKERS: ${{ inputs.pytest-markers }}
+       PYTEST_EXTRA_ARGS: ${{ inputs.pytest-extra-args }}
+       PYTEST_POST_ARGS: ${{ inputs.pytest-postargs }}
+       PYTEST_RUN: ${{ steps.output.pytest-run-command.PYTEST_RUN }}
+       ACTIVATE_VENV: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
       run: |
-        ${{ env.ACTIVATE_VENV }}
-        xvfb-run ${{ env.PYTEST_RUN }} ${{ inputs.pytest-markers }} ${{ inputs.pytest-extra-args }} ${{ inputs.pytest-postargs }}
+        $ACTIVATE_VENV
+        xvfb-run $PYTEST_RUN $PYTEST_MARKERS $PYTEST_EXTRA_ARGS $PYTEST_POST_ARGS
```
jorgepiloto, line 323:
```suggestion
        xvfb-run ${PYTEST_RUN} ${PYTEST_MARKERS} ${PYTEST_EXTRA_ARGS} ${PYTEST_POST_ARGS}
```

jorgepiloto (changes requested):
Thanks for this massive contribution, @moe-ad. I left some minor suggestions, most of them related with explicit variable expansion. Others include basic improvements.

Before merging this, all references to the current branch must be updated to point to the default branch.

_doc-build-windows/action.yml
```diff
@@ -291,64 +297,78 @@ runs:
...
         .venv\Scripts\Activate.ps1
         cd doc
-        ${{ env.SPHINX_BUILD_MAKE }} html SPHINXOPTS="${{ inputs.sphinxopts }}"
-        ${{ env.SPHINX_BUILD_MAKE }} pdf
-        if ("${{ inputs.check-links }}" -eq 'true' ) {
-          ${{ env.SPHINX_BUILD_MAKE }} linkcheck SPHINXOPTS="${{ inputs.sphinxopts }}"
+        & $env:SPHINX_BUILD_MAKE html SPHINXOPTS="$env:SPHINXOPTS"
+        & $env:SPHINX_BUILD_MAKE pdf
+        if ($env:CHECK_LINKS -eq 'true') {
+          & $env:SPHINX_BUILD_MAKE linkcheck SPHINXOPTS="$env:SPHINXOPTS"
```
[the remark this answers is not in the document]
moe-ad: This is a PowerShell requirement. When an environment variable expands into a command on pwsh, `&` must precede such expansion to "instruct" the shell to run the command.

_doc-build-linux/action.yml
```diff
@@ -131,24 +131,28 @@ inputs:
     required: true
     type: boolean
 
-runs:
+runs: # zizmor: ignore[github-env]
```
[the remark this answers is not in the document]
moe-ad: @jorgepiloto Thanks for catching this. woodruffw/zizmor#616 should have fixed this. I will make adjustments.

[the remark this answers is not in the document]
jorgepiloto: Thanks for clarifying.

.github/workflows/ci_cd_main.yml
```diff
@@ -12,16 +12,15 @@ concurrency:
...
+permissions: {}
 
 jobs:
 
   doc-build:
     name: "Doc build"
     runs-on: ubuntu-latest
     steps:
-      - uses: ansys/actions/doc-build@main
+      - uses: ansys/actions/doc-build@feat/check-actions-security-action
```
RobPasMue, line 23:
```suggestion
      - uses: ansys/actions/doc-build@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_main.yml
```diff
@@ -32,8 +31,10 @@ jobs:
     name: "Deploy development documentation"
     runs-on: ubuntu-latest
     needs: [doc-build]
+    permissions:
+      contents: write
     steps:
-      - uses: ansys/actions/doc-deploy-dev@main
+      - uses: ansys/actions/doc-deploy-dev@feat/check-actions-security-action
```
RobPasMue, line 37:
```suggestion
      - uses: ansys/actions/doc-deploy-dev@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_night.yml
```diff
@@ -22,7 +21,7 @@ jobs:
     name: "Doc build"
     runs-on: ubuntu-latest
     steps:
-      - uses: ansys/actions/doc-build@main
+      - uses: ansys/actions/doc-build@feat/check-actions-security-action
```
RobPasMue, line 24:
```suggestion
      - uses: ansys/actions/doc-build@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_night.yml
```diff
@@ -9,8 +9,7 @@ env:
   LIBRARY_NAME: 'ansys-actions'
   DOCUMENTATION_CNAME: 'actions.docs.ansys.com'
 
-permissions:
-  contents: write
+permissions: {}
```
RobPasMue, line 12:
```suggestion
```
If this is the case, I'd remove it altogether

.github/workflows/ci_cd_night.yml
```diff
@@ -33,8 +32,10 @@ jobs:
     name: "Deploy development documentation"
     runs-on: ubuntu-latest
     needs: [doc-build]
+    permissions:
+      contents: write
     steps:
-      - uses: ansys/actions/doc-deploy-dev@main
+      - uses: ansys/actions/doc-deploy-dev@feat/check-actions-security-action
```
RobPasMue, line 38:
```suggestion
      - uses: ansys/actions/doc-deploy-dev@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_night.yml
```diff
@@ -51,9 +52,11 @@ jobs:
 
     - name: "Install Git and clone project"
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        persist-credentials: false
 
     - name: "Set up Python"
-      uses: ansys/actions/_setup-python@main
+      uses: ansys/actions/_setup-python@feat/check-actions-security-action
```
RobPasMue, line 59:
```suggestion
      uses: ansys/actions/_setup-python@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_pr.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     steps:
       - name: Decide which environment should be used
         id: block
-        uses: ansys/actions/check-environment-approval@main
+        uses: ansys/actions/check-environment-approval@feat/check-actions-security-action
```
RobPasMue, line 28:
```suggestion
        uses: ansys/actions/check-environment-approval@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_pr.yml
```diff
@@ -80,11 +82,11 @@ jobs:
       pull-requests: write
     steps:
 
-      - uses: ansys/actions/check-pr-title@main
+      - uses: ansys/actions/check-pr-title@feat/check-actions-security-action
```
RobPasMue, line 85:
```suggestion
      - uses: ansys/actions/check-pr-title@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_pr.yml
```diff
@@ -80,11 +82,11 @@ jobs:
...
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      - uses: ansys/actions/doc-changelog@main
+      - uses: ansys/actions/doc-changelog@feat/check-actions-security-action
```
RobPasMue, line 89:
```suggestion
      - uses: ansys/actions/doc-changelog@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_pr.yml
```diff
@@ -98,7 +100,7 @@ jobs:
     needs: changelog-fragment
     steps:
       - name: "Run code style checks"
-        uses: ansys/actions/code-style@main
+        uses: ansys/actions/code-style@feat/check-actions-security-action
```
RobPasMue, line 103:
```suggestion
        uses: ansys/actions/code-style@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_pr.yml
```diff
@@ -108,7 +110,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: changelog-fragment
     steps:
-      - uses: ansys/actions/doc-style@main
+      - uses: ansys/actions/doc-style@feat/check-actions-security-action
```
RobPasMue, line 113:
```suggestion
      - uses: ansys/actions/doc-style@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_pr.yml
```diff
@@ -117,7 +119,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: doc-style
     steps:
-      - uses: ansys/actions/doc-build@main
+      - uses: ansys/actions/doc-build@feat/check-actions-security-action
```
RobPasMue, line 122:
```suggestion
      - uses: ansys/actions/doc-build@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_pr.yml
```diff
@@ -135,9 +137,11 @@ jobs:
 
     - name: "Install Git and clone project"
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        persist-credentials: false
 
     - name: "Set up Python"
-      uses: ansys/actions/_setup-python@main
+      uses: ansys/actions/_setup-python@feat/check-actions-security-action
```
RobPasMue, line 144:
```suggestion
      uses: ansys/actions/_setup-python@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_release.yml
```diff
@@ -33,47 +39,59 @@ jobs:
...
+          echo "Deleted v${MAJOR}.${MINOR} tag" || \
+          echo "Tag v${MAJOR}.${MINOR} not found"
 
     - name: "Create new tags"
+      env:
+        MAJOR: ${{ steps.current-branch-tag-validity.outputs.MAJOR }}
+        MINOR: ${{ steps.tag-components.outputs.MINOR }}
       run: |
-        git tag v${{ env.MAJOR }}.${{ env.MINOR }}
-        git tag v${{ env.MAJOR }}
```
RobPasMue, line 74: Why getting rid of all the ${{ env.* }} ?? That's the way to call the env variables on a github workflow :S

.github/workflows/ci_cd_release.yml
```diff
@@ -83,7 +101,7 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      - uses: ansys/actions/doc-deploy-changelog@main
+      - uses: ansys/actions/doc-deploy-changelog@feat/check-actions-security-action
```
RobPasMue, line 104:
```suggestion
      - uses: ansys/actions/doc-deploy-changelog@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_release.yml
```diff
@@ -95,7 +113,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: rolling-release
     steps:
-      - uses: ansys/actions/code-style@main
+      - uses: ansys/actions/code-style@feat/check-actions-security-action
```
RobPasMue, line 116:
```suggestion
      - uses: ansys/actions/code-style@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_release.yml
```diff
@@ -104,8 +122,10 @@ jobs:
     name: "Doc style"
     runs-on: ubuntu-latest
     needs: rolling-release
+    permissions:
+      contents: write
     steps:
-      - uses: ansys/actions/doc-style@main
+      - uses: ansys/actions/doc-style@feat/check-actions-security-action
```
RobPasMue, line 128:
```suggestion
      - uses: ansys/actions/doc-style@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_release.yml
```diff
@@ -114,7 +134,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: doc-style
     steps:
-      - uses: ansys/actions/doc-build@main
+      - uses: ansys/actions/doc-build@feat/check-actions-security-action
```
RobPasMue, line 137:
```suggestion
      - uses: ansys/actions/doc-build@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_release.yml
```diff
@@ -132,9 +152,11 @@ jobs:
 
     - name: "Install Git and clone project"
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        persist-credentials: false
 
     - name: "Set up Python"
-      uses: ansys/actions/_setup-python@main
+      uses: ansys/actions/_setup-python@feat/check-actions-security-action
```
RobPasMue, line 159:
```suggestion
      uses: ansys/actions/_setup-python@main
```
Needs to be reverted before merging

.github/workflows/ci_cd_release.yml
```diff
@@ -221,8 +243,10 @@ jobs:
     name: "Deploy stable documentation"
     runs-on: ubuntu-latest
     needs: release
+    permissions:
+      contents: write
     steps:
-      - uses: ansys/actions/doc-deploy-stable@main
+      - uses: ansys/actions/doc-deploy-stable@feat/check-actions-security-action
```
RobPasMue, line 249:
```suggestion
      - uses: ansys/actions/doc-deploy-stable@main
```
Needs to be reverted before merging

RobPasMue (commented):
We need to revert all the refs to the branch by "main". I left some of them only... Apart from that the action looks good

moe-ad:
Thanks for the review @RobPasMue. I am pointing to the PR branch because of testing. I will revert before merging.

.github/workflows/ci_cd_release.yml
```diff
@@ -13,15 +13,21 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions: {}
+
```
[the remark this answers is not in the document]
moe-ad: Reason given above.

.github/workflows/ci_cd_night.yml
RobPasMue: Gotcha, thanks for the explanation!

.github/workflows/ci_cd_release.yml
moe-ad: You have demanded security and now you have it haha. Get used to it, we are moving to a new rule regime... :)
They are all some of the security issues zizmor flags.

Two vulnerabilities related to your question are the following:
1. Moving forward, we can't write to `GITHUB_ENV` or `GITHUB_PATH`, at least not advisable (see [github-env](https://woodruffw.github.io/zizmor/audits/#github-env) audit). Only `GITHUB_OUTPUT` is safe to write to. We only get to override with `# zizmor: ignore[ ... ]` in edge cases where we surely need to write to those files (e.g. in actions where we support projects using `poetry`).
2. Within **run** steps, `${{ some-expression }}` is no longer allowed because of the [template-injection](https://woodruffw.github.io/zizmor/audits/#template-injection) audit. 

So if you want to pass information between job steps, you need to write to GITHUB_OUTPUT and pass that info through `env` in all steps where it is needed, and expand directly through relevant shell syntax. For example on bash:

```YAML
    - name: "Set environment variables"
      id: set-environment-variables
      run: |
        echo "VAR_A=value_a" >> ${GITHUB_OUTPUT}
        echo "VAR_A=value_b" >> ${GITHUB_OUTPUT}

    - name: "Use VAR_A and VAR_B"
      id: use-two-variables
      env:
        VAR_A: ${{ steps.set-environment-variables.VAR_A }}
        VAR_B: ${{ steps.set-environment-variables.VAR_B }}
      run: |
        # Remove leading "v" from env.X
        echo "VAR_C=${VAR_B}_and_${VAR_B}" >> ${GITHUB_OUTPUT}

    - name: "Use ${{ steps.use-two-variables.outputs.VAR_C }} value"         # This is not within run
      if: steps.use-two-variables.outputs.VAR_C == 'value_a_and_value_b'     # This is also not within run
      env:
        VAR_C: ${{ steps.current-branch-tag-validity.outputs.MAJOR }}
      run: |
        echo "I am using VAR_C value of \'$VAR_C\' value to do something"
```

RobPasMue: Gotcha, thanks for the explanation!

build-wheelhouse/action.yml
```diff
@@ -132,7 +132,7 @@ runs:
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
SMoraisAnsys, line 132:
```suggestion
      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
      with:
        persist-credentials: false
```
Following your previous changes (which make sense), shouldn't we also be using `persist-credentials: false` for every composite actions where we use `actions/checkout` and don't perform extra git calls ?
If yes, you should track the other locations and apply the changes :)

check-actions-security/action.yml
```diff
@@ -0,0 +1,149 @@
...
+      shell: bash
+      run: |
+        uv tool install zizmor==1.6.0
+        zizmor --version
+
+    - name: "Optionally set GH_TOKEN"
+      if: ${{ inputs.token != '' }}
+      shell: bash
+      run: |  # zizmor: ignore[template-injection]
+        echo 'GH_TOKEN=${{ inputs.token }}' >> "$GITHUB_ENV"
```
SMoraisAnsys, line 84:
```suggestion
    - name: "Optionally set GH_TOKEN"
      if: ${{ inputs.token != '' }}
      shell: bash
      env:
        TOKEN: ${{ inputs.token }}
      run: |  # zizmor: ignore[github-env]
        echo "GH_TOKEN=${TOKEN}" >> ${GITHUB_ENV}
```
Wouldn't that get rid of he template injection ?

check-vulnerabilities/action.yml
```diff
@@ -343,16 +366,20 @@ runs:
...
         safety check -o bare --save-json info_safety.json --continue-on-error $ignored_vulnerabilities
-        bandit -r ${{ inputs.source-directory }} -o info_bandit.json -f json --exit-zero
+        bandit -r "${SOURCE_DIRECTORY}" -o info_bandit.json -f json --exit-zero
 
     - name: "Run safety advisory checks"
       shell: bash
+      env:
+        ACTIVATE_VENV: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
+        ACTION_PATH: ${{ github.action_path }}
+        HIDE_LOG: ${{ inputs.hide-log }}
```
SMoraisAnsys, line 376: Shouldn't `DEPENDENCY_CHECK_PACKAGE_NAME` be defined as `env` from a GITHUB output here, and removed from ${GITHUB_ENV} ?

check-vulnerabilities/action.yml
```diff
@@ -172,25 +172,34 @@ runs:
...
+        if [[ "${DEV_MODE}" == 'true' ]]; then
           echo "DEPENDENCY_CHECK_DRY_RUN=1" >> $GITHUB_ENV
           echo "DEPENDENCY_CHECK_ERROR_EXIT=1" >> $GITHUB_ENV
         fi
-        if [[ ${{ inputs.create-issues }} == 'true' ]];
-        then
+
+        if [[ "${CREATE_ISSUES}" == 'true' ]]; then
           echo "DEPENDENCY_CHECK_CREATE_ISSUES=1" >> $GITHUB_ENV
         fi
```
SMoraisAnsys, line 205: Shouldn't `DEPENDENCY_CHECK_REPOSITORY` be defined as `env` from a GITHUB output here, and removed from ${GITHUB_ENV} ? See below

SMoraisAnsys (changes requested):
Very great job, thanks for your hard work !
I left minor comments, let me know if some of them do not make sense :D

Have you had the time to test your branch using all of the modified actions ? They imply a lot of changes and we might end up breaking multiple projects workflows since we have to cover a large pannel of cases (runners / os / package manager / ...)
If not, could you test all of them in a dummy repo with Github runners + linux and windows + pip and poetry ? This should cover most of the cases and only leave edge cases to handle (if any bug was introduced)

build-wheelhouse/action.yml
moe-ad: The permission should be set in the workflow calling the composite action. This can still be at the workflow/job level, providing enough granularity for control. This was why I made changes in `./.github/workflows/*.yml` but not `./*/action.yml`. This is also the reason why `zizmor` doesn't flag permissions audit within composite actions.

check-actions-security/action.yml
moe-ad: Thanks for the suggestion!

check-vulnerabilities/action.yml
moe-ad: Nice catch. Along with `DEPENDENCY_CHECK_TOKEN` and `DEPENDENCY_CHECK_REPOSITORY`.

jorgepiloto:
Sounds good to me, @SMoraisAnsys.

check-vulnerabilities/action.yml
moe-ad: See my response above.

Infact, we can also get rid of the others i.e. `DEPENDENCY_CHECK_DRY_RUN`, `DEPENDENCY_CHECK_ERROR_EXIT`, and `DEPENDENCY_CHECK_CREATE_ISSUES`.

moe-ad:
@SMoraisAnsys Thanks for the review. I have addressed your suggestions.

build-wheelhouse/action.yml
SMoraisAnsys: Gotcha !

tests-pytest/action.yml
```diff
@@ -277,13 +299,25 @@ runs:
     - name: "Executing test suite"
       if: inputs.requires-xvfb == 'false'
       shell: bash
+      env:
+       PYTEST_MARKERS: ${{ inputs.pytest-markers }}
+       PYTEST_EXTRA_ARGS: ${{ inputs.pytest-extra-args }}
+       PYTEST_POST_ARGS: ${{ inputs.pytest-postargs }}
+       PYTEST_RUN: ${{ steps.output.pytest-run-command.PYTEST_RUN }}
```
jorgepiloto, line 306:
```suggestion
       PYTEST_RUN: ${{ steps.pytest-run-command.outputs.PYTEST_RUN }}
```

jorgepiloto (changes requested):
Found thanks to https://github.com/ansys/pyaedt/actions/runs/14661989340/job/41148728765?pr=6078

_doc-build-linux/action.yml
```diff
@@ -266,82 +273,101 @@ runs:
...
+        CHECK_LINKS: ${{ inputs.check-links }}
+        SPHINX_BUILD_MAKE: ${{ steps.environment-variables.outputs.SPHINX_BUILD_MAKE }}
       shell: bash
       run: |
         source .venv/bin/activate
-        ${{ env.SPHINX_BUILD_MAKE }} -C doc html SPHINXOPTS="${{ inputs.sphinxopts }}"
-        ${{ env.SPHINX_BUILD_MAKE }} -C doc pdf
-        if [[ ${{ inputs.check-links }} == 'true' ]];
+        ${SPHINX_BUILD_MAKE} -C doc html SPHINXOPTS="${SPHINX_OPTS}"
+        $SPHINX_BUILD_MAKE -C doc pdf
```
SMoraisAnsys, line 337:
```suggestion
        ${SPHINX_BUILD_MAKE} -C doc pdf
```

RobPasMue:
> @RobPasMue @jorgepiloto @germa89 What do you guys think about testing this PR in our repos to see how it behaves ? There are a lot of changes and if nothing out of the ordinary happens, we'll be able to merge this PR with greater peace of mind :) I'll try it in `pyaedt` to see what's happening there. Waiting for @moe-ad 's feedback on my review though

I can give it a try on my side too, np! But it might take me some time, probably not until Tuesday.

moe-ad:
@jorgepiloto, @SMoraisAnsys, @RobPasMue, @germa89.
Validated against:
- PyAEDT: ansys/pyaedt#6078
- PyAnsys Geometry: ansys/pyansys-geometry#1939
- PyMAPDL: ansys/pymapdl#3875

I guess we can merge?

RobPasMue:
Works for me

SMoraisAnsys (approved):

PR #734, opened 2025-03-17
refactor: use SHA version for pypa/gh-action-pypi-publish

SMoraisAnsys:
Use a SHA instead of a label for `pypa/gh-action-pypi-publish`.
On top of that the changes consist in adding `_release-pypi` to dependabot.
It seems that the expression `"!(requirements)/**/*"` does not cover the repo as it was assumed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,11 @@ updates:
   - package-ecosystem: "github-actions"
     directories:
       - "_release-pypi"
+      - "release-github"
+      # The following actions do not use non ansys-action but are added in case of future changes
+      - "release-pypi-private" 
+      - "release-pypi-public"
+      - "release-pypi-test"
     schedule:
       interval: "daily"
     commit-message:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,16 @@ updates:
 
   - package-ecosystem: "github-actions"
     directories:
+      - "_release-pypi"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "build"
+    labels:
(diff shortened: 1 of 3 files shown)


SMoraisAnsys:
I'll do my test in another repo

.github/dependabot.yml
```diff
@@ -10,6 +10,16 @@ updates:
     labels:
       - "dependencies:requirements"
 
+  - package-ecosystem: "github-actions"
+    directories:
+      - "_release-pypi"
```
SMoraisAnsys, line 15:
```suggestion
    directories:
      - "_release-pypi"
      - "release-github"
      # The following actions do not use non ansys-action but are added in case of future changes
      - "release-pypi-private" 
      - "release-pypi-public"
      - "release-pypi-test"
```

.github/dependabot.yml
```diff
@@ -10,6 +10,16 @@ updates:
...
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "build"
```
SMoraisAnsys, line 19:
```suggestion
      prefix: "build(release)"
```

RobPasMue (approved):
Let's give this a try!

.github/dependabot.yml
```diff
@@ -10,6 +10,21 @@ updates:
     labels:
       - "dependencies:requirements"
 
+  - package-ecosystem: "github-actions"
+    directories:
+      - "_release-pypi"
+      - "release-github"
+      # The following actions do not use non ansys-action but are added in case of future changes
+      - "release-pypi-private"
```
RobPasMue, line 18:
```suggestion
      - "release-pypi-private"
```

PR #738, opened 2025-03-19
fix: correct shell expansion for pip extra targets

moe-ad:
Extra targets for pip install should be comma separated.

### DIFF
--- a/check-vulnerabilities/action.yml
+++ b/check-vulnerabilities/action.yml
@@ -316,7 +316,7 @@ runs:
         if [[ ${{ env.BUILD_BACKEND }} == 'poetry' ]]; then
           poetry install "${extra_targets[@]}"
         else
-          python -m pip install ."${extra_targets[@]}"
+          python -m pip install ."$(echo ${extra_targets[@]} | sed 's/ /,/g')"
         fi
 
     - name: "Install action requirements"
--- a/doc/source/changelog/738.fixed.md
+++ b/doc/source/changelog/738.fixed.md
@@ -1,0 +1,1 @@
+correct shell expansion for pip extra targets


check-vulnerabilities/action.yml
```diff
@@ -316,7 +316,7 @@ runs:
         if [[ ${{ env.BUILD_BACKEND }} == 'poetry' ]]; then
           poetry install "${extra_targets[@]}"
         else
-          python -m pip install ."${extra_targets[@]}"
+          python -m pip install ."$(echo ${extra_targets[@]} | sed 's/ /,/g')"
```
RobPasMue, line 319: Well - now you are assuming that the targets are being passed as ``targetA targetB``
If you look at the description of the action:

```yml
  extra-targets:
    description: |
      Extra targets to be evaluated by safety. By default, it is set to an
      empty string. This gets substituted to the library install command as
      follows (inside <INPUT>)::
        # For pip install
        pip install .[<INPUT>]
        # For poetry install
        poetry install --extras '<INPUT>'
    default: ''
    required: false
    type: string
```

We were pretty explicit on what was going to be done. Meaning that the user should be aware that an in-place substitution was going to be made.

Bottom line - I would leave it as it was unless there is a major concern

RobPasMue (changes requested):

jorgepiloto:
Is this coming from Zizmor, @moe-ad?

check-vulnerabilities/action.yml
moe-ad: Thanks for providing more context @RobPasMue.

**My concern is:** Poetry accepts both comma-separated and space separated values (inside quotes), both `poetry install --extras "target1 target2"` and `poetry install --extras target1,target2` are valid. In a situation where a library using poetry via the comma-separated approach wants to switch to pip, it falls on the end user to be aware of these nuances and change how <INPUT> is also being passed. This may not immediately be intuitive. Although vice-versa is not a problem.

Don't you think abstracting all of these nuances in <INPUT> passing is beneficial? It can be something as simple as recommending that inputs should be comma separated in the description, since that will work in both cases. I can close the PR regardless, if you consider it best that end users should be familiar with all of these nuances already.

SMoraisAnsys: Agreeing with @RobPasMue

SMoraisAnsys: > Thanks for providing more context @RobPasMue.
> 
> **My concern is:** Poetry accepts both comma-separated and space separated values (inside quotes), both `poetry install --extras "target1 target2"` and `poetry install --extras target1,target2` are valid. In a situation where a library using poetry via the comma-separated approach wants to switch to pip, it falls on the end user to be aware of these nuances and change how is also being passed. This may not immediately be intuitive. Although vice-versa is not a problem.
> 
> Don't you think abstracting all of these nuances in passing is beneficial? It can be something as simple as recommending that inputs should be comma separated in the description, since that will work in both cases. I can close the PR regardless, if you consider it best that end users should be familiar with all of these nuances already.

I'm not expecting someone moving from poetry to pip. If such case happens, doesn't an error pops up ? Or did you end up in a case where no error was triggered and it was though to find the origin of the CI false positive ?

moe-ad:
welp 😆 
Couldn't convince anyone. Closing.
Thank you all for the inputs.

check-vulnerabilities/action.yml
RobPasMue: Yeah - I'm not expecting someone moving easily back from poetry to pip either... without them having to adapt a lot of things in the repo. And this would be one of the things they need to update.

PR #739, opened 2025-03-19
refactor: use full length commit SHA instead of tags for external github actions

SMoraisAnsys:
This PR consists in following the good practices and helps mitigating the risk of a bad actor adding a backdoor to the actions repository, see https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions?learn=getting_started#using-third-party-actions

The pin to a full length commit SHA is only performed for external github actions and not for our own actions, I assume we trust ourselves here :D

> [!NOTE]
Dependabot is working correctly with full length commit SHA used to reference a stable version, see the changes introduced in #734 and the following dependabot PRs #735 and #736.

> [!WARNING]
The doc-style action is being extend with a step that is never triggered. It is only here to let us know when a new release of `vale-action` is available so that we can update the branch we are leveraging (`reviewdog20`).

---

*Notes related to dependabot for future references*
1. Using `"/"` as `directory` input or part of `directories` input does not cover our composite actions. For the github actions ecosystem, using `"/"` seems to only track changes in `.github` directory.
2. Using a commit hash not associated with a release may conflict with the internal logic of dependabot. The work around is to add an extra step using a commit hash associated to a release so that dependabot can propose changes to that hash. However, **the added step must be defined before the step with a non commit hash reference !** Otherwise, dependabot will not even propose a change to the step with a commit hash associated to a release.


doc-style/action.yml
```diff
@@ -195,3 +197,13 @@ runs:
         filter_mode: nofilter
         vale_flags: "${{ ENV.VALE_FLAGS }}"
         version: ${{ inputs.vale-version }}
+
+    # Contrary to the previous step, this step refers to a release of vale-action.
```
jorgepiloto, line 201: Excelent! Thanks a lot, @SMoraisAnsys.

jorgepiloto (approved):
Thanks for this long yet necessary pull-request, @SMoraisAnsys.

doc-style/action.yml
```diff
@@ -185,7 +185,9 @@ runs:
         fi
 
     - name: "Run Vale"
-      uses: errata-ai/vale-action@reviewdog20
+      # The vale-action points to the latest commit in the reviewdog20 branch
+      # Contrary to other pinned version, this action is pinned to a commit
```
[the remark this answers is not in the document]
RobPasMue: Will dependabot also trigger the update even if it is pointing to a commit in a branch? Or will we be pinned on this commit "for ever" unless we update it?

doc-style/action.yml
RobPasMue: Aha... now I understand. So your goal is that dependabot pings you on this one... and then we remember that we have to modify it on line 190, correct?

If that's the case, could we automate it inside CI/CD? a simple script that detects changes to this specific line (the one that has a version comment) and applies the changes to all the vale action references in the file?

RobPasMue: I know it will involve some scripting work but in my experience... we will end up forgetting about it/missing it.

RobPasMue (commented):
Overall, this is great @SMoraisAnsys - just left one enhancement request

doc-style/action.yml
SMoraisAnsys: I'm not sure to understand why you mean by applying the changes to all the vale action references. The problem we have here is that we are leveraging the branch (`reviewdog20`) that isn't the one used to make new releases (`reviewdog`) of the `vale-action`.

I think it could back fire to automate changes in other places as we could end up using a release of `vale-action` that doesn't contain the features we want. Here, worst case scenario is that we miss a new release because someone merges dependabot PR without noticing the need to modify the correct call. Whereas, the other approach could lead us to use a version that isn't compatible with our inputs. As a reminder, we now leverage the `fail_level` input that isn't available in the latest release of vale (and won't be unless the author wants to bump the default version of reviewdog to leverage the new features).

A point to consider is that, for every new release of `vale-action`, we'll have to ask/do the associated changes in `reviewdog20`. Best case scenario a new major release is performed and we won't have to handle that but that wasn't planned a few months ago.

@RobPasMue Would reworking the comments around the version / step be enough to redirect/remind us to what is import here ?

doc-style/action.yml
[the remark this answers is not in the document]
SMoraisAnsys: I've got no answer to that since I've never tried it out before. I'll give it a shot on my dummy repo to see how it evolves. Maybe the current changes are overcomplicated and we could remove the extra step added to track stable releases.

doc-style/action.yml
RobPasMue: Ah yeah.. this rings a bell now. Sorry, I might have mixed things here.

Comments is good enough I guess under these circumstances... but I am worried that we will eventually miss/forget about this. Could we have the doc-style updates separate from the rest of the actions updates (i.e. on a group of its own)?

That will help us make sure that we don't miss it when mixed with other action updates.

SMoraisAnsys: Problem solved, finally dependabot was able to update a commit that is not release related. Do not ask me why, I was probably not waiting long enough before (note that I was receiving the pip updates but not the github-actions ones...)

SMoraisAnsys:
@RobPasMue I added new changes (and updated the PR description) since dependabot works with any hash commit:
- removal of the `vale-action` duplicate that is no longer needed
- update comment on the `vale-action` use since we should never update the hash from a dependabot PR unless jdkato decides to provide a release which allow us to use recent features of reviewdog>=20 🤞
- add "catch all" group to avoid missing added external actions

SMoraisAnsys:
FYI, the update of a commit not associated with a release seem to work with dependabot BUT it seems to require that the action must first be assigned to an hash associated to a release.

Here is what I was able to see:
- When I use `errata-ai/vale-action` with an hash associated to `v2.1.0` and then with the hash associated to latest commit in branch `reviewdog20` then I get a PR that proposes to upgrade both calls with the latest hash, see below
image
- When I only work with the the hash of `reviewdog20` then I don't get a PR that propose the previous change (or it might take much longer?)
- When I use the hash of `reviewdog20` and then the hash of `v2.1.0` I don't get a PR that propose the hash changes (or it also might take much longer?)

SMoraisAnsys:
TODO: I'll monitor my dummy repo and update this PR according to the results:
- either add the removed duplicate call to `vale-action` to ensure that we get notifications
- keep the duplicate call removed

jorgepiloto (approved):
Excellent!

.github/dependabot.yml
```diff
@@ -6,31 +6,61 @@ updates:
...
+      release-related-actions:
+        patterns:
+          - "pypa/gh-action-pypi-publish"
+          - "softprops/action-gh-release"
+      # Missed non ansys actions that should by assigned to a group
+      must-be-assigned-actions:
+        patterns:
+          - "*"
+        exclude-patterns:
+          - "ansys/actions*"
```
RobPasMue, line 66:
```suggestion
          - "ansys/actions*"

```
EOF line missing

RobPasMue (approved):
LGTM! Just one small comment - thanks @SMoraisAnsys for this impl.!

PR #742, opened 2025-03-21
fix: sanity check on inputs for release-github action

RobPasMue:
As title says. Experienced issues in https://github.com/ansys/pyansys-geometry/actions/runs/13991469373/job/39179171722


RobPasMue:
Will perform a patch release right after

clatapie (approved):

SMoraisAnsys (approved):

PR #744, opened 2025-03-21
fix: ``generate_release_notes`` should be false

RobPasMue:
Had to do a hotfix on release... this has been solved now. It was completely breaking all workflows. Should have realized that the defaults were incompatible


RobPasMue:
Not sure what's going on here...

PR #750, opened 2025-03-25
ci: changelog action should depend on rolling release

RobPasMue:
If the changelog action and the rolling release action run at the same time, you enter race conditions and sometimes, the version might not be available for download. I suggest we put it as a dependent stage.


jorgepiloto (approved):
Thanks!

PR #751, opened 2025-03-25
ci: refactor logic for PRs opened by dependabot

SMoraisAnsys:
The goal of this PR is to improve our security to supply chain attack.
Since any PR opened by dependabot could expose us to this kind of issue, as dependabot would trigger the CI, the changes in this PR aim at controling how we handle dependabot PRs.
In many cases, the vulnerabilities are found in a few days and fixed on the fly. By waiting a bit (6/7 days) after the dependabot's PR, we would reduce the probability to be exposed to a vulnerability.

The changes of this PR consist in:
- stopping the workflow if dependabot is the `triggering_actor`;
- stopping the workflow if pyansys-ci-bot is the `triggering_actor` in a dependabot PR;
- adding a manual check job (use of an environment) to ensure that any manual modification to the PR is validated by someone with the correct permissions;

>[!WARNING]
To leverage the environment feature and avoiding it to be too disruptive, I added an `if` to each job of the PR workflow. Without a check like `if: needs.labeler.result == 'success'` then the `changelog-fragment` job would not be triggered and be skipped. This is probably due to the status of the workflow after the skip of `check-dependabot-pr`.
I remember seeing github logs (with debug) and having the feeling that if a job had an if condition, e.g. `if: my_condition()` then the job was executed through the runner with the condition `success() && job_if_condition()`.
Runs https://github.com/ansys/actions/actions/runs/14057907900/job/39361834331 and https://github.com/ansys/actions/actions/runs/14061079426/job/39371781138 illustrate my point: the if condition was only added to `labeler`, the result is that `changelog-fragment` was skipped despite `labeler` success!

>[!NOTE]
A follow up of this PR would be to added could down for dependabot PRs but we're still waiting for https://github.com/dependabot/dependabot-core/issues/3651 to handle github-actions ecosystem.

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -26,17 +26,17 @@ jobs:
     steps:
 
      - name: "Checkout project"
-       uses: actions/checkout@v4
+       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
      - name: "Sync labels"
-       uses: micnncim/action-label-syncer@v1
+       uses: micnncim/action-label-syncer@3abd5ab72fda571e69fffd97bd4e0033dd5f495c # v1.3.0
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
      - name: "Label pull-request"
        # HACK: skip if contributor does not have write permissions (forks)
        if: github.event.pull_request.head.repo.full_name == github.repository
-       uses: actions/labeler@v5.0.0
+       uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9 # v5.0.0
        with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
 
@@ -103,7 +103,7 @@ jobs:
     steps:
 
     - name: "Install Git and clone project"
-      uses: actions/checkout@v4
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
     - name: "Set up Python"
       uses: ansys/actions/_setup-python@main
@@ -126,7 +126,7 @@ jobs:
         ls -R && python -m build && python -m twine check dist/*
 
     - name: "Upload distribution artifacts to GitHub artifacts"
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
       with:
         name: ${{ env.LIBRARY_NAME }}-artifacts
         path: ~/${{ env.LIBRARY_NAME }}/dist/
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Exit if dependabot triggered the workflow
         if: github.triggering_actor == 'dependabot[bot]'
         run: |
-          echo "::warning::Dependabot is not allowed to trigger this workflow. Please check the changes before running the workflow manually."
+          echo "::warning::Dependabot is not allowed to trigger this workflow. Please review carefully the changes before running the workflow manually."
           exit 1
 
   block-pyansys-ci-bot:
@@ -33,7 +33,7 @@ jobs:
       - name: Exit if dependabot triggered the workflow
         if: github.triggering_actor == 'pyansys-ci-bot' && startsWith(github.head_ref, 'dependabot')
         run: |
-          echo "::warning::PyAnsys CI bot is not allowed to trigger this workflow in dependabot's PR. Please check the changes before running the workflow manually."
+          echo "::warning::PyAnsys CI bot is not allowed to trigger this workflow in dependabot's PR. Please review carefully the changes before running the workflow manually."
           exit 1
 
   check-dependabot-pr:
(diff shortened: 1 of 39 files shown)


RobPasMue:
@SMoraisAnsys - you should use ``github.triggering_actor`` rather than ``github.actor``

Thus we don't need to update the PR, we just need to rerun the checks.

RobPasMue:
> the if condition was only added to labeler, the result is that changelog-fragment was skipped despite labeler success!

Why not consider using ``continue-on-error`` https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error? I don't know if it would be useful or not though... I get a bit lost with all the scenarios

SMoraisAnsys:
> > the if condition was only added to labeler, the result is that changelog-fragment was skipped despite labeler success!
> 
> Why not consider using `continue-on-error` https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error? I don't know if it would be useful or not though... I get a bit lost with all the scenarios

This wouldn't solve the problem. The main issue which required the `if: ...` conditions is that we are adding a job which shouldn't be triggered in some cases. When the PR is not created by dependabot, the job shouldn't be blocking. At the opposite, when the PR is created by dependabot, the job should be blocking and wait for the approval of with allowed access to the environment. Since we can access the result of a job that the current job depends on, that's what we leverage to ensure that all the required jobs are triggered when the blocking job is skipped. 

See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context

RobPasMue:
Got it - all clear!

RobPasMue (commented):
Best way is to merge it and give it a try I guess.

RobPasMue (approved):

SMoraisAnsys:
> Best way is to merge it and give it a try I guess.

Yup ^^ @ansys/pyansys-core  any feedback on that ?
@RobPasMue Note that we'll have to create the environment after the PR is merged. I've not created it yet to avoid confusion.

.github/workflows/ci_cd_pr.yml
```diff
@@ -17,8 +17,40 @@ concurrency:
 
 jobs:
 
+  block-dependabot:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Exit if dependabot triggered the workflow
+        if: github.triggering_actor == 'dependabot[bot]'
+        run: |
+          echo "::warning::Dependabot is not allowed to trigger this workflow. Please check the changes before running the workflow manually."
```
MaxJPRey, line 26:
```suggestion
          echo "::warning::Dependabot is not allowed to trigger this workflow. Please review carefully the changes before running the workflow manually."
```

.github/workflows/ci_cd_pr.yml
```diff
@@ -17,8 +17,40 @@ concurrency:
 
 jobs:
 
+  block-dependabot:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Exit if dependabot triggered the workflow
```
MaxJPRey, line 23: @SMoraisAnsys I would suggest to add a comment to explain why dependabot is not allowed to trigger the workflow for new colleagues.

.github/workflows/ci_cd_pr.yml
```diff
@@ -17,8 +17,40 @@ concurrency:
 
 jobs:
 
+  block-dependabot:
```
SMoraisAnsys, line 20:
```suggestion
  # NOTE: We do not allow dependabot to trigger the CI/CD pipeline automatically.
  # This is to mitigate supply chain attacks, where a malicious dependency update
  # could execute arbitrary code in our build environment.
  # Dependabot PRs must be reviewed carefully and approved manually before
  # running the CI.
  block-dependabot:
```

.github/workflows/ci_cd_pr.yml
```diff
@@ -17,8 +17,40 @@ concurrency:
...
+        if: github.triggering_actor == 'dependabot[bot]'
+        run: |
+          echo "::warning::Dependabot is not allowed to trigger this workflow. Please review carefully the changes before running the workflow manually."
+          exit 1
+
+  block-pyansys-ci-bot:
+    needs: block-dependabot
+    runs-on: ubuntu-latest
+    steps:
+      - name: Exit if dependabot triggered the workflow
```
SMoraisAnsys, line 33:
```suggestion
      - name: Exit if pyansys-ci-bot triggered the workflow on dependabot's PR
```

.github/workflows/ci_cd_pr.yml
```diff
@@ -17,8 +17,40 @@ concurrency:
...
+  block-dependabot:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Exit if dependabot triggered the workflow
+        if: github.triggering_actor == 'dependabot[bot]'
+        run: |
+          echo "::warning::Dependabot is not allowed to trigger this workflow. Please review carefully the changes before running the workflow manually."
+          exit 1
+
+  block-pyansys-ci-bot:
```
SMoraisAnsys, line 29:
```suggestion
  # NOTE: We do not allow pyansys-ci-bot to trigger the CI/CD pipeline automatically
  # on dependabot's PR. This is to mitigate supply chain attacks, where a malicious
  # dependency update could execute arbitrary code in our build environment.
  # Dependabot PRs must be reviewed carefully and approved manually before
  # running the CI.
  block-pyansys-ci-bot:
```

.github/workflows/ci_cd_pr.yml
SMoraisAnsys: Handled in f3826e4e6a1c8028cc8881b1a47b425fa26999e3

RobPasMue:
Can we get this merged and start trying it?

PR #752, opened 2025-03-25
ci: add dependabot cooldown for pip

SMoraisAnsys:
Since we discovered that there is a beta feature for adding cooldown to pip, why not using it ? :D


MaxJPRey (approved):
LGTM.

PR #754, opened 2025-04-01
feat: uv as default package manager

### LINKED ISSUE #693: Set uv as the default package manager
[uv] is a package manager written in Rust that has proven to be really fast and versatile. Considering how many projects we have, the amount of minutes employed by CI/CD is large. Using [uv] could significantly reduce CI/CD times at the end of the month.

To implement [uv] in our actions, we must:

- [ ] Collect data on the amount of minutes used by CI/CD at org level before implementing this
- [ ] Replace any `python -m pip install` by the `uv pip install`
- [ ] Ensure any compatibility with Poetry and other build systems
- [ ] Collect data on the amount of minutes used by CI/CD at org level after implementing this

[uv]: https://github.com/astral-sh/uv

moe-ad:
Closes #693

The idea is to make `uv` the default package manager in actions where it can bring much value — those where potentially a lot of packages will be installed.

So far these include:
- Actions where a project's library and/or its dependencies can be installed:
  - `_doc-build-linux`
  - `_doc-build-windows`
  - `build-wheel-house`
  - `check-licenses`
  - `check-vulnerabilities`
  - `code-style`
  - `tests-pytest`

- Actions where > two packages are installed:
  - `doc-deploy-changelog`

In actions where ≤ 2 packages are being installed, the idea is that `pip` is enough and I didn't touch those. They include:
- `_release-pypi`
- `build-library`
- `doc-changelog`
- `hk-package-clean-except`
- `hk-package-clean-untagged`

Implementing `uv` support has been done in a non-intrusive way, to accommodate those that do not want to use it. In the actions where `uv` is supported, a `use-uv` input has been added which defaults to true. Therefore, setting `use-uv=false` essentially reverts those actions to how they were before.


moe-ad:
For `actions/build-wheelhouse`, there is no direct way to include dependency wheels if `uv` is used for wheel generation (see astral-sh/uv#1681). So I have limited `uv` usage to just package installation operations prior to the wheel building step.

moe-ad:
Changed base branch to `feat/check-actions-security-action`. Will revert to `main` when it's time to merge.

_setup-python/action.yml
jorgepiloto, line 1: Shouldn't we install Python via `uv` rather than installing both side by side?

jorgepiloto (changes requested):

_setup-python/action.yml
moe-ad: If we do this, that would mean projects won't have a way of not using `uv` when using affected actions. My understanding is that we will still allow this flexibility.

**Note** that `uv` by itself doesn't bundle any python versions, they are downloaded as needed if they aren't already available (or when forced to do so via some CLI arguments). `uv` is still able to do it's job with python installations not provisioned through it. Same goes for `astral/setup-uv`, it installs just the `uv` executable on the runner if additional options relating to how python should be set up aren't passed.

With all of this in mind, we can set things up in a non-intrusive way and this is what I have done. Users only need to set `use-uv: false` in affected actions and workflows becomes essentially the same as they were before. Win-win, if I am allowed to call it that ^^

_doc-build-linux/action.yml
```diff
@@ -254,13 +266,15 @@ runs:
...
         python -m pipx install poetry
-        python -m venv .venv
       env:
         PIPX_BIN_DIR: ${{ runner.temp }}/pipx/bin
         PIPX_HOME : ${{ runner.temp }}/pipx/home
 
+    # NOTE: A virtual environment is needed regardless in all cases. In uv's case,
+    # the installation speed "uv pip" offers is what needs to be taken advantage of,
+    # using "uv venv" in a separate step is not necessary since "uv pip" will still
+    # work with existing virtual environments.
```
jorgepiloto, line 276: Even when created using the `python -m venv .venv`?

jorgepiloto (commented):
One last comment before approving, @moe-ad.

_doc-build-linux/action.yml
moe-ad: Yes. That's correct.

moe-ad:
@jorgepiloto
One thing we should also decide on are the actions I decided aren't worth touching. Is that approach fine by you? The actions I touched so far are the ones where a lot of dependencies are being installed. I skipped actions where just 1 or 2 packages are being installed because I thought that would not offer much value.

I guess the question becomes if we want all our actions to have `use-uv` option for the sake of consistency across board, or if we want only the actions that can benefit from `uv` to have it as I've already done.

jorgepiloto:
Thanks for pointing this out @moe-ad. I would provide the option to use `uv` across all actions. The reason is that, even if we are installing a single package, we are not aware about the number of dependencies it has.

jorgepiloto (approved):
After a quick discussion, the install of some packages used during the CI/CD for certain actions do not have a long number of dependencies. Therefore, I am approving on my side.

moe-ad:
Pinging @ansys/pyansys-core for their inputs as well. Especially regarding this:

> @jorgepiloto One thing we should also decide on are the actions I decided aren't worth touching. Is that approach fine by you? The actions I touched so far are the ones where a lot of dependencies are being installed. I skipped actions where just 1 or 2 packages are being installed because I thought that would not offer much value.
> 
> I guess the question becomes if we want all our actions to have `use-uv` option for the sake of consistency across board, or if we want only the actions that can benefit from `uv` to have it as I've already done.

_doc-build-linux/action.yml
```diff
@@ -254,13 +266,15 @@ runs:
...
     - name: "Create a virtual environment"
-      if: steps.environment-variables.outputs.BUILD_BACKEND == 'pip'
```
SMoraisAnsys, line 263: Unfortunately, I think that the behavior we want to have with poetry requires that `PIPX_BIN_DIR` and `PIPX_HOME` are defined when creating the virtual environment. I think it was an edge case when working with poetry + self hosted runners where caching behavior had an impact. I'd recommend adding an OR statement in the if condition. Another option would be to add the environment variable by default and hope that adding them doesn't add any side effect (I don't like that approach though)

_doc-build-windows/action.yml
```diff
@@ -278,23 +292,27 @@ runs:
...
+      env:
+        USE_UV: ${{ inputs.use-uv }}
       run: |
         .venv\Scripts\Activate.ps1
-        python -m pip install -U pip
+        if ($env:USE_UV -eq "true") {
+            uv pip install -U pip
+        } else {
+            python -m pip install -U pip
+        }
```
SMoraisAnsys, line 315:
```suggestion
        if ($env:USE_UV -eq "true") {
          uv pip install -U pip
        } else {
          python -m pip install -U pip
        }
```
To match files formatting

_doc-build-windows/action.yml
```diff
@@ -308,10 +326,15 @@ runs:
...
+        USE_UV: ${{ inputs.use-uv }}
       if: ${{ steps.check-requirements-file-exists.outputs.EXISTS_DOC_REQUIREMENTS == 'true' }}
       run: |
         .venv\Scripts\Activate.ps1
-        python -m pip install -r "$env:REQUIREMENTS_FILE"
+        if ($env:USE_UV -eq "true") {
+            uv pip install -r "$env:REQUIREMENTS_FILE"
+        } else {
+            python -m pip install -r "$env:REQUIREMENTS_FILE"
+        }
```
SMoraisAnsys, line 337:
```suggestion
        if ($env:USE_UV -eq "true") {
          uv pip install -r "$env:REQUIREMENTS_FILE"
        } else {
          python -m pip install -r "$env:REQUIREMENTS_FILE"
        }
```
To match files formatting

_doc-build-windows/action.yml
```diff
@@ -320,10 +343,12 @@ runs:
         BUILD_BACKEND: ${{ steps.determine-environment-variables.outputs.BUILD_BACKEND }}
       run: |
         .venv\Scripts\Activate.ps1
-        if ("$env:BUILD_BACKEND" -eq 'poetry' ) {
-          poetry install
+        if ($env:BUILD_BACKEND -eq "poetry") {
+            poetry install
+        } elseif ($env:BUILD_BACKEND -eq "uv") {
+            uv pip install .
```
SMoraisAnsys, line 349:
```suggestion
        if ($env:BUILD_BACKEND -eq "poetry") {
          poetry install
        } elseif ($env:BUILD_BACKEND -eq "uv") {
          uv pip install .
```
To match files formatting

_doc-build-windows/action.yml
```diff
@@ -320,10 +343,12 @@ runs:
...
         } else {
-          python -m pip install .
+            python -m pip install .
```
SMoraisAnsys, line 351:
```suggestion
          python -m pip install .
```

_doc-build-windows/action.yml
```diff
@@ -333,10 +358,12 @@ runs:
         BUILD_BACKEND: ${{ steps.determine-environment-variables.outputs.BUILD_BACKEND }}
       run: |
         .venv\Scripts\Activate.ps1
-        if ("$env:BUILD_BACKEND" -eq 'poetry' ) {
-          poetry install --with doc
+        if ($env:BUILD_BACKEND -eq "poetry") {
+            poetry install --with doc
+        } elseif ($env:BUILD_BACKEND -eq "uv") {
+            uv pip install .[doc]
```
SMoraisAnsys, line 364:
```suggestion
        if ($env:BUILD_BACKEND -eq "poetry") {
          poetry install --with doc
        } elseif ($env:BUILD_BACKEND -eq "uv") {
          uv pip install .[doc]
```

_doc-build-windows/action.yml
```diff
@@ -333,10 +358,12 @@ runs:
...
         } else {
-          python -m pip install .[doc]
+            python -m pip install .[doc]
```
SMoraisAnsys, line 366:
```suggestion
          python -m pip install .[doc]
```

_setup-python/action.yml
```diff
@@ -32,6 +32,11 @@ inputs:
     description: 'Whether to use Python cache or not. Requires a pyproject.toml file.'
     required: true
     type: boolean
+  provision-uv:
```
SMoraisAnsys, line 35: Very good idea

build-wheelhouse/action.yml
```diff
@@ -136,17 +144,22 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
+        provision-uv: ${{ inputs.use-uv == 'true' && 'true' || 'false' }}
```
SMoraisAnsys, line 147: I'm not sure about this line. Either inputs.use-uv is perceived as boolean and thus ${{ inputs.use-uv }} should be enough or it's perceived as a string and then I also don't see why we would need `== 'true' && 'true' || 'false'`.
Am I missing something or can this be updated ? :)

check-vulnerabilities/action.yml
```diff
@@ -233,6 +245,7 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
         use-cache: false
+        provision-uv: ${{ inputs.use-uv == 'true' && 'true' || 'false' }}
```
SMoraisAnsys, line 248: Same comment

check-vulnerabilities/action.yml
```diff
@@ -262,13 +275,15 @@ runs:
...
       env:
         PIPX_BIN_DIR: ${{ runner.temp }}/pipx/bin
         PIPX_HOME : ${{ runner.temp }}/pipx/home
 
+    # NOTE: A virtual environment is needed regardless in all cases. In uv's case,
+    # the installation speed "uv pip" offers is what needs to be taken advantage of,
+    # using "uv venv" in a separate step is not necessary since "uv pip" will still
+    # work with existing virtual environments.
     - name: "Create a virtual environment"
-      if: steps.github-environment-variables.outputs.BUILD_BACKEND == 'pip'
```
SMoraisAnsys, line 271: Again

code-style/action.yml
```diff
@@ -147,6 +159,7 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
+        provision-uv: ${{ inputs.use-uv == 'true' && 'true' || 'false' }}
```
SMoraisAnsys, line 162: I'm stopping to report echos

SMoraisAnsys (changes requested):
I left some comments. The comment on venv creation needs to be addressed as it could introduce unexpected behaviors for users.

SMoraisAnsys:
Btw, why not adding the possibility to use `uv` everywhere and only activate it for the actions that you already have modified ? This way, actions with less than 2 dependencies will keep using the old behavior and only actions where it makes sense are affected. On top of that, that will give us freedom later on if we want to move one action from "no uv" to "uv by default".

Note that I'm fine with not adding the possibility to use `uv` everywhere. I just think that's a nice occasion to allow it for later (if ever needed) :)

_doc-build-linux/action.yml
moe-ad: @SMoraisAnsys thanks for noting this. Maybe I am missing something, but the current implementation takes this into account. It hasn't changed how `poetry` was being handled previously.

Before:
- step-A (`PIPX_BIN_DIR` and `PIPX_HOME` set):
  - if `BUILD_BACKEND == poetry`:
    - ... some actions
    - create `venv`
- step-B:
  - if `BUILD_BACKEND == pip`:
    - create `venv`
  
A `venv` is still created in both cases and in this PR, I have just unified the creation of the `venv` such that adding `uv` to the mix becomes easier to handle:
- The virtual environment creation step is always executed.
- When the build backend is poetry, `PIPX_BIN_DIR` and `PIPX_HOME` are still set before the virtual environment creation step.

So when the build backend is poetry, the way the action works is essentially the same as before.

_doc-build-windows/action.yml
moe-ad: Thanks for catching this and subsequent ones @SMoraisAnsys!

moe-ad:
> Btw, why not adding the possibility to use `uv` everywhere and only activate it for the actions that you already have modified ? This way, actions with less than 2 dependencies will keep using the old behavior and only actions where it makes sense are affected. On top of that, that will give us freedom later on if we want to move one action from "no uv" to "uv by default".
> 
> Note that I'm fine with not adding the possibility to use `uv` everywhere. I just think that's a nice occasion to allow it for later (if ever needed) :)

Good point. We can easily extend to other actions as well like you suggested. Jorge also thought of doing it this way but we decided to get all of your opinions regarding it.
Tagging @jorgepiloto for visibility.

build-wheelhouse/action.yml
moe-ad: Great catch! That's correct. We shouldn't need to do this.

moe-ad: Fixed!

_doc-build-linux/action.yml
SMoraisAnsys: Maybe I'm missing something but, AFAIK, the current step creating the .venv does not contain the `PIPX_BIN_DIR` and `PIPX_HOME` environment variable since the step does not contain any `env:`

Anyway, after some though I think the environment variables are mainly here to ensure that poetry does not contain much cached packages. Without the use of pipx, we wouldn't be sure that poetry gets installed in it's own isolated venv.

I'll leave that discussion unsolved so that you can read my answer. Feel free to resolve it afterward :)

moe-ad: Sorry, I mixed things up. Apologies for the confusion. I was talking about this part:

```
# NOTE: Installation of poetry in a separate environment to the project to
# avoid situations in which both poetry and the project have shared
# dependencies with different version. This can lead to CICD failures. For
# more information, see https://github.com/ansys/actions/pull/560
- name: "Add pipx/bin directory to Github Path"
  if: steps.environment-variables.outputs.BUILD_BACKEND == 'poetry'
  shell: bash
  run: echo "${{ runner.temp }}/pipx/bin" >> $GITHUB_PATH  # zizmor: ignore[github-env]

# NOTE: Poetry uses virtual environments when installing a project. As we
# want to control that creation, we store POETRY_VIRTUALENVS_CREATE=false
# in the GitHub environment.
- name: "Set poetry environment variable(s)"
   if: steps.environment-variables.outputs.BUILD_BACKEND == 'poetry'
   shell: bash
   run: echo "POETRY_VIRTUALENVS_CREATE=false" >> $GITHUB_ENV # zizmor: ignore[github-env]
```

For whatever reason, my brain registered that we were writing `PIPX_BIN_DIR` and `PIPX_HOME` to `GITHUB_ENV`.
😄

moe-ad:
> Btw, why not adding the possibility to use `uv` everywhere and only activate it for the actions that you already have modified ? This way, actions with less than 2 dependencies will keep using the old behavior and only actions where it makes sense are affected. On top of that, that will give us freedom later on if we want to move one action from "no uv" to "uv by default".
> 
> Note that I'm fine with not adding the possibility to use `uv` everywhere. I just think that's a nice occasion to allow it for later (if ever needed) :)

This is implemented now.

_release-pypi/action.yml
```diff
@@ -96,6 +96,13 @@ inputs:
     required: true
     type: boolean
 
+  use-uv:
+    description: |
+      Whether to use uv as the default package manager instead of pip. Default value is ``true``.
+    default: true
```
SMoraisAnsys, line 102:
```suggestion
    default: false
```
I would default to `false` for the actions were you didn't implement `uv` initially:
- _release-pypi
- build-library
- doc-changelog
- hk-package-clean-except
- hk-package-clean-untagged

@jorgepiloto @RobPasMue What do you think ?

If this is validated, the current change needs to be applied in other actions

jorgepiloto: Thanks for raising this, @SMoraisAnsys. Agreed. If we proceed implementing `uv` everywhere, then at least let's disable it for "critical" actions.

RobPasMue: I completely agree to , thanks @SMoraisAnsys and @moe-ad!

moe-ad: @SMoraisAnsys @jorgepiloto @RobPasMue 
Implemented. `ansys/actions/build-wheelhouse` was also added.

doc-changelog/action.yml
```diff
@@ -105,6 +105,13 @@ inputs:
     default: true
     type: boolean
 
+  use-uv:
+    description: |
+      Whether to use uv as the default package manager instead of pip. Default value is ``true``.
+    default: false
```
SMoraisAnsys, line 111: Sry but why default to false here when it defaults to true in doc-deploy-changelog ? The behavior of both actions w.r.t package management seems to be the same, no ?

release-github/action.yml
```diff
@@ -144,6 +144,13 @@ inputs:
     required: false
     type: boolean
 
+  use-uv:
+    description: |
+      Whether to use uv as the default package manager instead of pip. Default value is ``true``.
+    default: true
```
SMoraisAnsys, line 150: Wondering if we shouldn't move to false here since only 2 dependencies seem to be installed.

_release-pypi/action.yml
SMoraisAnsys: > Thanks for raising this, @SMoraisAnsys. Agreed. If we proceed implementing `uv` everywhere, then at least let's disable it for "critical" actions.

Did you mean "non critical" actions ? My idea was to follow your thoughs on having a treshold to default to `uv` or not. Like "if the number of packages installed is >2" then "true" else "false".

build-wheelhouse/action.yml
```diff
@@ -123,6 +123,14 @@ inputs:
     default: false
     type: boolean
 
+  use-uv:
+    description: |
+      Whether to use uv for installation operations instead of pip. By default it is set
+      to ``true``.
```
MaxJPRey, line 129: Just to be consistent with the other descriptions.
```suggestion
      Whether to use uv as the default package manager instead of pip. Default value is ``true``.
```

code-style/action.yml
```diff
@@ -103,6 +103,14 @@ inputs:
     default: true
     type: boolean
 
+  use-uv:
+    description: |
+      Whether to use uv for installation operations instead of pip. By default it is set
+      to ``true``.
```
MaxJPRey, line 109:
```suggestion
      Whether to use uv as the default package manager instead of pip. Default value is ``true``.
```

MaxJPRey (approved):
Excellent work @moe-ad .

doc-changelog/action.yml
moe-ad: Thanks @SMoraisAnsys, I missed this. I implemented changes in-line with the original request, which is to default to `false` for actions I didn't implement `use-uv` for previously.

So what's the final verdict here? Default to `false` for both or default to `true`?

release-github/action.yml
moe-ad: That makes sense.

build-wheelhouse/action.yml
moe-ad: Thanks for catching that @MaxJPRey

release-github/action.yml
MaxJPRey: Good idea.

_release-pypi/action.yml
jorgepiloto: Oh! I thought we were discretizing based on actions that build, release and deploy "uv: false" and the rest "uv: true".

doc-changelog/action.yml
jorgepiloto: We all agreed to use `uv` everywhere by default.

moe-ad:
Latest changes tested on a release workflow: https://github.com/ansys-internal/test-repo-moe/actions/runs/14926303459

PR #755, opened 2025-04-02
docs: do not check link on www.x.org/*

SMoraisAnsys:
Every two weeks the servers associated to www.x.org are down and our CICD fails because we do have a reference to https://www.x.org/releases/X11R7.6/doc/man/man1/Xvfb.1.xhtml. This ~fixes~ ignores it.


RobPasMue (approved):
I mean... it doesn't "fix" it but it ignores it 😄 sounds good to me.

RobPasMue:
All good - let's merge!

PR #756, opened 2025-04-02
fix: drop build and wheel packages

moe-ad:
We use [pip wheel](https://github.com/ansys/actions/blob/af045ca0e768847401c655294e48fb9981f5dcb1/build-wheelhouse/action.yml#L191) for building wheels in `actions/build-wheelhouse`. `build` and `wheel` packages are not needed dependencies to be able to use `pip wheel` because "wheel" is a command option for pip in this case.

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -130,10 +130,10 @@ runs:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
 
-    - name: "Update pip and install the build and wheel libraries"
+    - name: "Update pip"
       shell: bash
       run: |
-        python -m pip install --upgrade pip build wheel
+        python -m pip install --upgrade pip
 
     - name: "Check if specific target is requested"
       shell: bash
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):
LGTM!

SMoraisAnsys (approved):
LGTM

RobPasMue:
Aha... so third parties are not shipping the wheels and you need to build them on the fly. I'd suggest we revert it @moe-ad @SMoraisAnsys - this is a use case in which it can come in handy to have wheel and build defined. We can also expose it as an optional argument/input if you want -- like, ``install-wheel-build: false``. But I doubt SimBA would be the only one affected... I would just revert it IMO.

moe-ad:
@anscfrisson thanks for bringing this to our attention. This will surely break things for your use case truly and we will expose it as an optional argument like @RobPasMue has suggested

One weird thing remains though, we haven't created a patch release with changes from this PR yet. The version of `ansys/actions/buildwheelhouse` you are using should still be installing build and wheel.

anscfrisson:
> > One weird thing remains though, we haven't created a patch release with changes from this PR yet. The version of `ansys/actions/buildwheelhouse` you are using should still be installing build and wheel.
> 
> @moe-ad I agree with you. I am investigating.

@moe-ad @RobPasMue I have found the fix to my initial problem: https://github.com/ansys/actions/pull/761

PR #758, opened 2025-04-03
refactor!: remove trusted publisher

SMoraisAnsys:
Remove trusted publisher input from our PyPI release actions.
Add documentation on that on the migration file (also adding the wheel-house breaking change).

### DIFF
--- a/_release-pypi/action.yml
+++ b/_release-pypi/action.yml
@@ -81,30 +81,9 @@ inputs:
     required: true
     type: boolean
 
-  use-trusted-publisher:
-    description: |
-      Whether to use the OIDC token for releasing.
-      This is useful when you want to publish to PyPI through a CI/CD pipeline
-      as a Trusted Publisher. It is necessary that your repository has been authorized
-      to use Trusted Publisher's. For more information, see
-      `Using a Trusted Publisher <https://docs.pypi.org/trusted-publishers/>`_.
-
-      .. note::
-
-          ``permissions: id-token: write`` is mandatory for trusted publishing.
-
-    required: true
-    type: boolean
-
 runs:
   using: "composite"
   steps:
-
-    - name: "Notify to use trusted publisher"
-      shell: ${{ runner.os == 'Windows' && 'powershell' || 'bash' }}
-      if: inputs.use-trusted-publisher == 'false'
-      run: |
-        echo "::notice::We recommend using trusted publishers to securely publish packages on PyPI. You should consider moving to this approach."
 
     - name: "Set up Python"
       uses: ansys/actions/_setup-python@main
@@ -138,7 +117,7 @@ runs:
 
     - name: "Upload artifacts to PyPi"
       shell: bash
-      if: inputs.dry-run == 'false' && inputs.use-trusted-publisher == 'false'
+      if: inputs.dry-run == 'false'
       run: |
         python -m twine upload --verbose ${{ env.SKIP_EXISTING }} ${{ inputs.library-name }}-artifacts/*.whl
         python -m twine upload --verbose ${{ env.SKIP_EXISTING }} ${{ inputs.library-name }}-artifacts/*.tar.gz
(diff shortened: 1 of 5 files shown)


_release-pypi/action.yml
```diff
@@ -81,31 +81,10 @@ inputs:
...
-      as a Trusted Publisher. It is necessary that your repository has been authorized
-      to use Trusted Publisher's. For more information, see
-      `Using a Trusted Publisher <https://docs.pypi.org/trusted-publishers/>`_.
-
-      .. note::
-
-          ``permissions: id-token: write`` is mandatory for trusted publishing.
-
-    required: true
-    type: boolean
```
RobPasMue, line 97: I wouldn't remove it yet, leave the input and exit on error if it is set to true. That way people know how to adapt (at least in v9). I'm sure many of them haven't even seen the notice we were throwing on v8. We can remove the input on v10

release-pypi-private/action.yml
```diff
@@ -63,22 +63,6 @@ inputs:
     default: ''
     type: string
 
-  use-trusted-publisher:
```
RobPasMue, line 66: Same for all of these

doc/source/migrations/index.rst
```diff
@@ -7,6 +7,82 @@ This guide provides information on new features, breaking changes, how to migrat
 from one version of the actions to another, and other upstream dependencies that
 have been updated.
 
+Version ``v9.0``
+----------------
+
+**Breaking changes:**
+
+- Use ``ansys/actions/build-wheelhouse`` or ``ansys/actions/check-licenses`` actions with Python version 3.10 or higher.
```
RobPasMue, line 15: Shouldn't we implement a check inside these actions to verify that Python 3.10 or above is being passed as input?

RobPasMue (changes requested):
Some comments

_release-pypi/action.yml
SMoraisAnsys: Very solid point, I was in a talk with @jorgepiloto  and he pointed that providing an input that doesn't exist won't break the action. So it's even more logic to keep logging an error and exiting the workflow.
Thanks both of you for the feedback !

doc/source/migrations/index.rst
SMoraisAnsys: From the implementation of https://github.com/ansys/actions/pull/698, the installation process will fail since the package isn't 3.10 compatible. Following your approach, we could make the workflow break with a clear message though. I'll do that :)

RobPasMue: Thanks!

_release-pypi/action.yml
```diff
@@ -100,6 +100,20 @@ runs:
...
+        message: >
+          As the Python ecosystem is moving to project metadata version 2.4,
+          we need to use a more recent version of pypa/gh-action-pypi-publish.
+          However, this action is not compatible with composite actions.
+          Therefore, this private action can no longer be used with trusted
+          publishers. To keep using trusted publisher, and we encourage that,
+          we invite all users to define their release job manually.
+          As an example, see
+          https://actions.docs.ansys.com/version/stable/release-actions/index.html#release-pypi-template
+
```
RobPasMue, line 116: Hopefully we should never arrive here since it is part of the private action. I think that as long as we keep it in the public facing ones, that's fine.

MaxJPRey (approved):
LGTM.

_release-pypi/action.yml
SMoraisAnsys: I added it here following your previous comment but I also though about not adding it here. I checked that no one is using the private action and that's the case so I'll remove it from the private action :)

RobPasMue: Reopening the thread -- on the private action I would also remove the input statement. Let's just leave it on the public facing ones. Thanks for handling that @SMoraisAnsys.

Yeah I might have placed my comment in the wrong file, sorry for that. In my mind I meant it to only be present in the public facing ones.

RobPasMue (approved):
LGTM!

RobPasMue (approved):
Looks good to go, thanks a lot @SMoraisAnsys 🚀

PR #769, opened 2025-04-09
docs: fix link

RobPasMue:
As title says. We should do a patch release so that the docs get updated on the released version too.

PR #771, opened 2025-04-09
docs: fixing links rendering in migration guide

clatapie:
As title says


clatapie:
Closing because duplicated of #769

PR #774, opened 2025-04-09
ci: rewrite dependabot checks

RobPasMue:
Coming from experience in #770

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -55,7 +55,7 @@ jobs:
   labeler:
     name: "Labels"
     needs: check-dependabot-pr
-    if: ${{ success() }}
+    if: needs.check-dependabot-pr.result == 'skipped' || needs.check-dependabot-pr.result == 'success'
     permissions:
       contents: read
       pull-requests: write
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -38,7 +38,7 @@ jobs:
             echo "environment=dependabot" >> "$GITHUB_OUTPUT"
             echo "::warning::PyAnsys CI bot is not allowed to trigger this workflow in dependabot's PR. Please review carefully the changes before running the workflow manually."
           else
-            echo "environment=" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


SMoraisDev (approved):
Reviewing while students are working :p
I like the changes you propose but I'm not sure to understand what went wrong before. I had performed testing on a dummy repo and everything was working as expected (I might have missed something though but I don't know why).
I had though about using `outputs` too but I had in mind that we wanted to avoid using them. If that's fine with you, I think your refactoring is better than the previous approach.

RobPasMue:
> Reviewing while students are working 📦 

You naughty teacher... I hope they never find your secondary GitHub username 😄

RobPasMue:
> I like the changes you propose but I'm not sure to understand what went wrong before. I had performed testing on a dummy repo and everything was working as expected (I might have missed something though but I don't know why). I had though about using `outputs` too but I had in mind that we wanted to avoid using them. If that's fine with you, I think your refactoring is better than the previous approach.

Yeah I'm not sure what went wrong... I just outlined the behavior. I think it's more bulletproof this way though. Thanks for agreeing on it! I'll have it merged once @ansys/pyansys-core reviews and approves. And I'll give it a try to rebasing #770 again

.github/workflows/ci_cd_pr.yml
```diff
@@ -1,7 +1,6 @@
 name: pull-request
 on:
   pull_request:
-    types: [opened, synchronize, reopened, labeled]
```
SMoraisAnsys, line 4: Wondering if this shouldn't this be changed to the previous value ?
image

SMoraisAnsys (commented):
With your changes, I'm wondering if we still need to keep tweaking the if statement as we did in the previous PR. I don't see the value of keeping the "skipped" check. Did you give it a shot without ? Also, we might no longer need the `success() ||`

RobPasMue:
Gave it a try to removing the conditionals... but not working sady

RobPasMue:
We decided to move to an "empty" environment approach if the triggering actor is not ``dependabot`` or the ``pyansys-ci-bot`` accounts. Otherwise, the ``dependabot`` environment type will be activated and it will request manual approval.

Reverting the types as well and adding ``edited`` conditions as well for triggering it.

PR #777, opened 2025-04-14
build(actions): update action-gh-release

SMoraisAnsys:
When using the full hash for actions commit I left the old version that was used hoping for an automatic update from dependabot.
However, it seems that the action repo does not always handle new versions as one would expect (e.g. there is no 1.0 version). I checked and the only "big change" was that they moved to `node20` so I'm not expecting any issue with bumping to a newer version.

### DIFF
--- a/doc/source/changelog/777.dependencies.md
+++ b/doc/source/changelog/777.dependencies.md
@@ -1,0 +1,1 @@
+update action-gh-release
--- a/release-github/action.yml
+++ b/release-github/action.yml
@@ -348,10 +348,7 @@ runs:
         } >> "$GITHUB_ENV"
 
     - name: "Release to GitHub"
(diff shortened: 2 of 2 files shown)


SMoraisAnsys:
@RobPasMue Rofl, seems like this is an edge case xD

From the logs:  `  elif [[ "pyansys-ci-bot" == "pyansys-ci-bot" && false ]]; then` but we still go through...

SMoraisAnsys:
> @RobPasMue Rofl, seems like this is an edge case xD
> 
> From the logs: ` elif [[ "pyansys-ci-bot" == "pyansys-ci-bot" && false ]]; then` but we still go through...

Found the issue I think, we are note testing that `false == true` and just using `false` as an executable... I'll patch my PR

RobPasMue (approved):

RobPasMue:
Thanks @SMoraisAnsys!

PR #776, opened 2025-04-10
feat: check environment approval

SMoraisAnsys:
Following recent developments to add a manual approval for dependabot PR (see #751 and #774), this PR gathers the decision logic of requiring a manual approval or not.
The action allows user to specify which environment they want to use when a workflow is running because either `dependabot[bot]` triggered it or the workflow is associated to a PR of `dependabot[bot]` and `pyansys-ci-bot` is the one triggering the workflow.  Another input is allowed to provide a "skip" environment for other cases.

> [!NOTE]
Currently, passing `skip-manual-check-environment: ''` behaves as if no environment was defined. Therefore, no comment such as the one below appear in the PR conversation.
image
That was tested in a1d92c8, see below
image

PR #778, opened 2025-04-17
fix: dependabot checks

moe-ad:
Even if `${{ startsWith(github.head_ref, 'dependabot/') }}` evaluates to "false", it is still considered `truthy` in the context of bash if-statements since it is just a string to bash. We need to explicitly compare the values. Maybe I have misunderstood the intent here?

Related to #774


moe-ad:
Synced with @SMoraisAnsys offline. Already taken care of in #776.

PR #780, opened 2025-04-17
chore: add upper bound on marshmallow

SMoraisAnsys:
Avoid using marshmallow v4.0.0 otherwise running safety ends up with an exception being triggered:

`Unhandled exception happened: post_dump() got an unexpected keyword argument 'pass_many'`

Here is a failing workflow: https://github.com/ansys/pyacp/actions/runs/14518929119/job/40734794587

Note that this is affecting multiple projects and not only PyACP


MaxJPRey (approved):
LGTM.

RobPasMue:
Thanks a lot for this quick fix @SMoraisAnsys!

PR #782, opened 2025-04-21
fix: skip installation on doc-deploy-changelog

### LINKED ISSUE #779: doc-deploy-changelog action is unexpectedly platform-dependent

### 🐞 Description of the bug

I did not expect the doc-deploy-changelog action to require any specific platform, but it seems like we require package installation for some reason? This was a surprise to me when publishing `https://github.com/ansys/grantami-dataflow-extensions`, as that is a windows only package.

Can we try and remove the requirement to install the package in order to update some text files on disk? At worse it would mean parsing the METADATA file in the wheel, which surely pip can do?

Here is an example of a failing workflow with this symptom - https://github.com/ansys/grantami-dataflow-extensions/actions/runs/14520092650/job/40738678503

And attempting to use windows-2022 as a runner for this step also fails with another unrelated error - https://github.com/ansys/grantami-dataflow-extensions/actions/runs/14520414794/job/40739742199

### 📝 Steps to reproduce

Use the action against a project which does not support linux, I assume a similar problem would occur with mac

Windows

n/a

3.11

### 📦 Installed packages

```shell
n/a
```

RobPasMue:
We shouldn't need to install the project -- we can benefit directly from the version and name inputs.

Related to #723 -- not necessary anymore.

Closes #779

### DIFF
--- a/doc-deploy-changelog/action.yml
+++ b/doc-deploy-changelog/action.yml
@@ -239,6 +239,7 @@ runs:
             directory = config["tool"]["towncrier"]["directory"]
             project_name = config["tool"]["towncrier"].get("package")
 
+
         # If the project name is not specified, use the project name from the pyproject.toml
         # Last resource - just call it DNE. In any case, based on our template, this value is not used
         if project_name is None:
--- a/doc-deploy-changelog/action.yml
+++ b/doc-deploy-changelog/action.yml
@@ -137,9 +137,18 @@ inputs:
     description: |
       Whether to skip the installation of the project or not. If ``true``, the
       project is not installed. If ``false``, the project is installed. By
-      default, the project is not installed.
+      default, the project is installed.
     required: false
     default: false
+    type: boolean
+
+  skip-deps-install:
+    description: |
+      Whether to skip the installation of the project's dependencies or not. If ``true``, the
+      dependencies are not installed. If ``false``, the dependencies are installed. By
+      default, the project dependencies are NOT installed.
+    required: false
+    default: true
     type: boolean
 
   release-from-main:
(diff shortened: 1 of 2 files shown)


doc-deploy-changelog/action.yml
```diff
@@ -252,13 +237,19 @@ runs:
         with open(towncrier_config, 'r') as f:
             config = toml.load(f)
             directory = config["tool"]["towncrier"]["directory"]
+            project_name = config["tool"]["towncrier"].get("package")
+        
+        # If the project name is not specified, use the project name from the pyproject.toml
```
RobPasMue, line 242:
```suggestion
        # If the project name is not specified, use the project name from the pyproject.toml
        # Last resource - just call it DNE. In any case, based on our template, this value is not used
```

doc-deploy-changelog/action.yml
```diff
@@ -252,13 +237,20 @@ runs:
         with open(towncrier_config, 'r') as f:
             config = toml.load(f)
             directory = config["tool"]["towncrier"]["directory"]
+            project_name = config["tool"]["towncrier"].get("package")
```
jorgepiloto, line 240: Maybe, we also need to specify the `package_dir` as per this documentation: https://towncrier.readthedocs.io/en/stable/configuration.html#extra-top-level-keys-for-python-projects

jorgepiloto (commented):
Thanks for simplifying this. Just left a minor comment about wether we should specify the path to the `src/` directory or not.

doc-deploy-changelog/action.yml
RobPasMue: I would say that, in that case, projects should define it themselves -- if they need it. It is not strictly needed for building the towncrier changelog IMO. As long as we pass the package name (even if it is not used) and the version, that's good enough

jorgepiloto: Excellent. Thanks for clarifying this.

jorgepiloto (approved):

RobPasMue:
Merge and patch release?

PR #783, opened 2025-04-21
fix: add package key in towncrier.toml

jorgepiloto:
Fix the issues raised in https://github.com/ansys/actions/actions/runs/14574889354/job/40878852430 by adding the necessary keys in the `towncrier.toml` file.

### DIFF
--- a/towncrier.toml
+++ b/towncrier.toml
@@ -1,6 +1,5 @@
 [tool.towncrier]
-name = "ansys-actions"
-package = "ansys-actions"
+package = "ansys.actions"
 package_dir  = ".ci/ansys-actions/src"
 directory = "doc/source/changelog"
 filename = "doc/source/changelog.rst"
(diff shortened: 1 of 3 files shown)


towncrier.toml
```diff
@@ -1,5 +1,7 @@
 [tool.towncrier]
 name = "ansys-actions"
```
klmcadams, line 2: I think you can remove `name` and use `package` instead

towncrier.toml
```diff
@@ -1,5 +1,7 @@
...
+package = "ansys-actions"
```
klmcadams, line 3: I'm not sure if it matters, but other projects that specify the package name use `[tool.flit.module]`'s name. For example, `package = "ansys.mechanical.core"` in pymechanical

towncrier.toml
jorgepiloto: Oh! Excellent! I will merge this suggestion with your previous one, @klmcadams.

klmcadams (approved):

PR #784, opened 2025-04-21
fix: Add newline between environment variables in Python

klmcadams:
Newline between `TOWNCRIER_DIR` and `TOWNCRIER_NAME` was missing, causing the `TOWNCRIER_DIR` env variable's value to be `doc/source/changelogTOWNCRIER_NAME=ansys.actions.core`


jorgepiloto (approved):

RobPasMue (approved):
Thanks @klmcadams -- good catch

PR #794, opened 2025-04-29
feat: add compatibility with dependency groups

SMoraisAnsys:
Add compatibility with dependency groups from [PEP 735](https://peps.python.org/pep-0735/).

Since groups can be expressed using other groups, this PR also updates the python package used to read TOML files because `toml` does not work with such expressions. The new package used is `tomli` (note that a version of `tomli`, the `tomllib` module, was added to the standard library in Python 3.11 via [PEP 680](https://www.python.org/dev/peps/pep-0680/).

### DIFF
--- a/_doc-build-linux/action.yml
+++ b/_doc-build-linux/action.yml
@@ -200,7 +200,7 @@ runs:
 
     # ------------------------------------------------------------------------
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       with:
         level: "INFO"
         message: >
@@ -229,7 +229,7 @@ runs:
           echo "USE_DEPENDENCY_GROUPS=false" >> $GITHUB_OUTPUT
         fi
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       with:
         level: "INFO"
         message: >
@@ -239,7 +239,7 @@ runs:
 
     # ------------------------------------------------------------------------
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       with:
         level: "INFO"
         message: >
@@ -366,7 +366,7 @@ runs:
 
     # ------------------------------------------------------------------------
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       with:
         level: "INFO"
         message: >
@@ -416,7 +416,7 @@ runs:
 
     # ------------------------------------------------------------------------
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       with:
         level: "INFO"
@@ -430,14 +430,14 @@ runs:
         source .venv/bin/activate
         python ${{ github.action_path }}/../python-utils/parse_doc_conf.py
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       if: ${{ (inputs.add-pdf-html-docs-as-assets == 'true' ) && (env.PDF_FILENAME == '') }}
       with:
         level: "ERROR"
         message: >
           Unable to determine PDF filename using conf.py file.
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       if: ${{ (inputs.add-pdf-html-docs-as-assets == 'true' ) && (env.PDF_FILENAME != '') }}
       with:
         level: "INFO"
@@ -446,7 +446,7 @@ runs:
 
     # ------------------------------------------------------------------------
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       if: ${{ inputs.add-pdf-html-docs-as-assets == 'true' }}
       with:
         level: "INFO"
@@ -469,7 +469,7 @@ runs:
       run: |
         echo "EXISTS_EXPECTED_BUILD_DIR=$( [ -d "${EXPECTED_BUILD_DIR}" ] && echo 'true' || echo 'false' )" >> ${GITHUB_OUTPUT}
 
-    - uses: ansys/actions/_logging@main
+    - uses: ansys/actions/_logging@feat/work-with-dependency-groups
       if: ${{ (inputs.add-pdf-html-docs-as-assets == 'true') && (steps.check-expected-build-dir.outputs.EXISTS_EXPECTED_BUILD_DIR == 'false') }}
       with:
         level: "ERROR"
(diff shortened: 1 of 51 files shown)


_doc-build-linux/action.yml
```diff
@@ -210,6 +210,11 @@ runs:
           echo "BUILD_BACKEND=$(echo 'pip')" >> $GITHUB_OUTPUT
           echo "SPHINX_BUILD_MAKE=$(echo 'make')" >> $GITHUB_OUTPUT
         fi
+        if [[ -f "pyproject.toml" ]] && grep -q "\[project.optional-dependencies\]" "pyproject.toml"; then
```
RobPasMue, line 213: My feeling is that the if-statement is wrong... we should be checking for "dependency-groups", not "optional-dependencies" @SMoraisAnsys 

https://peps.python.org/pep-0735/#example-dependency-groups-table

This applies to the entire implementation

RobPasMue (changes requested):

_doc-build-linux/action.yml
SMoraisAnsys: Indeed, typo spotted

doc-changelog/action.yml
```diff
@@ -108,6 +115,18 @@ inputs:
 runs:
   using: "composite"
   steps:
+
+    - uses: ansys/actions/_logging@main
```
RobPasMue, line 119: This will make it a breaking change I guess. The PR can only be exposed/released starting on v10.

tests-pytest/action.yml
```diff
@@ -267,15 +272,20 @@ runs:
...
+        OPTIONAL_DEPENDENCIES_NAME: ${{ inputs.optional-dependencies-name }}
+        USE_PIP_DEPENDENCY_GROUPS: ${{ steps.github-environment-variables.outputs.USE_PIP_DEPENDENCY_GROUPS }}
       run: |
         $ACTIVATE_VENV
         if [[ "$BUILD_BACKEND" == 'poetry' ]]; then
           poetry install --with "$OPTIONAL_DEPENDENCIES_NAME"
         else
-          python -m pip install ".[${OPTIONAL_DEPENDENCIES_NAME}]"
+          if [[ "${USE_PIP_DEPENDENCY_GROUPS}" == 'true' ]]; then
+            python -m pip install --group "$OPTIONAL_DEPENDENCIES_NAME" .
```
RobPasMue, line 285: Does it accept several groups in comma-separated format? I.e. ``pip install --group group1,group2``

RobPasMue (approved):
Apart from my comments, LGTM! Thanks @SMoraisAnsys for paving the way to dependency group support!

tests-pytest/action.yml
SMoraisAnsys: Just tested it and it's not working. You'd have to add both group separately `pip install --group=group1 --group=group2`

doc-changelog/action.yml
SMoraisAnsys: Technically that's the case. I just checked our two orgs and couldn't find a repo where `toml-version` was leveraged except for our ansys/actions. So... :p
Still thinking that we should add this in v10

RobPasMue: I would wait for v10 nonetheless 😄

jorgepiloto (changes requested):
Thanks for adding this, @SMoraisAnsys. Should we enforce this? Knowing that we are going to release v10, it is the perfect excuse to only support groups. However, I may be missing something.

doc-changelog/action.yml
```diff
@@ -279,8 +298,8 @@ runs:
         towncrier_config = pyproject_file if pyproject_file.exists() else towncrier_file
 
         # Load pyproject.toml
-        with open(towncrier_config, 'r') as f:
-            config = toml.load(f)
+        with open(towncrier_config, 'rb') as f:
```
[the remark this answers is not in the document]
SMoraisAnsys: Package `tomli` requires the file to be opened in binary mode, see https://github.com/hukkin/tomli?tab=readme-ov-file#parse-a-toml-file

tests-pytest/action.yml
SMoraisAnsys: Oh, you are right. I was assuming that we should only work with a single install target.

SMoraisAnsys:
> Thanks for adding this, @SMoraisAnsys. Should we enforce this? Knowing that we are going to release v10, it is the perfect excuse to only support groups. However, I may be missing something.

Indeed, this would be a nice timing to enforce that. Let us discuss on it and I'll extend the PR with changes if needed.

MaxJPRey (approved):
LGTM.

_doc-build-linux/action.yml
```diff
@@ -222,6 +222,11 @@ runs:
           echo "BUILD_BACKEND=$(echo 'pip')" >> $GITHUB_OUTPUT
           echo "SPHINX_BUILD_MAKE=$(echo 'make')" >> $GITHUB_OUTPUT
         fi
+        if [[ -f "pyproject.toml" ]] && grep -q "\[dependency-groups\]" "pyproject.toml"; then
+          echo "USE_PIP_DEPENDENCY_GROUPS=true" >> $GITHUB_OUTPUT
+        else
+          echo "USE_PIP_DEPENDENCY_GROUPS=false" >> $GITHUB_OUTPUT
+        fi
```
jorgepiloto, line 229: We'll need to update this now that `uv` support has been implemented. Probably, we could resolve the package manager being used and pass the appropriate flags.

jorgepiloto (changes requested):

_doc-build-linux/action.yml
SMoraisAnsys: [8d10504](https://github.com/ansys/actions/pull/794/commits/8d105043c0666455156ec260e651f4514acae48e) adds support for `uv`

jorgepiloto (approved):
Awesome! Thanks for adding support for this, @SMoraisAnsys. Approving on my side.

SMoraisAnsys:
I just run a test with this branch in pyaedt https://github.com/ansys/pyaedt/actions/runs/15016189392
Let's wait to see if this works and merge it if it does :)

SMoraisAnsys:
Switching to draft mode as the PR points to the dev branch for testing, I'll do the revert once the pyaedt's PR run is passing

SMoraisAnsys:
CI passed in https://github.com/ansys/pyaedt/actions/runs/15020668511

_doc-build-windows/action.yml
```diff
@@ -248,6 +248,11 @@ runs:
           echo "BUILD_BACKEND=$(echo 'pip')" | Out-File -FilePath $env:GITHUB_OUTPUT -Encoding utf8 -Append
           echo "SPHINX_BUILD_MAKE=$(echo '.\make.bat')" | Out-File -FilePath $env:GITHUB_OUTPUT -Encoding utf8 -Append
         }
+        if ((Test-Path "pyproject.toml") -and (Get-Content "pyproject.toml" | Select-String -Pattern '\[dependency-groups\]')) {
```
SMoraisAnsys, line 251:
```suggestion
        # NOTE: Dependency groups are only used with pip and uv
        if ((Test-Path "pyproject.toml") -and (Get-Content "pyproject.toml" | Select-String -Pattern '\[dependency-groups\]')) {
```

tests-pytest/action.yml
```diff
@@ -147,6 +147,11 @@ runs:
         else
           echo "BUILD_BACKEND=$(echo 'pip')" >> $GITHUB_OUTPUT
         fi
+        if [[ -f "pyproject.toml" ]] && grep -q "\[dependency-groups\]" "pyproject.toml"; then
```
SMoraisAnsys, line 150:
```suggestion
        # NOTE: Dependency groups are only used with pip and uv
        if [[ -f "pyproject.toml" ]] && grep -q "\[dependency-groups\]" "pyproject.toml"; then
```

Issue #692, opened 2025-02-26
 Implement [zizmor] to fix common vulnerability issues in our actions


jorgepiloto:

jorgepiloto:
The author of Zizmor maintains a `pre-commit hook` in https://github.com/woodruffw/zizmor-pre-commit.

Maybe, we could just add this hook to our `.pre-commit-config.yaml` file and fix all the issues raised by the linter.

PR #796, opened 2025-04-30
fix: env var is missing after zizmore refactor

RobPasMue:
After #692 this doesn't work


moe-ad (approved):
Thanks for catching this @RobPasMue

PR #805, opened 2025-05-09
fix(wheelhouse): install from poetry.lock if present

### LINKED ISSUE #789: Bug located in actions/build-wheelhouse

### 🐞 Description of the bug

When building a wheelhouse for a poetry project I would expect the locked dependencies to be used, however the action installs the package using pip directly, so bypassing the locked dependencies and using some uncontrolled set.

For example: https://github.com/ansys/grantami-recordlists/tree/maint/update-serverapi-openapi

If you clone this branch and install it to a venv using `pip install .` you'll get one set of dependencies, but if you use `poetry install` and remove the dev dependencies you get another set.

| Package | Version (pip) | Version (poetry) |
| --------- | -------------- | ----------------- |
| ansys-grantami-recordlists | 1.3.0 | 1.3.0 |
| ansys-grantami-serverapi-openapi | 4.0.1 | 4.0.1 |
| ansys-openapi-common | 2.3.0 | 2.3.0 |
| certifi | 2025.1.31 | 2024.8.30 |
| cffi | 1.17.1 | 1.17.1 |
| charset-normalizer | 3.4.1 | 3.4.0 |
| cryptography | 44.0.2 | 43.0.3 |
| idna | 3.10 |  3.10 |
| pip | 24.0 | 24.0 |
| pycparser | 2.22 | 2.22 |
| pyparsing | 3.2.3 | 3.2.0 |
| pypiwin32 | 223 | 223 |
| pyspnego |  0.11.2 | 0.11.1 |
| python-dateutil | 2.9.0.post0 | 2.9.0.post0 |
| pywin32 | 310 | 308 |
| requests | 2.32.3 | 2.32.3 |
| requests-negotiate-sspi | 0.5.2 | 0.5.2 |
| requests_ntlm | 1.3.0 | 1.3.0 |
| setuptools | 65.5.0 | 65.5.0 |
| six | 1.17.0 | 1.16.0 |
| sspilib | 0.3.0 | 0.2.0 |
| urllib3 | 2.4.0 | 2.2.3 |

This is a problem both because its making CI fail due to the issue with metadata and pip licensing, but more importantly because the lock file is no longer the source of truth for our package.

### 📝 Steps to reproduce

As above

Windows

n/a

3.9

### 📦 Installed packages

```shell
n/a
```

jorgepiloto:
Fix #789 by installing from the `poetry.lock` file if present. This pull-request also adds the `working-directory` input so mono-repos can build the wheelhouse for its various libraries, as the `working-directory` input is not supported by GitHub with the `uses` syntax.

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -148,15 +148,6 @@ runs:
       run: |
         python -m pip install build wheel
 
-    - name: "Check if specific target is requested"
-      id: specific-target-requested
-      shell: bash
-      env:
-        TARGET: ${{ inputs.target }}
-      run: |
-        echo "install_target=$( [[ "${TARGET}" == '' ]] && echo '.' || echo ".[${TARGET}]")" >> ${GITHUB_OUTPUT}
-        echo "wheelhouse_target=$( [[ "${TARGET}" == '' ]] && echo 'wheelhouse' || echo "${TARGET}-wheelhouse")" >> ${GITHUB_OUTPUT}
-
     - name: "Identify the build system"
       id: build-system
       shell: bash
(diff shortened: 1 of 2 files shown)


build-wheelhouse/action.yml
```diff
@@ -153,16 +161,32 @@ runs:
...
+
+    - name: "Install the library"
       env:
         INSTALL_TARGET: ${{ steps.specific-target-requested.outputs.install_target }}
+        BUILD_BACKEND: ${{ steps.build-system.outputs.BUILD_BACKEND }}
+      shell: bash
       run: |
-        python -m pip install "${INSTALL_TARGET}"
+        if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
+          poetry install ${INSTALL_TARGET}
```
RobPasMue, line 186: We should check that we are doing this as well in other actions -- I'm thinking mostly the testing and docs actions, although I think it's already the case

jorgepiloto: It looks like the other actions are using poetry install commands. I moved the changes for the working directory in https://github.com/ansys/actions/pull/807.

RobPasMue:
Can we get this merged and patch released?

da1910 (changes requested):
I _think_ we need to make sure that we are not installing dev or any other groups, it should only be the root unless we manually request additional depedency groups.

jorgepiloto:
Patched in https://github.com/ansys/actions/actions/runs/14972143164

PR #806, opened 2025-05-09
fix(vulnerabilities): remove marshmallow dependency limit

jorgepiloto:
Changes introduced in https://github.com/ansys/actions/pull/780/ are no longer required. Packages are now compatible.


SMoraisAnsys (approved):
Seems like it's working yeah. Double checked to be sure and no issues locally :)
Thanks @jorgepiloto

RobPasMue (approved):

PR #810, opened 2025-05-10
feat: Add breaking change section

### LINKED ISSUE #722: Add breaking changes to the changelog

A great enhancement (very subjective) to the changelog would be to have a section "breaking changes" when a PR has introduced such kind of changes.

Following the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), introducing a breaking change can be expressed in a commit (here PR) if it has a footer (here PR title) `BREAKING CHANGES: ` OR appends a `!` after the type/scope.

If such feature makes sense to you too, we can even think about having a common doc files for the breaking changes, e.g. the same kind of file we are using for the migration in the ansys/actions, and have a direct link added to this file in the change log or the release notes !

_No response_

- [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

SMoraisAnsys:
Adds checking for breaking change in the PR title and body. On repos following [conventionnal commits](https://www.conventionalcommits.org/en/v1.0.0/), this will greatly help:
- users to know what breaking changes are introduced in a specific version;
- developers to have an idea of how often they are breaking their API.

I think that this could be used as a metric to know whether a repo is ready to move to version `1.0.0`

Close #722

### DIFF
--- a/python-utils/parse_pr.py
+++ b/python-utils/parse_pr.py
@@ -1,4 +1,5 @@
 import os
+import re
 from pathlib import Path
 
 import toml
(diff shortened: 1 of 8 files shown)


RobPasMue:
Hmmm... Weren't breaking changes done as ``feat:!...``

Can't access easily the conventional commits site (security issue apparently) but see https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13#breaking-changes-indicator

python-utils/parse_pr.py
```diff
@@ -61,20 +62,67 @@ def get_first_letter_case(pr_title: str):
...
+    pr_body: str
+        The pull request body.
+
+    Returns
+    -------
+    bool
+        True if the pull request body indicates a breaking change, False otherwise.
+    """
+    if not pr_body:
+        raise False
```
SMoraisAnsys, line 102:
```suggestion
        return False
```

PR #811, opened 2025-05-10
fix: avoid using pip cache with uv

SMoraisAnsys:
Seems like, for some workflows, the fact that we use `uv` and keep using `pip` cache by default leads to a workflow error when the post run action is performed. The associated error message is:

Error: Cache folder path is retrieved for pip but doesn't exist on disk: /home/runner/.cache/pip. This likely indicates that there are no dependencies to cache. Consider removing the cache step if it is not needed.

See:
- failure when using cache (by default) https://github.com/ansys/actions/actions/runs/14945638158/job/41988577383
- success when disable the cache (manually) https://github.com/ansys/actions/actions/runs/14945529201

### DIFF
--- a/_setup-python/action.yml
+++ b/_setup-python/action.yml
@@ -29,7 +29,7 @@ inputs:
     required: true
     type: string
   use-cache:
-    description: 'Whether to use Python cache or not. Requires a pyproject.toml file.'
+    description: 'Whether to use Python/uv cache or not. Requires a pyproject.toml file.'
     required: true
     type: boolean
   provision-uv:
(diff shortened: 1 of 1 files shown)


_setup-python/action.yml
```diff
@@ -42,20 +42,25 @@ runs:
   using: "composite"
   steps:
 
+    # NOTE: When a workflow is triggered without using the cache, it can lead to a failure
+    # in the Post Run action sayng that the cache folder doesn't exist on disk.
```
moe-ad, line 46:
```suggestion
    # in the Post Run action sayng that the cache folder doesn't exist on disk.
```
```suggestion
    # in the Post Run action sayng that the cache folder doesn't exist on disk.
```
```suggestion
    # in the Post Run action saying that the cache folder doesn't exist on disk.
```

moe-ad (approved):
Thanks for this @SMoraisAnsys!

PR #816, opened 2025-05-13
fix: crash when reading pyproject (broken toml library)

### LINKED ISSUE #809: doc-style action: toml decoder crash on valid pyproject
We would like to switch to UV in <https://github.com/ansys/pysimai/pull/136>.
But the doc-style action crashes because `toml` can't decode this:

```toml
[dependency-groups]
dev = [
    "ruff",
    { include-group = "docs" },
]
docs = [
    "sphinx",
]
```

See action log:
https://github.com/ansys/pysimai/actions/runs/14931356113/job/41948167837?pr=136

Offending line:
https://github.com/ansys/actions/blob/6570eec9ebb74f76baadb8fdda443fd00a0bcf24/doc-style/action.yml#L167C1-L167C38

awoimbee:
Fixes #809

Replaces the outdated `toml` by the python stdlib's `tomllib`.

This might be deemed breaking since:

* I'm removing an input
* The action now requires python >= 3.11 (ubuntu 24.04 has 3.12 by default)

### DIFF
--- a/doc/source/changelog/809.fixed.md
+++ b/doc/source/changelog/809.fixed.md
@@ -1,0 +1,1 @@
+replace broken `toml` by `tomllib`: allow more contructs in pyproject.toml
--- a/doc/source/changelog/809.fixed.md
+++ b/doc/source/changelog/809.fixed.md
@@ -1,1 +1,0 @@
-replace broken `toml` by `tomllib`: allow more contructs in pyproject.toml
--- a/doc-style/action.yml
+++ b/doc-style/action.yml
@@ -62,13 +62,6 @@ inputs:
     required: false
     type: string
 
-  toml-version:
-    description: |
-      Toml version used for retrieving the towncrier directory.
-    default: '0.10.2'
-    required: false
-    type: string
-
   ignore-changelogd:
     description: |
       Whether or not to ignore markdown files in doc/changelog.d/.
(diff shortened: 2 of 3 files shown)


doc/source/changelog/809.fixed.md
RobPasMue, line 1: Weird... why is this file part of your PR? We should probably get rid of it

RobPasMue (approved):
Apart from my comment - LGTM! This change will have to wait for v10 to be released though (breaking change -- removing the input)

doc/source/changelog/809.fixed.md
awoimbee: Ah I didn't know these were created by a bot

RobPasMue:
The issue you are hitting is probably related to #811

doc/source/changelog/809.fixed.md
RobPasMue: Thanks @awoimbee!

SMoraisAnsys (changes requested):
Please do not merge this PR ftm. This problem is already handled in https://github.com/ansys/actions/pull/794 which introduces dependency groups.

I'm open to discussion on the package to use though but I'd adbise tomli since it's the base for the toml builtin of python

SMoraisAnsys:
Thanks again for this PR @awoimbee. Until we release `v10` you might want to use `ansys/action@main` to be able to use dependency groups with our actions.

PR #822, opened 2025-05-14
fix(wheelhouse): virtual environment activation

jorgepiloto:
Looks like activating the virtual environment when using Poetry causes some environment variables to be hidden by the new environment. This prevents projects from installing, for instance, from the private PIP indices despite having set the right environment variables.


jorgepiloto:
Closing this. The issue is now solved in the main branch.

PR #825, opened 2025-05-15
chore: update uv settings

SMoraisAnsys:
While checking the current main status this warning appeared

image

Seems like the default glob pattern have been updated and we could leverage that to match our default behavior (use of `pyproject.toml`). Also:
- Action `astral-sh/setup-uv` was not tracked by dependabot so it has been added.
- A new input has been added to `actions/_setup-python` to allow state if `uv` cache should be pruned or not. The default value is following `setup-uv`'s default behavior, see https://github.com/astral-sh/setup-uv?tab=readme-ov-file#disable-cache-pruning for more information.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -60,6 +60,7 @@ updates:
         patterns:
           - "docker/setup-qemu-action"
           - "pypa/cibuildwheel"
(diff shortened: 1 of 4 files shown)


jorgepiloto (approved):
Good catch, @SMoraisAnsys!

PR #826, opened 2025-05-16
fix: build wheelhouse with poetry

SMoraisAnsys:
This PR strengthen the compatibility of the `build-wheelhouse` action with `poetry`.

Currently, if one uses the main branch, the following error might appear https://github.com/ansys/pydpf-composites/actions/runs/15046028749/job/42288662063. Long story short, `pip freeze` doesn't use the revision defined in the poetry lock file and that can lead to a failure when building the wheelhouse. In the workflow failure, the project was installed with revision "1453403c76957981c84432cf67b43932ca9e99a5" while pip freeze doesn't contain this information and leads to the latest reference in main, i.e. revision "06170e14d2f47a9e53b59015b72e7b0dd897af9c".

This changes are successfully tested in https://github.com/ansys/pydpf-composites/actions/runs/15069361200

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -203,7 +203,7 @@ runs:
       run: |
         python -m pip install pipx
         python -m pipx install poetry
-        python -m pipx upgrade poetry
+        python -m pipx inject poetry poetry-plugin-export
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):
LGTM - Thanks @SMoraisAnsys

PR #828, opened 2025-05-16
docs: v10

moe-ad:
Closes #797

PR #833, opened 2025-05-20
feat: ensure matching and metadata version

### LINKED ISSUE #824: Add action to check that the version in a tag matches the version in the package metadata

The recommended workflow strategy (see readme of https://github.com/ansys/actions) includes pushing a tag to trigger the release of the package to PyPI, create the release on Github, and deploy the stable docs.

The recommended tag format is `v{version}`. We've accidently attempted to push version tags that did not match the version specified in the project's pyproject file and that results in inconsistent releases:
- wheels are built with the version from the pyproject file
- wheels are published to PyPI with the version from the pyproject file, and when using pypa/gh-action-pypi-publish, the publish step does not fail and does not log any errors/warnings if the version already exists.
- the Github release is created after the tag version but includes the wheels with the version from the pyproject file

We're considering writing a re-usable workflow to validate that the version from the tag matches the version in the pyproject file.
Is that something that could be done as an ansys action? Or added to one of the existing actions? For example an additional check on the build-library action or build-wheelhouses action to check that, if there is a tag, the version matches the metadata version.

_No response_

_No response_

moe-ad:
Closes #824


moe-ad:
`build-library` tests:
- Positive test: https://github.com/ansys-internal/test-repo-moe/actions/runs/15135197584/job/42546518504
- Negative test: https://github.com/ansys-internal/test-repo-moe/actions/runs/15135972164/job/42547883972

SMoraisAnsys:
Some project are dynamically specifying their version through python files. I'm not sure the current logic works for them

moe-ad:
@SMoraisAnsys can you provide an example? So I can better understand that other scenario.

RobPasMue:
Agreed. See the API packages for example: https://github.com/ansys/ansys-api-geometry/blob/main/setup.py
https://github.com/ansys/ansys-api-geometry/blob/main/ansys/api/geometry/VERSION

I would recommend to use ``importlib_metadata`` for this purpose. Just install the library, check the version... and make sure it matches the tag. ``importlib_metadata`` parses versions properly AFAIK

moe-ad:
Thanks for the explanation @RobPasMue !

moe-ad:
> `build-library` tests:
> 
> * Positive test: https://github.com/ansys-internal/test-repo-moe/actions/runs/15157917819/job/42617360774
> * Negative test: https://github.com/ansys-internal/test-repo-moe/actions/runs/15158196949/job/42618148432
> 
> `build-wheelhouse` tests:
> 
> * Positive test: https://github.com/ansys-internal/test-repo-moe/actions/runs/15158500348/job/42619057329
> * Negative test: https://github.com/ansys-internal/test-repo-moe/actions/runs/15158680847/job/42619637647

@RobPasMue @SMoraisAnsys 
The test case links have been updated for tests against latest changes.

build-library/action.yml
```diff
@@ -108,29 +108,84 @@ runs:
...
+      shell: bash
+      env:
+        PYTHON_VERSION: ${{ inputs.python-version }}
+      run: |
+        python_version="${PYTHON_VERSION}"
+        minor_python_version=$(echo "$python_version" | awk -F "." '{print $2}')
+        major_python_version=$(echo "$python_version" | awk -F "." '{print $1}')
+        if (( $(( $major_python_version == 3 )) )) && (( $(( $minor_python_version < 8 )) )); then
+          echo "needs_importlib_metadata=true" >> ${GITHUB_OUTPUT}
+        elif (( $(( $major_python_version < 3 )) )); then
```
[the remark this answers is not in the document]
moe-ad: Hmmn.
We should drop it then. I will update the actions accordingly.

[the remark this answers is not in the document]
SMoraisAnsys: I agree, I removed it yesterday from the wheelhouse action in #834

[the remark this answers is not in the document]
moe-ad: Based on your notes on #834, I guess we do not need to mention it then in the docs(#828) then?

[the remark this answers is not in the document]
RobPasMue: I would add it to the release notes as breaking changes that we are stopping to support <3.8 logic

build-library/action.yml
```diff
@@ -115,6 +115,46 @@ runs:
...
+        INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --system' || 'python -m pip install' }}
+      run: |
+        $INSTALL_COMMAND .
+
+    - name: "Verify library is properly installed and get its version number"
+      id: library-version
+      shell: bash
+      env:
+        LIBRARY_NAME: ${{ inputs.library-name }}
+      run: |
```
RobPasMue, line 130: If this is only needed in case of releasing, I would move the logic inside the next step. We don't need to run it otherwise I believe.

RobPasMue: Although it doesn't hurt either

moe-ad: That makes sense. Initially I just reused the logic from `build-wheelhouse` so I missed that.
Thanks!

build-library/action.yml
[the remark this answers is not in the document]
moe-ad: Added.

build-library/action.yml
```diff
@@ -105,6 +105,31 @@ runs:
...
+          echo "python_version_supported=true" >> ${GITHUB_OUTPUT}
+        fi
+
+    - uses: ansys/actions/_logging@main
+      if: steps.python-version-support.outputs.python_version_supported == 'false'
+      with:
+        level: "ERROR"
+        message: >
+          Python version ${{ inputs.python-version }} is no longer supported. Please use a
+          version greater than 3.7.
```
RobPasMue, line 131: We may want to move this into our setup-python action? What do you think @SMoraisAnsys @moe-ad @jorgepiloto ?

RobPasMue (approved):
One small comment - otherwise LGTM!

build-library/action.yml
moe-ad: Yes, it will be nice to centralize things. But that would mean we are dropping python 3.7 support across all our actions.
I guess one could argue that in the edge case of needing to use our some of actions for `python <= 3.7`, `v9` or below can be used.

jorgepiloto: Yes, let's do it in this same pull-request, @moe-ad. Thanks for suggesting this, @RobPasMue.

moe-ad: Done.

jorgepiloto (approved):
Eggselent! 🥚 Thanks for adding this, @moe-ad.

RobPasMue (approved):
LGTM! Thanks @moe-ad

SMoraisAnsys (approved):
Thanks a lot, this will help preventing misses !

moe-ad:
Tests:
- Positive test: https://github.com/ansys-internal/test-repo-moe/actions/runs/15186003845
- Negative tests:
  - Unmatching tags: https://github.com/ansys-internal/test-repo-moe/actions/runs/15186225363
  - Unsupported python version: https://github.com/ansys-internal/test-repo-moe/actions/runs/15186648402

build-library/action.yml
```diff
@@ -109,12 +109,43 @@ runs:
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
     - name: "Set up Python"
-      uses: ansys/actions/_setup-python@main
+      uses: ansys/actions/_setup-python@feat/ensure-matching-tag-and-metadata-version
```
RobPasMue, line 112: Let's not forget to remove this

moe-ad: Done.

PR #834, opened 2025-05-20
feat: add SBOM to wheelhouse action

SMoraisAnsys:
This PR adds the generation of SBOM to the wheelhouse action. It follows a bullet point in https://github.com/ansys/pyansys-dev-guide/issues/558#issuecomment-2721532087

> [!WARNING]
The PR introduces breaking changes to `build-wheelhouse`. Now, the action doesn't install packages at system level and leverages a virtual environment. This change will break some users workflows because some of them were performing smoke tests just after calling `build-wheelhouse`. Therefore, another input `smoke-test-module` is added to allow for smoke tests on the fly.

The changes have been tested and succeeded in :
- https://github.com/ansys-internal/simba-plugin-post/actions/runs/15213284768
- https://github.com/ansys/pydpf-composites/actions/runs/15213301612
- https://github.com/ansys/pyaedt/actions/runs/15212332836/job/42789517492
Note that I tested this action on another private repo and using input `smoke-test-module` allowed to detect a missing dependency :)

The approach used consist in using [Syft](https://github.com/anchore/syft) a CLI tool and library for generating a Software Bill of Materials. The choice of using this tool is that it is well known, supports SPDX format (which we agreed on using) and supports Python. If we ever want to use another format, we could investigate [CycloneDX](https://cyclonedx.org/) which has a python binding [cyclonedx-python](https://github.com/CycloneDX/cyclonedx-python).

The implementation of Syft follows the [recommended approach](https://github.com/anchore/syft?tab=readme-ov-file#recommended) and arbitrary sets the version to the latest available.

The SBOM is currently generated from a virtual environment where we install the project with the specified install target. Another approach could be to dump the dependencies into a sbom_req.txt file and use it to generate the SVOM. Let me know if you think that it makes more sense or not.

Note that the SBOM file generated from the venv can contain file information and not only package information. For example, when run on pyaedt, it ends up with METADATA files (e.g. \typing_extensions-4.13.2.dist-info\METADATA) or DLL files (e.g. \pythonnet\runtime\Python.Runtime.dll)


RobPasMue:
The approach seems reasonable and allows us to move independently from the package managers used. I would definitely avoid ``pipdeptree`` for this since we would have to build it ourselves (the SBOM) if I am not mistaken.

Having a list of the dependencies used in a ``reqs.txt`` file format (with the pinned versions) is not a bad idea anyway (additionally to the SBOM)  but in that case I would just do a "pip freeze > reqs.txt``.

I'm eager to see what the ``uv`` team comes up with when implementing this. Maybe they end up having a built-in feature too.

da1910:
> @RobPasMue @jorgepiloto @greschd @da1910 Can you guys give me some feedback on this way of generating SBOM ? Would the approach of dumping deps into a TXT file (using `pipdeptree` or `poetry export`) would be a better choice ?

I actually like this approach, it's what we did with poetry for our pyinstaller based products. Rather than relying on the lock file we actually verify what's present.

The approach for SPDX files has, historically, been to only list the packages present rather than the individual files. We would for example list `setuptools version 75.11.0` rather than each component file. However if this tool works with individual files then there's no issue with being more thorough.

RobPasMue:
> I do think that cycloneDX is going to _win_ in the long term, does syft support cyclonedx export?

Looks like it does: https://github.com/anchore/syft?tab=readme-ov-file#output-formats

.github/workflows/ci_cd_main.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     name: "Doc build"
     runs-on: ubuntu-latest
     steps:
-      - uses: ansys/actions/doc-build@main
+      - uses: ansys/actions/doc-build@feat/add-sbom-to-wheelhouse-action
```
RobPasMue, line 23: Let's remember reverting all this

RobPasMue (approved):
Pending confirmation on poetry-based projects.. LGTM

SMoraisAnsys:
Going in draft due to failure in pyaedt https://github.com/ansys/pyaedt/actions/runs/15210265720/job/42782538786

SMoraisAnsys:
@RobPasMue @jorgepiloto The action scope has changed a bit and the (breaking) changes are introduced in the warning section of the description. Otherwise, I think this is ready and tested. Let me know if you have any feedback :)

PR #836, opened 2025-05-21
refactor: allow to mix dependency groups and optional targets

SMoraisAnsys:
As title says. The PR aims at allowing `tests-pytest`  and `doc-build` to work with both optional targets and group dependency.

Latest workflow run : [https://github.com/ansys/pyaedt/actions/runs/15256855864 (currently in progress can't state the success or not)](https://github.com/ansys/pyaedt/actions/runs/15256855864/job/42961696773?pr=6093)
Note that the linux failure is related to the same artifacts being uploaded with the same name... :)


tests-pytest/action.yml
```diff
@@ -96,15 +96,24 @@ inputs:
...
       default file is ``requirements/requirements_tests.txt``.
     default: 'tests'
     required: false
     type: string
 
+  group-dependencies-name:
+    description: |
+      Any valid dependency groups defined in the ``pyproject.toml`` file. The associated
+      dependencies are installed before running the tests.
+      The default value is ``''``.
```
moe-ad, line 112: Do you think mentioning that multiple groups must be comma separated helps for more clarity? I guess one could say the same for ``optional dependency names`` too.
But maybe users are already used to passing them as comma separated.

tests-pytest/action.yml
```diff
@@ -301,24 +305,30 @@ runs:
...
+          optional_dependencies_flag=""
+          if [ -n "${OPTIONAL_DEPENDENCIES_NAME}" ]; then
+            optional_dependencies_flag="[${OPTIONAL_DEPENDENCIES_NAME}]"
+          fi
+          # Group dependencies
+          group_dependencies_flag=""
+          if [ -n "${GROUP_DEPENDENCIES_NAME}" ]; then
+            IFS=',' read -ra group_list <<< "${GROUP_DEPENDENCIES_NAME}"
+            for group in "${group_list[@]}"; do
+              group_dependencies_flag="$group_dependencies_flag --group $group"
```
jorgepiloto, line 324:
```suggestion
              group_dependencies_flag="$group_dependencies_flag --group ${group}"
```

jorgepiloto (approved):
Just left one minor suggestion. Thanks for setting this in place, @SMoraisAnsys.

tests-pytest/action.yml
SMoraisAnsys: @jorgepiloto Do you think we should apply that behavior everywhere ? AFAIK, we use both ${var} or $var in our code base. For example here there is also `group_dependencies_flag` and there are no `{...}`.

If that's the case, I propose to leave the code as it is and to perform that in another PR. However, I'm not too sure about the gain of using `{...}` unless the variable is used in composition with something else, e.g. `"{dependency}.zip"`

jorgepiloto: Sure, no problem. It would be great if we could have shellcheck at some point for this. however, it does not work with actions.

jorgepiloto (commented):
Almost there. Just one minor question, @SMoraisAnsys.

_doc-build-windows/action.yml
```diff
@@ -391,16 +402,34 @@ runs:
...
+        if ($env:BUILD_BACKEND -eq "poetry") {
+          poetry run -- make.bat html SPHINXOPTS="$env:SPHINXOPTS"
+          poetry run -- make.bat pdf
+          if ($env:CHECK_LINKS -eq 'true') {
+            poetry run -- make.bat pdf linkcheck SPHINXOPTS="$env:SPHINXOPTS"
+          }
+        }
+        else {
+          if ($env:BUILD_BACKEND -eq "uv") {
+            uv run --no-sync -- .\make.bat html SPHINXOPTS="$env:SPHINXOPTS"
```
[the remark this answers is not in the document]
SMoraisAnsys: I don't have an answer for that, as I simply reused what was already defined in the previous implementation.
However, it seems that locking and synchronizing are always performed before doing any run, see https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync
I think, that's why :)

moe-ad:
@SMoraisAnsys I found another repo to test the `uv` fix in this PR.

Success run(against this branch): https://github.com/ansys/pyscadeone/actions/runs/15296186237/job/43025805839?pr=39
Failed run(against main): https://github.com/ansys/pyscadeone/actions/runs/15295952487/job/43025058449

Thanks!

_doc-build-windows/action.yml
```diff
@@ -445,7 +474,7 @@ runs:
       id: set-expected-build-directory
       shell: powershell
       run: |
-        echo "EXPECTED_BUILD_DIR=$(echo 'doc\_build')" | Out-File -FilePath $env:GITHUB_OUTPUT -Encoding utf8 -Append
+        echo "EXPECTED_BUILD_DIR=$(echo '_build')" | Out-File -FilePath $env:GITHUB_OUTPUT -Encoding utf8 -Append
```
jorgepiloto, line 477: I am surprised this works. I thought that even if using `cd` in previous every step defaults to the base directory.

RobPasMue: Yeah... me too.

RobPasMue: Verified here by @SMoraisAnsys -- the previous location is not kept. We need to check what's going on here.

See https://github.com/ansys/actions/pull/870#issuecomment-2929321191

RobPasMue (approved):
LGTM! Thanks @SMoraisAnsys!

jorgepiloto:
Awesome, thanks for implementing and testing this, @SMoraisAnsys.

PR #839, opened 2025-05-22
fix(_setup-python): update to latest version

jorgepiloto:
This pull-request ensures we are using the latest version of the `actions/setup-python` in our actions.


RobPasMue (approved):

SMoraisAnsys (approved):
I don't get why dependabot didn't open a PR, the last one we have is 3 weeks ago https://github.com/ansys/actions/pull/795

This should be investigated

PR #840, opened 2025-05-22
fix: discovery of artifacts in release-github

jorgepiloto:
The current logic of the `release-github` does not guarantee the finding of the right artifacts. Also, additional artifacts may be included.

This pull-request ensures all wheel and sdist artifacts are found and listed down in the release notes.


RobPasMue (approved):
One small typo only

release-github/action.yml
```diff
@@ -347,26 +347,24 @@ runs:
...
+          echo "You can verify this release's artifact attestations using [GitHub's command-line tool](https://cli.github.com/). For example, download the files from this release page or from [PyPI](https://pypi.org/). Then, run the following commands:"
           echo ""
           echo "\`\`\`bash"
-          echo "$ gh attestation verify --owner ${{ github.repository_owner }} ${SDIST_FILE}"
-          echo "\`\`\`"
-          echo "and"
-          echo "\`\`\`bash"
-          echo "$ gh attestation verify --owner ${{ github.repository_owner }} ${WHEEL_FILE}"
+  
+          for artifact in $(find dist -type f \( -name "*.whl" -o -name "*.tar.gz" \)); do
```
SMoraisAnsys, line 361: Should we extend the description to state that wheelhouse available on Github release page can also be verified ?

SMoraisAnsys (commented):
Left a minor proposition, thanks for the changes @jorgepiloto

SMoraisAnsys:
> The current logic of the `release-github` does not guarantee the finding of the right artifacts. Also, additional artifacts may be included.
> 
> This pull-request ensures all wheel and sdist artifacts are found and listed down in the release notes.

I didn't consider that a package may have multiple wheels depending on python compatibility and arch. Thanks for catching that

release-github/action.yml
jorgepiloto: Sure, sounds good.

jorgepiloto: Done in https://github.com/ansys/actions/pull/840/commits/4e45153ed35cb883a727d25fe27e679a009e2da9

release-github/action.yml
```diff
@@ -347,23 +347,22 @@ runs:
...
           echo ""
           echo "\`\`\`bash"
-          echo "$ gh attestation verify --owner ${{ github.repository_owner }} ${SDIST_FILE}"
-          echo "\`\`\`"
-          echo "and"
-          echo "\`\`\`bash"
-          echo "$ gh attestation verify --owner ${{ github.repository_owner }} ${WHEEL_FILE}"
+  
+          # Find the following artifacts: wheels, source distribution, and wheelhouse files
+          for artifact in $(find dist -type f \( -name "*.whl" -o -name "*.tar.gz" -o name "*.zip" \)); do
```
SMoraisAnsys, line 362:
```suggestion
          for artifact in $(find dist -type f \( -name "*.whl" -o -name "*.tar.gz" -o name "*wheelhouse*.zip" \)); do
```
I'm fearing that documentation artifacts are uploaded as ZIP files

RobPasMue: Ah good catch but... aren't they signed too? We can also provide signed docs 😆

SMoraisAnsys: Yeah, I understood after typing and edited but you answered faster :')

SMoraisAnsys (approved):

PR #841, opened 2025-05-22
ci: remove dependabot cooldown ftm

SMoraisAnsys:
Assuming that our lack of updates from dependabot is associated to the fact the github actions are not yet handled by the cooldown feature, we should avoid using it for the moment.

Also, JSYK, the compatibility with `pip` is not yet dry as install targets are not well handled, see https://github.com/dependabot/dependabot-core/issues/12102

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,17 +1,10 @@
 version: 2
-enable-beta-ecosystems: true
 updates:
   - package-ecosystem: "pip"
     directories:
       - "/requirements"
     schedule:
       interval: "daily"
-    cooldown:
-      default-days: 7
-      include:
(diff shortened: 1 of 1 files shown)


SMoraisAnsys:
@RobPasMue Double checked the docs and by default github actions are looked in `.github` even if we state `directory="/"`, see https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions
Hopefully this changes should handle the problem. Thanks for pointing that log

RobPasMue (approved):
LGTM! 🤞🏼

PR #850, opened 2025-05-22
fix: default value for prune-uv-cache

moe-ad:
As the title says. Just a small change to #825.

### DIFF
--- a/.github/workflows/ci_cd_night.yml
+++ b/.github/workflows/ci_cd_night.yml
@@ -61,7 +61,7 @@ jobs:
         python-version: ${{ env.MAIN_PYTHON_VERSION }}
         use-cache: false
         provision-uv: false
-        prune-uv-cache: true
+        prune-uv-cache: false
 
     - name: "Install build and twine"
       shell: bash
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -177,7 +177,7 @@ jobs:
         python-version: ${{ env.MAIN_PYTHON_VERSION }}
         use-cache: false
         provision-uv: false
-        prune-uv-cache: true
+        prune-uv-cache: false
 
     - name: "Install build and twine"
       shell: bash
--- a/.github/workflows/ci_cd_release.yml
+++ b/.github/workflows/ci_cd_release.yml
@@ -160,7 +160,7 @@ jobs:
         python-version: ${{ env.MAIN_PYTHON_VERSION }}
         use-cache: false
         provision-uv: false
-        prune-uv-cache: true
+        prune-uv-cache: false
 
     - name: "Install build and twine"
       shell: bash


moe-ad:
@SMoraisAnsys it should be this way right?

_setup-python/action.yml
```diff
@@ -47,8 +47,9 @@ inputs:
         When uv cache is pruned, pre-built wheels are removed but wheels built from source
         are retained.
 
-    required: true
+    required: false
```
SMoraisAnsys, line 50:
```suggestion
    required: true
```
By convention all inputs of "private" action are required

build-library/action.yml
```diff
@@ -114,6 +114,7 @@ runs:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
         provision-uv: ${{ inputs.use-uv }}
+        prune-uv-cache: true
```
SMoraisAnsys, line 117: Shouldn't we have something that aligns with `inputs.use-python-cache` ? If use cache is activated then we should not prune uv cache, right ?

RobPasMue: I was thinking the same... why not reusing the logic in ``inputs.use-python-cache`` but negated, right?

.github/workflows/ci_cd_night.yml
```diff
@@ -60,6 +60,8 @@ jobs:
       with:
         python-version: ${{ env.MAIN_PYTHON_VERSION }}
         use-cache: false
+        provision-uv: false
+        prune-uv-cache: true
```
jorgepiloto, line 64: From the documentation of the action, it is stated that the pruning of `uv` cache only applies if the `provision-uv` is set to `true`.

_release-pypi/action.yml
```diff
@@ -133,6 +133,7 @@ runs:
         python-version: ${{ inputs.python-version }}
         use-cache: false
         provision-uv: ${{ inputs.use-uv }}
+        prune-uv-cache: true
```
jorgepiloto, line 136: This depends on the value of the `provision-uv` input.

check-vulnerabilities/action.yml
```diff
@@ -245,6 +245,7 @@ runs:
         python-version: ${{ inputs.python-version }}
         use-cache: false
         provision-uv: ${{ inputs.use-uv }}
+        prune-uv-cache: true
```
jorgepiloto, line 248: Same dependency on the `provision-uv` input here.

jorgepiloto (changes requested):
Thanks for adding this. I think you must apply the condition on the value of the `provision-uv` in the inputs.

.github/workflows/ci_cd_night.yml
moe-ad: Synced with Jorge offline to explain the rationale a bit further. Agreed that we should let set `prune-uv-cache` to the value of `provision-uv` (_false_) for the CI/CD files. It makes sense to turn everything regarding `uv` off in those cases, even if the variable is not being used. Besides that, we are good to go.

PR #853, opened 2025-05-26
fix: check licenses logic alignment

RobPasMue:
After investigations, this might be causing it. Let's align the criteria


build-wheelhouse/action.yml
```diff
@@ -391,7 +391,7 @@ runs:
         subject-path: ${{ inputs.library-name }}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}.zip
 
     - name: Check library's dependencies license
-      uses: ansys/actions/check-licenses@main
+      uses: ansys/actions/check-licenses@fix/check-licenses-logic
```
RobPasMue, line 394:
```suggestion
      uses: ansys/actions/check-licenses@main
```
To be reverted

PR #855, opened 2025-05-27
fix: housekeeping package clean actions

moe-ad:
Detected [here](https://github.com/ansys/pyacp/pull/858).
Failed run: https://github.com/ansys/pyacp/actions/runs/15270039065/job/42943287636
Test against this fix: https://github.com/ansys/pyacp/actions/runs/15270429867/job/42944531744


jorgepiloto (approved):
Thanks for fixing this, @moe-ad.

PR #861, opened 2025-05-27
chore: improve smoke tests handling

### LINKED ISSUE #859: `smoke-test-module` input
I do not think it is a good idea to have a fixed way to test the module.

https://github.com/ansys/actions/blob/da9e73e851cffbff51f850c19c639fb9738e22fa/build-wheelhouse/action.yml#L410

Why not?

```yml
  smoke-test-module:
    description: |
      Code to execute to test the python import.
    default: ''
    required: false
    type: string

...

    - name: Run smoke test
      if: inputs.smoke-test-module != ''
      env:
      shell: bash
      run: |
        ${ACTIVATE_VENV}
        python -c "${{ inputs.smoke-test-module }}"
```

Futhermore, I would probably make the testing mandatory by:

```yml
  smoke-test-module:
    description: |
      Code to execute to test the python import.
    default: "import ${MODULE}; from ${MODULE} import __version__; print(f'Imported: {${MODULE}}, version: {__version__}')"
    required: false
    type: string

...

    - name: Run smoke test
      if: inputs.smoke-test-module != ''
      env:
        MODULE: ${{ inputs.smoke-test-module }}
      shell: bash
      run: |
        ${ACTIVATE_VENV}
        python -c "${{ inputs.smoke-test-module }}" # check for right combination of single and double ticks
```

You still dan disable it by leaving it empty... but by default it will try to run your code unless specific string is provided.

moe-ad:
Closes #859.

I propose keeping what was initially done by @SMoraisAnsys (i.e. current behavior of `smoke-test-module`) and simply adding an output like @germa89 suggested for custom smoke testing.

That is simple and avoids having to cram several smoke test lines into one single string input.

Tested and working: https://github.com/ansys/pymapdl/actions/runs/15282156639?pr=3959

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -147,6 +147,16 @@ inputs:
     required: false
     type: string
 
+outputs:
+
+  activate-venv:
+    description: |
+      Command to activate the virtual environment created by this action.
+      This can be used in subsequent steps within the same job to perform
+      certain actions in the context of the virtual environment, such as
+      running smoke tests.
+    value: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
+
 runs:
   using: "composite"
   steps:
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -151,10 +151,9 @@ outputs:
 
   activate-venv:
     description: |
-      Command to activate the virtual environment created by this action.
-      This can be used in subsequent steps within the same job to perform
-      certain actions in the context of the virtual environment, such as
-      running smoke tests.
+      Command used to activate the virtual environment set up by this action.
+      It can be reused in later steps of the same job to execute tasks within
+      the virtual environment, such as running smoke tests.
     value: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
 
 runs:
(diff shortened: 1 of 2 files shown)


jorgepiloto (changes requested):
**This is is an open door to executing arbitrary code.** 

Either we have a fix command or none.

build-wheelhouse/action.yml
```diff
@@ -147,6 +147,16 @@ inputs:
...
     type: string
 
+outputs:
+
+  activate-venv:
+    description: |
+      Command to activate the virtual environment created by this action.
+      This can be used in subsequent steps within the same job to perform
+      certain actions in the context of the virtual environment, such as
+      running smoke tests.
```
SMoraisAnsys, line 157:
```suggestion
      Command used to activate the virtual environment set up by this action.
      It can be reused in later steps of the same job to execute tasks within
      the virtual environment, such as running smoke tests.
```

SMoraisAnsys (approved):

RobPasMue:
> The input is already implemented, and followed the SBOM action update.

As discussed offline -- we will remove the input and rework it so that no code can be injected

germa89:
@moe-ad aren't you missing the changes in the docs? in the migration for instance??

germa89:
> > @moe-ad aren't you missing the changes in the docs? in the migration for instance??
> 
> @germa89 we are continuously updating the docs in #828.

Ohh.. Sorry about that then! 😅

PR #866, opened 2025-05-28
feat: implementing basic smoke test for import

RobPasMue:
Related to #861 #859 #834

### DIFF
--- a/build-wheelhouse/smoke_test.py
+++ b/build-wheelhouse/smoke_test.py
@@ -1,0 +1,55 @@
+import importlib
+from importlib.metadata import PackageNotFoundError, distribution
+
+
+def find_module_from_dist(pkg_name: str, attr="__version__"):
+    try:
+        dist = distribution(pkg_name)
+    except PackageNotFoundError:
+        raise ImportError(f"Package '{pkg_name}' is not installed")
+
+    # Examine files in the distribution to find likely module paths
+    candidate_paths = [path for path in dist.files if path.name == "__init__.py"]
+    if not candidate_paths:
+        raise ImportError(f"No __init__.py found in package '{pkg_name}'")
+
+    # Find the shortest paths to the init files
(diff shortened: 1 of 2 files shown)


build-wheelhouse/smoke_test.py
```diff
@@ -0,0 +1,55 @@
...
+    candidate_paths.sort(key=lambda p: len(p.parts))
+
+    # Drop the paths that are longer than the shortest one
+    shortest_length = len(candidate_paths[0].parts)
+    candidate_paths = [path for path in candidate_paths if len(path.parts) == shortest_length]
+
+    # Try to import each parent package and check for the desired attribute
+    for path in candidate_paths:
+        if not str(path).endswith("__init__.py"):
+            continue
```
SMoraisAnsys, line 26:
```suggestion
```
Shouldn't be needed since `candidate_paths` is first defined as `candidate_paths = [path for path in dist.files if path.name == "__init__.py"]` right ?

build-wheelhouse/smoke_test.py
```diff
@@ -0,0 +1,55 @@
+import importlib
+from importlib.metadata import PackageNotFoundError, distribution
+
+
+def find_module_from_dist(pkg_name: str, attr="__version__"):
```
moe-ad, line 5: Maybe we shouldn't have a default value for `attr` since `attribute_name = sys.argv[2] if len(sys.argv) > 2 else "__version__"` already takes care of it?

RobPasMue: Sure, makes sense too

RobPasMue: See https://github.com/ansys/actions/pull/866/commits/66d692cfdb53a942223d7ed9bf069237fe016a80

SMoraisAnsys (approved):
LGTM, thanks @RobPasMue

PR #868, opened 2025-05-30
fix: advanced search path

Revathyvenugopal162:
This PR updates the index.html file to revert the advanced search path to the stable version URL for landing pages WRT to ansys-sphinx-theme version 1.5.0


RobPasMue:
> Awesome! Can we cherry-pick this in v9? Just in case users update the theme before we release v10.

Yes let's do this please

PR #871, opened 2025-06-02
fix: release-github python setup and sbom pattern 

SMoraisAnsys:
As title says. WIP


RobPasMue (approved):

SMoraisAnsys:
@RobPasMue I think virtual env can be associated to "python setup" :) Thanks for pointing out this issue

PR #874, opened 2025-06-03
fix: remove the usage of uv for doc-deploy-changelog

RobPasMue:
We shouldn't use uv or virtual environments in general for workflows that make use of the ``shell: python`` option. Those workflows use the global environment and they are pretty simple in terms of dependencies so...

The usage of uv only adds complexity. We should just avoid it IMO.

This requires to be added to v10

### DIFF
--- a/doc/source/changelog/874.fixed.md
+++ b/doc/source/changelog/874.fixed.md
@@ -1,0 +1,1 @@
+Remove the usage of uv for doc-deploy-changelog


doc-deploy-changelog/action.yml
```diff
@@ -203,8 +196,7 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
-        provision-uv: ${{ inputs.use-uv }}
-        prune-uv-cache: ${{ inputs.use-python-cache != 'true' }}
+        provision-uv: false
```
RobPasMue, line 199:
```suggestion
        provision-uv: false
        prune-uv-cache: ${{ inputs.use-python-cache != 'true' }}
```

Issue #882, opened 2025-06-05
Only support trusted publishers for public release actions


jorgepiloto:
To enforce the use of trusted publishers in our repositories, the `release-public-pypi` and `release-test-pypi` must only support the trusted publishers approach.

Sadly, we still need to support tokens for the `release-private-pypi` due to the limitations of this index.

MaxJPRey:
Please add a very clear error message in the action when you will deal with it.

PR #883, opened 2025-06-05
fix(release-actions): raise warning if not using trusted publishers

jorgepiloto:
Related with #882. Raises a deprecation warning if not using the trusted publishers solution when releasing a package to the public domain.


RobPasMue (approved):
LGTM!

PR #890, opened 2025-06-06
fix: release-github artifacts attestation

SMoraisAnsys:


RobPasMue (approved):
Good catch @SMoraisAnsys!

Issue #894, opened 2025-06-07
Allow uv (by default) to install pre-release versions of dependencies


RobPasMue:
If a package's dependencies also define other third-party dependencies.. and they are using pre-release versions of those third-parties... we have an issue with UV

UV doesn't allow to install the secondary-level dependencies if they are pre-release candidates AND you have not enabled the ``--prerelease=allow`` flag for uv. I have the feeling that this should always be allowed.

See example of problem here: https://github.com/ansys/pyansys/pull/972

When we first introduced pygranta==2025.2.0rc0 ... uv complained about pygranta defining a dependency on ansys-grantami-bomanalytics with a pre-release version... and thus failing the installation. See https://github.com/ansys/pyansys/actions/runs/15494472742/job/43627723690

The moment we added ansys-grantami-bomanalytics==2.3.0rc0 as a direct dependency of the metapackage, it stopped complaining about this specific package and moved on to the next problematic dependency. See https://github.com/ansys/pyansys/actions/runs/15509809000/job/43669259901?pr=972

This is a no-go. IMO we should just enable pre-releases by default with UV. 

Refs: https://docs.astral.sh/uv/concepts/resolution/#pre-release-handling and https://docs.astral.sh/uv/pip/compatibility/#pre-release-compatibility

MaxJPRey:
+1. I agree. I think we should allow this flag as I don't see yet any side effect related to it.

RobPasMue:
Not only that.. we might even have to expose something like ``uv-options`` to allow for additional flags moving forward. But let's start with adding this one by default at least

PR #895, opened 2025-06-09
fix(uv): allow the install of pre-releases

jorgepiloto:
Related with #894. Updates the `uv pip install` commands so that pre-releases can be installed when resolving dependencies.

### DIFF
--- a/doc/source/changelog/895.fixed.md
+++ b/doc/source/changelog/895.fixed.md
@@ -1,1 +1,1 @@
-Allow the install of prereleases
+Allow the install of pre-releases


RobPasMue:
Sounds good! Thanks @jorgepiloto!

MaxJPRey (approved):
LGTM.

jorgepiloto:
We found the [UV_PRERELEASE](https://docs.astral.sh/uv/reference/environment/#uv_prerelease) environment variable. Investigating its use directly. If it solves the issue in the metapackage, we'll close this to avoid projects installing pre-release candidates across its dependency tree.

RobPasMue:
Fully agreed @jorgepiloto!

> It may be better to allow projects to set any of the [uv environment variables](https://docs.astral.sh/uv/reference/environment/) rather than allowing to pass an arbitrary string that executes in the pipelines.

Issue #681, opened 2025-02-18
Unify artifacts file names


SMoraisAnsys:
Currently, when one wants to release a project using our actions, the file names are not unified.

While we use `twine` to create both the wheels and source distribution of the package, the wheelhouse are created by our `build-wheelhouse` action. However, our action leverages `-` instead of `_` as `twine` does. See below

![Image](https://github.com/user-attachments/assets/29d0cf1f-61e2-463e-99eb-6e04320953e3)

It would be great if we could standardize the naming of the artifacts. I'm not 100% sure but capitals might also need to be handled differently than we are doing now.

jorgepiloto:
The actions which generate these artifacts are the `build-wheelhouse` and `build-library` actions.

The wheelhouse is generating artifact names with hyphens `-` whereas the build library action generates artifacts with underscores `_`. The `python -m build` command, which is the one used to generate the wheel and sdist artifacts always uses `underscores`.

We need to identify which symbol is the correct one and try to unify (if possible).

RobPasMue:
Honestly, I prefer hyphens -- it's more readable. But if the standard for the build package is to use underscores, we should align to underscores throughout the actions.

jorgepiloto:
Turns out that this PEP <https://peps.python.org/pep-0427/#escaping-and-unicode> is the one replacing any non-alphanumeric ascii character with underscores.

About the wheels and sdist, there is nothing we can do. However, we can modify the name of ZIP file for the wheelhouse.

RobPasMue:
Let's align the wheelhouse we have in that case

rborsaru-ansys:
Hi guys,
I have tried to test this solution in isolation inside the actions repository, but could not make it working.
I have one idea detailed below and waiting your comments. 
As this issue was triggered by pymapdl repository can i create a branch there and build that branch with my changes in the actions repository.
What do you think?
Thank you!

RobPasMue:
There is something that I don't really like - but it's not your fault @rborsaru-ansys I think your implementation is great and solves the request on the issue.

Now we have a mix of hyphens and underscores which makes it very complicated to read honestly... I would try to move to all underscores if we could but the problem is that we reuse many of the built-in GH names (which use hyphens! e.g. ``ubuntu-latest``)...

I think we might need a better thought on this whole approach... Also, rn the SBOM is fully with hyphens, that one should get adapted too right?

rborsaru-ansys:
Hi @RobPasMue,
[...]

PR #907, opened 2025-06-12
feat: generate wheels package with underscores instead of dashes

rborsaru-ansys:
Initial commit for this branch

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -177,17 +177,17 @@ jobs:
           python-version: ${{ matrix.python }}
           working-directory: .ci/ansys-actions
 
-  package-library:
-    name: "Package library"
-    needs: wheelhouse
-    runs-on: ubuntu-22.04
-    steps:
-      - name: "Build library source and wheel artifacts"
-        uses: ansys/actions/build-library@fix/unify-artifacts-file-names
-        with:
-          library-name: ${{ env.LIBRARY_NAME }}
-          python-version: ${{ env.MAIN_PYTHON_VERSION }}
-          working-directory: .ci/ansys-actions
+  # package-library:
+  #   name: "Package library"
+  #   needs: wheelhouse
+  #   runs-on: ubuntu-22.04
+  #   steps:
+  #     - name: "Build library source and wheel artifacts"
+  #       uses: ansys/actions/build-library@fix/unify-artifacts-file-names
+  #       with:
+  #         library-name: ${{ env.LIBRARY_NAME }}
+  #         python-version: ${{ env.MAIN_PYTHON_VERSION }}
+  #         working-directory: .ci/ansys-actions
 
   tests:
     name: "Tests"


.github/workflows/ci_cd_pr.yml
```diff
@@ -233,3 +233,12 @@ jobs:
...
           trust-ansys-actions: true
+  build-library:
+    name: "Build library basic example"
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Build library source and wheel artifacts"
+        uses: ansys/actions/build-library@fix/unify-artifacts-file-names
+        with:
+          library-name: ${{ env.LIBRARY_NAME }}
+          working-directory: ".ci"
```
jorgepiloto, line 244: This expects the path to the directory containing the actual project file. Thus:

```suggestion
          working-directory: ".ci/ansys-actions"
```

jorgepiloto (commented):
Just left one suggestion for this, @rborsaru-ansys. Knowing that the build library accepts now a working directory, maybe we can revisit the `tests` job in this file.

rborsaru-ansys:
Hi @jorgepiloto and @SMoraisAnsys ,
Thank you for your comments.
Let me answer your comments.
I am trying to resolve this issue from milestone 10.1 - [Unify artifacts file names](https://github.com/ansys/actions/issues/681).
I have created a branch and now i am using this pr in order to trigger the gh action build. I am using a reference to the action on my branch, but was not sure how to trigger the pipeline without raising a pr.
I have added the build-library to this pipeline so i can test my solution. 
I have used working-folder to give the path to the test library that is set for the actions repository. 
Have used .ci/ansys-actions first but that did not work. After looking through some other repos that are using this action tried the .ci folder as well, but that failed as well, as expected. 
Not sure what should i be using in my branch to be able to build the test ansys-actions-core library.

Thank you.

jorgepiloto (approved):
Awesome, @rborsaru-ansys. Thanks for fixing this.

.github/workflows/ci_cd_pr.yml
rborsaru-ansys: Not needed anymore. Testing done in pymapdl repository.

SMoraisAnsys (changes requested):
Requesting changes to avoid this PR to be merged as long as the discussion in https://github.com/ansys/actions/issues/681 is not over.

.github/workflows/ci_cd_pr.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 env:
   MAIN_PYTHON_VERSION: '3.12'
-  LIBRARY_NAME: 'ansys-actions'
+  LIBRARY_NAME: 'ansys-actions-core'
```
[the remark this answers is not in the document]
rborsaru-ansys: From what is inside the code i would say that the name of the library is ansys-actions-core
__init__. py is here .ci/ansys-actions/src/ansys/actions/core
this is the line inside  pyproject.toml
name = "ansys-actions-core"

Is this correct? Thanks.

rborsaru-ansys:
I have fixed the build-library action code to properly use the working-directory. Now my job only fails because of the final tests and that is because of the name of the library. 
I will add the dash2underscore changes to SBOM.

PR #915, opened 2025-06-26
ci: Use dependabot cooldown for pip and github actions

SMoraisAnsys:
The cooldown feature is out of beta and should be working for github actions, see https://github.com/dependabot/dependabot-core/issues/3651#issuecomment-3006962980


MaxJPRey (approved):
LGTM.

RobPasMue (approved):

PR #916, opened 2025-06-26
docs: adding mention of PyAnsys dev guide for vulnerabilities

clatapie:
Small PR to link the action to common fixes available in the dev guide.


RobPasMue:
We will do a quick patch release with this change so that it is available in the stable docs

jorgepiloto (approved):
Thanks for adding this reference, @clatapie. Approving on my side.

PR #918, opened 2025-06-27
docs: Use single quotes for literal in GitHub action expression

Andy-Grigg:
Replace double quotes with single quotes in GitHub action expression. See https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#literals.

There's some indirection here around how jinja2 is including the yaml in the rst. If this doesn't work, the quotes can probably be omitted entirely.


RobPasMue:
That worked, thanks @Andy-Grigg!

PR #929, opened 2025-07-16
feat: strengthen testing

### LINKED ISSUE #858: Strengthen testing
This past weeks, a lot of huge changes and new features have been added and we have faced many cases where this changes broke CICD for users. The main reason for that, IMO, is that we don't have enough tests performed. It would be great if we could extend our CICD with, at least, testing the different build backend we have, i.e. `uv`, `pip` and `poetry`. Also I would like the doc-build to be tested with windows as some package(s) require the action to be run with windows.

moe-ad:
Closes #858

The following have been done so far on the PR:
- Building docs for ansys/actions on windows.
- Building docs for other repos on linux/windows. This includes only pyansys-geometry and pydpf-composites for now, mainly because their doc builds relatively fast, the build process includes generation of api/examples/etc, and they provide docker containers for their servers thereby making them easier to setup on Github hosted runners.
- There are now two small internal packages for testing some of the actions, `ansys-actions-flit` and `ansys-action-poetry` targeting `flit` and `poetry` build backends respectively. Because trusted publishing accepts only one package per workflow, two separate workflow files had to be created `ci_cd_pr_flit` and `ci_cd_pr_poetry`.
- Taking advantage of the above, some actions are now being tested against external repos as well, grouped by their build backends.
- Actions are now referenced locally, this means we don't have to manually switch the branch for testing and revert before merging as we have always done. The only exceptions are actions depending on other private internal actions, references in those cases will have to be changed manually if changes need to be tested.

The following internal changes have been made to some actions following issues (and improvement opportunities) identified through the implemented tests:
- A `checkout` option defaulting to `true` has been added `build-library`, `check-actions-security`, and `code-style`.
- How the `working-directory` input is handled in `build-wheelhouse` has been modified.
- A `upload-artifact-name-prefix` option defaulting to "documentation" has been added to `doc-build`.
- The handling of groups and extras for poetry-based projects has been modified for consistency in the `doc-build` action.

Open to discussions on what else we can explore before wrapping this PR up @jorgepiloto @SMoraisAnsys @RobPasMue @germa89

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -299,7 +299,6 @@ jobs:
           check-links: false
           checkout: false
           uploaded-artifact-name-prefix: pydpf-composites-documentation-${{ matrix.os.name }}
-          use-uv: false
 
   doc-deploy-pr:
     name: "Deploy PR documentation"


.ci/ansys-actions-flit/pyproject.toml
```diff
@@ -4,9 +4,9 @@ build-backend = "flit_core.buildapi"
 
 [project]
 # Check https://flit.readthedocs.io/en/latest/pyproject_toml.html for all available sections
-name = "ansys-actions-core"
+name = "ansys-actions-flit-core"
```
jorgepiloto, line 7: Let's name the package `ansys-actions-flit`. Otherwise, it gets too long.

jorgepiloto (changes requested):

.github/workflows/ci_cd_pr.yml
jorgepiloto, line 1: Let's remove those commented jobs in this file.

jorgepiloto (changes requested):
Just one minor request. Thanks for adding this, @moe-ad.

moe-ad:
We are still yet to conclude on testing poetry-based projects that are pulling dependencies from the private PYPI.

moe-ad:
@jorgepiloto this is similar to what I encountered initially with the `doc-deploy-changelog` action. As can be seen in [this run](https://github.com/ansys/actions/actions/runs/16717095387/job/47313200319), placing the checkout in a different directory still resulted in the same error.

Although the error eventually stopped occurring in `doc-deploy-changelog`'s case without me really doing anything, which implies it being a random error, I don't want to assume the same in this case. I suspect that the issue might be because of the checkout step in `_pr-doc-deployment`. I will investigate a bit more.

jorgepiloto:
Thanks for clarifying, @moe-ad. Since this has nothing to do with the action itself but with the CI/CD pipelines, I think it is safe to merge things.

PR #932, opened 2025-07-24
fix: Keep running when advisories are not reachable

AlejandroFernandezLuces:
## Overview

If current advisories are not reachable, you should still be able to check the vulnerabilities. This should fix the issue.


RobPasMue (approved):
LGTM! Thanks @AlejandroFernandezLuces

PR #943, opened 2025-08-04
fix: correct deleted branch references

moe-ad:
feat/extend-testing references were not changed to main before merging #929


jorgepiloto (approved):
Thanks for the quick fix, @moe-ad. Sorry for introducing this issue.

PR #944, opened 2025-08-05
test: dynamic-uses action

moe-ad:
Testing an action that allows working around not being able to use expressions and contexts in the value of the `uses` key.


moe-ad:
Closing. A new GH feature has been added to support this directly. This is being implemented in https://github.com/ansys/actions/pull/1548.

PR #947, opened 2025-08-06
ci: update ci_cd_night workflow

moe-ad:
Following the changes in #929, the package for the nightly release test has been updated to `ansys-actions-flit`. Note that it is not possible to test the release of `ansys-actions-poetry` from the same workflow file and I don't think it is necessary to split the nightly workflow into two like we did in #929.


moe-ad:
No it doesn't. Like I said, I don't think there is much value in splitting the workflow for the nightly build into two. We will have to do that if we want to be testing the poetry package as well.

jorgepiloto (approved):

PR #961, opened 2025-08-25
fix: correct spelling and improve description for changelog category determination

germa89:
As the title.

Just to clarify the usage of `use-conventional-commits`.


klmcadams (approved):
LGTM - thanks for improving this!

PR #962, opened 2025-08-25
refactor: `use-pull-request-title` instead of `use-conventional-commits`

germa89:
To be merged after #961

### DIFF
--- a/doc-changelog/action.yml
+++ b/doc-changelog/action.yml
@@ -91,7 +91,7 @@ inputs:
     default: true
     type: boolean
 
-  use-conventional-commits:
+  use-pull-request-title:
     description: |
       Use pull request title to determine the changelog category based on
       conventional commits. If ``false``, the labels in the pull request are used
@@ -99,6 +99,15 @@ inputs:
     required: false
     default: false
     type: boolean
+
+  use-conventional-commits: # TODO: Remove deprecated input in v11
+    description: |
+      [DEPRECATED] Use pull request title to determine the changelog category based on
+      conventional commits. If ``false``, the labels in the pull request are used
+      to determine the changelog category.
+    required: false
+    default: ''
+    type: string
 
   use-default-towncrier-config:
     description: |
@@ -136,6 +145,14 @@ runs:
           want to use a specific version of ``tomli``, please set the ``tomli-version``
           input accordingly. The ``toml-version`` input will be removed in v11.
 
+    - uses: ansys/actions/_logging@main
+      if: inputs.use-conventional-commits != ''
+      with:
+        level: "ERROR"
+        message: >
+          The ``use-conventional-commits`` input is deprecated and will be removed in v11.
+          Please use the ``use-pull-request-title`` input instead.
+
     - name: "Install Git and clone project"
       env:
         PR_BRANCH: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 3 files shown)


doc-changelog/action.yml
```diff
@@ -100,6 +100,15 @@ inputs:
     default: false
     type: boolean
 
+  use-conventional-commits: # TODO: Remove deprecated input in v11
```
RobPasMue, line 103: Deprecation would have to happen in v12 I believe. On v10 this should throw a warning. On v11, an error. On v12 (or even later) we can completely remove the input.

RobPasMue: I'm saying this because I don't know if your plan is to have this PR released ASAP. If that's the case it would have to go into v10... and we shouldn't break things. Thus, the warning.

doc-changelog/action.yml
```diff
@@ -136,6 +145,14 @@ runs:
           want to use a specific version of ``tomli``, please set the ``tomli-version``
           input accordingly. The ``toml-version`` input will be removed in v11.
 
+    - uses: ansys/actions/_logging@main
+      if: inputs.use-conventional-commits != ''
+      with:
+        level: "ERROR"
+        message: >
+          The ``use-conventional-commits`` input is deprecated and will be removed in v11.
+          Please use the ``use-pull-request-title`` input instead.
```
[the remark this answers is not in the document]
RobPasMue: And following these lines, for v10 we should check whether both inputs are passed or not, and asign them properly..

RobPasMue (commented):
Is this just a renaming? Just trying to understand.

SMoraisAnsys (commented):
Agreeing with @RobPasMue comment, we should log a warning in current major (`v10`), switch to an error with `v11` and remove everything in `v12`. Below is an example of what was performed when releasing v10 (warning on v9 if I'm not wrong)

```
      if: inputs.toml-version != ''
      with:
        level: "ERROR"
        message: >
          With the adoption of dependency groups in PyAnsys packages, the ``toml`` library
          is no longer used because it cannot parse ``pyproject.toml`` files where groups
          include other groups. This action now uses the ``tomli`` library instead. If you
          want to use a specific version of ``tomli``, please set the ``tomli-version``
          input accordingly. The ``toml-version`` input will be removed in v11.
```

doc-changelog/action.yml
germa89: Done.

doc-changelog/action.yml
```diff
@@ -136,6 +145,27 @@ runs:
           want to use a specific version of ``tomli``, please set the ``tomli-version``
           input accordingly. The ``toml-version`` input will be removed in v11.
 
+    - uses: ansys/actions/_logging@main
+      if: inputs.use-conventional-commits != ''
```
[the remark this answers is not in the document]
germa89: You are right! Thanks for the suggestion!

doc-changelog/action.yml
```diff
@@ -136,6 +145,27 @@ runs:
           want to use a specific version of ``tomli``, please set the ``tomli-version``
           input accordingly. The ``toml-version`` input will be removed in v11.
 
+    - uses: ansys/actions/_logging@main
```
SMoraisAnsys: Can we add a comment  here to not miss the need from switch warning -> error at some point ?
I like the `# TODO: ...` note and the use of `[DEPRECATED]` as it would ease finding such location.

PR #968, opened 2025-08-29
feat: improve check actions security

moe-ad:
WIP
Goals:
 - [x] Convert the `bash` parsing script to `python` script. Makes it cross platform, hence, users will be able to use the script locally.
 - [x] Improve the documentation.


check-actions-security/zizmor-summary.py
```diff
@@ -0,0 +1,50 @@
...
+from collections import Counter
+from subprocess import DEVNULL, PIPE
+
+FORMAT = "{:<50}{:<20}"
+PATH_RE = re.compile(r"[^ /]+/(?:[^ /]+/)?[^ ]*\.yml")
+
+
+def main():
+    high = os.getenv("HIGH_AUDIT_LEVEL", "")
+    strict = os.getenv("STRICT_AUDIT_LEVEL", "")
```
jorgepiloto, line 14: Can we make sure that this value belongs to a list of supported values? Since this affects the final command in the subprocess, the best is to make sure it is constrained as possible.

jorgepiloto (commented):
Minor request. Thanks for adding this. @moe-ad.

check-actions-security/zizmor-summary.py
moe-ad: Added in https://github.com/ansys/actions/pull/968/commits/66b55c2365f671cb23838e51484ba5e4ba1f2511.

PR #971, opened 2025-08-31
build: temporary pin `typer` dependency (transitive)

SMoraisAnsys:
As title says. Might not be needed if https://github.com/pyupio/safety/pull/779 get merged quickly

Close #970

PR #972, opened 2025-09-01
build: typer package causing breaking changes on safety

RobPasMue:
See https://github.com/ansys/ansys-tools-visualization-interface/actions/runs/17366739733/job/49294788127?pr=342 as an example

Safety hasn't updated its requirements yet - I will keep an eye out but in the meantime we should limit the `typer` package version. Release 0.17 went out just a few days ago. https://github.com/fastapi/typer/releases

> [!IMPORTANT]
> This requires a patch release of our actions.


RobPasMue:
Closing in favor of #971

PR #986, opened 2025-09-05
test: add more pre-release tests

moe-ad:
Imagine this scenario:
 - A project has made `v0.11.0a0` and `v0.11.0b0` pre-releases.
 - Then proceeds to patch `v0.10` for whatever reason.

The current logic will remove deployed documentation for the pre-releases and this shouldn't be the case. This PR fixes that.

### DIFF
--- a/doc/source/changelog/986.miscellaneous.md
+++ b/doc/source/changelog/986.miscellaneous.md
@@ -1,0 +1,1 @@
+Pre release support
--- a/doc/source/changelog/986.miscellaneous.md
+++ b/doc/source/changelog/986.miscellaneous.md
@@ -1,1 +1,1 @@
-Pre release support
+Pre-release support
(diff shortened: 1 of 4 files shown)


moe-ad:
@jorgepiloto thanks for the comment. I went back to test this scenario (see the [workflows here](https://github.com/ansys-internal/test-repo-moe/actions)) and confirmed the docs were not removed truly (for a different reason though):
- It is because of [this check](https://github.com/ansys/actions/blob/669f14aa3b1b4e2c44060ff6abd70d7fbcb7dd65/python-utils/versions.py#L185) I am already making, I totally forgot I was already doing that 😪.
- The scenario you described is a bit different, remember that `doc-deploy-stable` on `release/10` branch is still the old one, so even if my wrong diagnosis had been right, we would have still needed to do this manual removal step.

I will update this PR's purpose to only adding more unit tests.

PR #1014, opened 2025-10-01
fix: allow script to run on Windows

RobPasMue:
Path file separators are different - this will allow to work on both Windows and Linux

### DIFF
--- a/python-utils/zizmor-summary.py
+++ b/python-utils/zizmor-summary.py
@@ -6,7 +6,11 @@ from collections import Counter
 from subprocess import DEVNULL, PIPE
 
 FORMAT = "{:<50}{:<20}"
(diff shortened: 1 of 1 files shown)


python-utils/zizmor-summary.py
```diff
@@ -6,7 +6,7 @@
 from subprocess import DEVNULL, PIPE
 
 FORMAT = "{:<50}{:<20}"
-PATH_RE = re.compile(r"[^ /]+/(?:[^ /]+/)?[^ ]*\.yml")
+PATH_RE = re.compile(r"[^ /\\]+[\\/](?:[^ /\\]+[\\/])?[^ ]*\.yml" if os.name == "nt" else r"[^ /]+/(?:[^ /]+/)?[^ ]*\.yml")
```
RobPasMue, line 9:
```suggestion
PATH_RE = re.compile(
    r"[^ /\\]+[\\/](?:[^ /\\]+[\\/])?[^ ]*\.yml"
    if os.name == "nt"
    else r"[^ /]+/(?:[^ /]+/)?[^ ]*\.yml"
)
```

moe-ad (approved):
I ran the script locally and everything works fine. Thanks for this @RobPasMue.

PR #1016, opened 2025-10-01
chore: bump zizmor version to 1.14.2

### LINKED ISSUE #1015: Bump zizmor version
Bump the version of zizmor `ansys/actions/check-actions-security` installs and fix any new detected vulnerabilities. Similar to what was done in #967.

moe-ad:
Closes #1015.


RobPasMue:
Looks like things have changed indeed... This is why I wanted to avoid bumping it while teams are still adapting to use the action... We should hold off and wait for teams to adapt before we deliver them new advisories IMO.

PR #1037, opened 2025-10-20
ci: fix dependabot settings

SMoraisAnsys:
Seems like using `directories` disable the checking of .github/workflows. Note that I'm not sure if that was the case when we first started to discretize in groups, but this is currently the case... :/

Adding .github/workflows seems to be handling the problem when testing on a fork:

- before
image

- after a PR is correctly opened, cf https://github.com/SMoraisAnsys/actions/pull/4


RobPasMue:
Nice catch!

PR #1038, opened 2025-10-20
feat(actions): expose python-path output and propagate it to build-wheelhouse

germa89:
I believe it makes more sense to export the `python` path, so it can be used as an env var:

```yml
- name: "Build wheelhouse"
  uses: ansys/actions/build-wheelhouse@main
  with:
    library-name: ${{ env.LIBRARY_NAME }}
    working-directory: .ci/${{ env.LIBRARY_NAME }}
    attest-provenance: true

- name: "do something..."
  env:
    PYTHON: ${{ steps.build-wheelhouse.output.python-path }}
  run:
    python something.py
```

Potentially you could do (I will need a double check):

```yml
- name: "do something..."
  shell: python
  env:
    PYTHON: ${{ steps.build-wheelhouse.output.python-path }}
  run:
    # my python code
    from ansys.mapdl.core import launch_mapdl
```

Or **maybe** we can do something like:

```yml
- name: "do something..."
  shell: ${{ steps.build-wheelhouse.output.python-path }} {0}
  run:
    # my python code
    from ansys.mapdl.core import launch_mapdl
```

## Changes
- Add python-path output and a "Get Python path" step to the _setup-python composite action.
- Add id: setup-python to the build-wheelhouse setup step and expose python-path via steps.setup-python.outputs.python-path.

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -147,6 +147,13 @@ outputs:
       the virtual environment, such as running smoke tests.
     value: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
 
+  python-path:
+    description: |
+      Path to the Python interpreter within the virtual environment set up by this action.
+      It can be reused in later steps of the same job to execute tasks within
+      the virtual environment, such as running smoke tests.
+    value: ${{ steps.setup-python.outputs.python-path }}
+
 runs:
   using: "composite"
   steps:
@@ -158,6 +165,7 @@ runs:
         persist-credentials: false
 
     - name: "Set up Python ${{ inputs.python-version }}"
+      id: setup-python
       uses: ansys/actions/_setup-python@main
       with:
         python-version: ${{ inputs.python-version }}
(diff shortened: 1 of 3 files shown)


build-wheelhouse/action.yml
```diff
@@ -147,6 +147,13 @@ outputs:
       the virtual environment, such as running smoke tests.
     value: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
 
+  python-path:
```
SMoraisAnsys: Shouldn't you be retrieving the virtual env instead of the python installation ? If yes, I think that it would make more sense to expose `steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV` instead.

jorgepiloto (commented):
The actions were designed to be unique jobs. I have the feeling that the more flexibility and options we add, the more challenging it is to maintain them.

germa89:
> I fully agree with @jorgepiloto on this one... also, we shouldn't be performing extra operations after the wheelhouse has been built

I agree... but I think we mentioned that we might need to run some custom smoke tests after the build. That is why the other output `activate_venv` was implemented. But that option (export commands to activate a venv) is a bit non-standard, hence I was proposing this solution.

RobPasMue:
Why not activate the venv and run python normally then?

PR #1045, opened 2025-10-28
feat: improve-doc-deploy-pr

### LINKED ISSUE #942: Simplify deployed doc removal logic in doc-deploy-pr action
Currently, doc-deploy-pr action automatically removes the deployed documentation once the PR is closed/merged. This removal step depends on a "closed" event that triggers the workflow containing the action when the PR closes. This requires adding more logic to the workflow to prevent running other actions besides doc-deploy-pr during the "closed" event (as can be seen [here](https://github.com/ansys/actions/pull/802/files) for example).

This is not clean and adds some noise. The main reason for needing the closed event is the ability to leverage the `github` context for easy retrieval of information about the PRs whose documentation require removal. We can leverage the `PyGithub` package to retrieve this information. This implies the following:
  - Unifying the deployment and removal sections of the action into one.
  - In this unified action, a step will use the `PyGithub` package to identify closed pull-requests among those whose documentation are still present in the `pull` directory.
  - This means when a PR is closed, it's documentation removal will occur when the action is triggered from a different PR.

moe-ad:
Closes #942.

Simplifies deployed documentation removal for PRs by eliminating the need for a "closed" event to trigger the removal. Changes:
- ~Through a python script, `PyGithub` is used to figure out which documentation needs to be removed and the removal itself is done via this script.~ `gh-cli` is used to determine the pr documentation requiring removal.
- Since this removal still needs to be triggered in some way, we need to shove it into a documentation action that frequently runs in PyAnsys pipelines (but not too frequent). I chose `doc-deploy-dev` for this. Another candidate would be `_pr-doc-deployment` but the potential -ves of that are:
  - Too frequent. 10 opened PRs (with docs-deployed) -> 10 locations where API calls are being made. I am not sure how quickly that will shrink the github REST API rate limits
  - Scenario: `Repo has a bunch of pr docs deployed` -> `Repo deactivates doc-deploy-pr for whatever reason`. In this scenario, deployed documentation will be left hanging and will never get the chance to be removed.
  - It may impact existing `_pr-doc-deployment` logic, especially the one that has to do with determining when `maximum-pr-doc-deployments` have been reached.
- Updated relevant documentation.
- Updated the migration guide in preparation for v10.2 release (I didn't want to open a separate PR for this... 😫 ).
- Updated `_pr-doc-clean` to just output an informational message about the new changes.

Ideally, things should keep working normally (even for repos that leave their current `doc-deploy-pr` setup as is), as long as they are using compatible versions of `doc-deploy-pr` and `doc-deploy-dev` which is usually the case when they do a version bump.

Now I need to test some more, but you can already start reviewing and sharing your thoughts... :)

### DIFF
--- a/doc/source/migrations/docs-deploy-pr-setup.rst
+++ b/doc/source/migrations/docs-deploy-pr-setup.rst
@@ -49,7 +49,7 @@ Two setup options are provided based on the desired level of control:
           if: github.event.action != 'closed'
           runs-on: ubuntu-latest
           steps:
-            - uses: ansys/actions/doc-build@v10
+            - uses: ansys/actions/doc-build@v10.1
               with:
                 python-version: ${{ env.MAIN_PYTHON_VERSION }}
 
@@ -59,7 +59,7 @@ Two setup options are provided based on the desired level of control:
           needs: doc-build
           if: always() && (needs.doc-build.result == 'success' || needs.doc-build.result == 'skipped')
           steps:
-            - uses: ansys/actions/doc-deploy-pr@v10
+            - uses: ansys/actions/doc-deploy-pr@v10.1
               with:
                 cname: ${{ env.DOCUMENTATION_CNAME }}
                 token: ${{ secrets.GITHUB_TOKEN }}
@@ -81,7 +81,7 @@ Two setup options are provided based on the desired level of control:
           name: "Doc build"
           runs-on: ubuntu-latest
           steps:
-            - uses: ansys/actions/doc-build@v10
+            - uses: ansys/actions/doc-build@v10.2
               with:
                 python-version: ${{ env.MAIN_PYTHON_VERSION }}
 
@@ -90,7 +90,7 @@ Two setup options are provided based on the desired level of control:
           runs-on: ubuntu-latest
           needs: doc-build
           steps:
-            - uses: ansys/actions/doc-deploy-pr@v10
+            - uses: ansys/actions/doc-deploy-pr@v10.2
               with:
                 cname: ${{ env.DOCUMENTATION_CNAME }}
                 token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


python-utils/remove_closed_pr_deployed_doc.py
RobPasMue: I've got conflictive opinions about all these "minimal functions"... you are using most of them just once. And it is making it difficult to follow the inputs/outputs. Wouldn't it make more sense to just have a script (without additional 2-3 line methods) and inline comments explaining the operations you are doing?

moe-ad: I have changed things a bit. Have a look, I can still squeeze logic together some more if required.

RobPasMue (approved):
LGTM!

python-utils/remove_closed_pr_deployed_doc.py
jorgepiloto: I think we could try to leverage the `gh-cli` here. It comes already installed in runners and protects secrets. 

This simple script should allow us to identify the state of a pull-request:

```bash
for dir in "pull/*; do

    # Skip if not a directory
    [ -d "${dir}" ] || continue

    # Extract PR number from directory name
    pr_number=$(basename "${dir}")
    
    # Check PR state using gh-cli
    state=$(gh pr view "${pr_number}" --json state --jq .state 2>/dev/null)
    
    # If PR does not exist, skip
    if [ -z "${state}" ]; then
        echo "PR #${pr_number}: not found, skipping."
        continue
    fi

    echo "PR #${pr_number}: $state"

    # Remove directory if PR is CLOSED or MERGED
    if [[ "${state}" == "CLOSED" || "${state}" == "MERGED" ]]; then
        echo "Removing directory $dir..."
        rm -rf "${dir}"
    fi
done
```

The gh-cli is very powerful and I bet we can leverage it way more across our actions.

jorgepiloto (changes requested):
I left a suggestion on migrating to gh-cli for implementing this. It avoids using Python.

moe-ad:
Tests after implementing `gh-cli` approach:

1. Behaviour when there are pr deployments to remove: https://github.com/ansys-internal/test-repo-moe/actions/runs/19231403485/job/54984352969
2. Behaviour when there aren't pr deployments to remove: https://github.com/ansys-internal/test-repo-moe/actions/runs/19236192602/job/54987952000

RobPasMue (approved):

jorgepiloto (approved):
Thanks for implementing this, @moe-ad. Merging.

PR #1056, opened 2025-11-10
ci: add job to check dependabot configuration coverage

### LINKED ISSUE #843: Add workflow to check nothing is missing with respect to dependabot's configuration
Following https://github.com/ansys/actions/pull/841#discussion_r2102351646, we should have a workflow dedicated to dependabot and which checks that:
- all files are tracked by dependabot (i.e. listing directories correctly);
- all actions are associated to a meaningful group (we missed the uv action recently).

moe-ad:
Closes #843.

@SMoraisAnsys @RobPasMue.
Do you have some ideas about how point 2 of the issue can be handled automatically? Seems like a manual task to me.

### DIFF
--- a/.github/workflows/ci_cd_night.yml
+++ b/.github/workflows/ci_cd_night.yml
@@ -3,6 +3,7 @@ name: night
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch:
 
 env:
   MAIN_PYTHON_VERSION: '3.12'


moe-ad:
FYI I left the failing workflow as is so reviewers can see how the action currently works. I will fix it once this has been approved for merging.

.github/workflows/dependabot_coverage.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: "Check third-party actions are in dependabot.yml"
+        shell: bash
+        run: |
+          echo "Checking third-party actions coverage in dependabot.yml"
+          echo "========================================================"
+          echo ""
+
+          # Get all third-party actions from workflows and composite actions
+          ACTIONS=$(
+              (find . -maxdepth 2 -type f -name "action.yml" && \
```
RobPasMue: Can we avoid setting a maxdepth or is this intended?

RobPasMue:
> I didn't know this was to be offered as a reusable action for other repos to consume.

I am not saying it should be consumed by other repos, not at all in fact.

What I am suggesting is that the same workflow you implemented should also open a PR on the ansys/actions repo and update the dependabot.yml file to include the missing actions. We shouldn't just get informed that some actions are missing. We should remediate the problem

.github/workflows/dependabot_coverage.yml
moe-ad: Intended. Since all our composite actions are in `./<action-name>/action.yml`, there is no need to search further. What do you think?

RobPasMue: Okay... Problem I see is if we ever decide to build "nested" actions... they won't be caught by default. And I don't think we lose too much time by not looking at all depths anyway (this task is not time/perf critical)

RobPasMue:
> @RobPasMue @SMoraisAnsys The challenge with opening a PR is related to my opening question on this PR. The problem of classifying an action under a meaningful group seems seems like a manual task. I guess an issue should be opened instead?

Since this part is not working:

https://github.com/ansys/actions/blob/bc44fb8ec2879b27f14d2ee5910e7b94ce72ae7e/.github/dependabot.yml#L110-L115

I'd recommend removing the asterisk at large and adding the individual actions found to this group. We can move them manually later if desired

RobPasMue:
I would run this on a scheduled basis as well @moe-ad - not only on workflow dispatch

RobPasMue:
> What about moving this to the `ci_cd_night.yml`? Just to reduce the number of workflow files to `pr`, `night, and `release`.

Works for me

PR #1078, opened 2025-12-02
chore: bump main branch version

moe-ad:
As the title states. I just released `v10.2`.

### DIFF
--- a/doc/source/changelog/1078.maintenance.md
+++ b/doc/source/changelog/1078.maintenance.md
@@ -1,1 +1,1 @@
-Chore: bump main branch version
+Bump main branch version


.ci/ansys-actions-flit/pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "flit_core.buildapi"
 [project]
 # Check https://flit.readthedocs.io/en/latest/pyproject_toml.html for all available sections
 name = "ansys-actions-flit"
-version = "10.1.dev0"
+version = "10.2.dev0"
```
SMoraisAnsys: Since the latest version is `10.2.0`, shouldn't this version be `10.3.dev0` ?

moe-ad: My bad, thanks for catching that. I must be growing old.

PR #1080, opened 2025-12-02
ci: fix if statement logic in doc-changelog

SMoraisAnsys:
Fix recent issue faced when using the doc-changelog action with labels.

### DIFF
--- a/doc/source/changelog/1080.maintenance.md
+++ b/doc/source/changelog/1080.maintenance.md
@@ -1,1 +1,1 @@
-Ci: fix if statement logic
+Ci: fix if statement logic in doc-changelog


klmcadams:
I tested it with ``use-pull-request-title``, ``use-conventional-commits`` & without either of them. The doc-changelog generation worked. I think there may be something up with the fragment content though. The info in the changelog fragment is "Test: doc changelog test" when I think it should just be "Doc changelog test." Maybe we can fix that in a different PR?

Also thanks for debugging this so quickly @SMoraisAnsys :)

jorgepiloto (approved):
Many thanks for this quick-fix, @SMoraisAnsys.

PR #1111, opened 2026-01-14
feat: check version match

### LINKED ISSUE #671: PyPI release does not fail when it already exists.
So... yesterday I tried to do a "quick" release (disclaimer, it is never quick on PyMAPDL).

And I forgot to update the `pyproject.toml` to the new version (yeeeeahhh.. I know...), however the tag was correct. Hence I had tag with `v0.69.2` and the `pyproject.toml` with `v0.69.1`.

Surprisingly, the CICD didn't fail, hence I didn't realise (thanks @greschd for letting me know).

Shouldn't the workflow fail? It is showing a warning though.

Image

One could argue that it is ok in case you want to attach new files after the release is done. Is this something we want to support??

Reference: https://github.com/ansys/pymapdl/actions/runs/12891417279/job/35962119607#step:3:994

jorgepiloto:
Fix #671 by making sure artifacts version matches the git tag that triggered a release.


_release-pypi/action.yml
```diff
@@ -148,6 +148,52 @@ runs:
         name: ${{ inputs.library-name }}-artifacts
         path: ${{ inputs.library-name }}-artifacts
 
+    - name: "Verify tag and artifacts version agree"
```
MaxJPRey:
```suggestion
    - name: "Verify tag and artifacts version match"
```

moe-ad:
@RobPasMue @SMoraisAnsys 
Finalized things here as well, taking into consideration the changes in #1112.
Please give that PR and this one a final look.

PR #1112, opened 2026-01-14
feat: update wheelhouse artifacts names

### LINKED ISSUE #681: Unify artifacts file names

jorgepiloto:
There is no convention for wheelhouses as opposite to Python distribution artifacts. Knowing that these last impose underscores in the name of the projects but use hyphens for separating metadata, I decided to follow the same approach with the wheelhouse. Fixes #681.


RobPasMue (commented):
We should also check that other actions like the GitHub release are not impacted by this change. I'm not sure what naming convention it was expecting for wheelhouse artifacts.

build-wheelhouse/action.yml
```diff
@@ -389,22 +401,28 @@ runs:
...
 
     - name: "Upload the compressed wheelhouse"
       uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+      env:
+        LIBRARY_NAME: ${{ steps.sanitize.outputs.sanitized_library_name }}-
       with:
-        name: ${{ inputs.library-name }}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}
-        path: ${{ inputs.library-name }}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}.zip
+        name: "${LIBRARY_NAME}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}"
+        path: "${LIBRARY_NAME}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}.zip"
```
moe-ad:
```suggestion
        name: "${{ steps.sanitize.outputs.sanitized_library_name }}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}"
        path: "${{ steps.sanitize.outputs.sanitized_library_name }}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}.zip"
```

SMoraisAnsys (changes requested):
@moe-ad Could you perform a test from an external repo to ensure that nothing is missed ? This would require to point to the PR's branch. Still thinking that this is fine but calling a request for changes to ensure that someone does take the time to test before merging.

PR #1116, opened 2026-01-19
fix: hk-* actions migrated to using requirements file

### LINKED ISSUE #1115: hk-package-clean-untagged fails with SyntaxError
### 🐞 Description of the bug

The has started failing with the following exception:

```
[53](https://github.com/ansys/pyacp/actions/runs/21104761542/job/60694221384#step:2:262)
Traceback (most recent call last):
  File "/home/runner/work/pyacp/pyacp/clean-untagged.py", line 5, in <module>
    from ghapi.all import GhApi
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/ghapi/all.py", line 1, in <module>
    from .core import *
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/ghapi/core.py", line 9, in <module>
    from fastcore.all import *
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/fastcore/all.py", line 9, in <module>
    from .script import *
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/fastcore/script.py", line 15, in <module>
    from .docments import docments
  File "/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/fastcore/docments.py", line 359
    sig_str = f"def {get_name(self.obj)}(\n    {'\n    '.join(lines)}\n{self._ret_str}"
                                                                                       ^
SyntaxError: f-string expression part cannot include a backslash
```

Failing run: https://github.com/ansys/pyacp/actions/runs/21104761542
Last successful run (Jan 17th): https://github.com/ansys/pyacp/actions/runs/21086995784

Looking at the action source, neither `ghapi` nor its dependencies are pinned, so this is likely caused by a bad unintentional dependency upgrade.

RobPasMue:
Closes #1115 

@ansys/pyansys-core - should we consider removing the Python version for actions such as this one? This was proposed by @greschd in #1115 as well. I haven't got a strong opinion on this: I agree with @greschd that it's not needed and users shouldn't have to worry about it, but it also provided a way for a quick fix in case somebody needed to "solve" this issue right away.

### DIFF
--- a/hk-package-clean-except/action.yml
+++ b/hk-package-clean-except/action.yml
@@ -71,7 +71,7 @@ runs:
     - uses: ansys/actions/_logging@main
       if: inputs.python-version != '' || inputs.use-uv != ''
       with:
-        level: "ERROR"
+        level: "WARNING"
         message: >
           'The "python-version" and "use-uv" inputs are deprecated and will be
           removed in future releases. This action uses a specific Python version set
--- a/hk-package-clean-untagged/action.yml
+++ b/hk-package-clean-untagged/action.yml
@@ -61,7 +61,7 @@ runs:
     - uses: ansys/actions/_logging@main
       if: inputs.python-version != '' || inputs.use-uv != ''
       with:
-        level: "ERROR"
+        level: "WARNING"
         message: >
           'The "python-version" and "use-uv" inputs are deprecated and will be
           removed in future releases. This action uses a specific Python version set


hk-package-clean-except/action.yml
```diff
@@ -76,14 +76,15 @@ runs:
         provision-uv: ${{ inputs.use-uv }}
         prune-uv-cache: true
 
-    - name: "Install ghapi"
+    - name: "Install required packages"
       shell: bash
       env:
+        ACTION_PATH: ${{ github.action_path }}
         INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --system' || 'python -m pip install' }}
       run: |
```
RobPasMue:
```suggestion
```

greschd:
> I think it should be left exposed exactly for the reason you mentioned i.e. fixing convenience.

There are numerous ways the action could break. Only a small number of them are fixed by changing the Python version.

To me, the Python version used is an internal of the action, and shouldn't be exposed.

The downside of having a user-overridden Python version is that:
- it's less well-tested compared to the default version; this could hide subtle bugs
- it increases maintenance overhead on the `actions` repository if it needs to implicitly support multiple Python versions / dropping Python versions means a breaking change
- inevitably, the user will eventually need to update or remove the override again

PR #1156, opened 2026-02-09
feat: use prek as pre-commit alternative

### LINKED ISSUE #1173: Add `prek` as a `pre-commit` alternative
Providing the option for users to run `ansys/actions/code-style` using `prek`, a drop-in replacement for `pre-commit`.

SMoraisAnsys:
Optional use of [prek](https://prek.j178.dev/) through a new input. This tool is working faster (install & run) than pre-commit.
**On top of that change, I also removed the (unnecessary) hook installation from the `pre-commit` approach**

~I tested it locally for fun and the difference was noticeable. Here is the use of ansys/actions/code-style on PyAEDT using :~
~- the [main branch](https://github.com/ansys/pyaedt/actions/runs/21827078470/job/62975904337) (1min21s)~
~- the [current branch](https://github.com/ansys/pyaedt/actions/runs/21827078470/job/62975904381) (43s)~

*Edit after modification*

I tested it locally for fun and the difference was noticeable. Here is the use of ansys/actions/code-style on PyAEDT using:
- [pre-commit](https://github.com/ansys/pyaedt/actions/runs/21862059941/job/63093519571) (1min20s)
- [prek](https://github.com/ansys/pyaedt/actions/runs/21862059941/job/63093519668) (57s)

Note that:
- using `prek` shouldn't be disruptive for any user, see [already using pre-commit](https://prek.j178.dev/quickstart/#already-using-pre-commit)
- users could use multiple pre-commit-config.yaml files (but lose the pre-commit.ci integration once public) an example of a workflow to be used in that case could be find [here](https://github.com/fastapi/fastapi/blob/master/.github/workflows/pre-commit.yml)

Closes #1173.

### DIFF
--- a/hk-package-clean-except/requirements.txt
+++ b/hk-package-clean-except/requirements.txt
@@ -1,3 +1,3 @@
-fastcore==1.12.12
+fastcore==1.12.13
 ghapi==1.0.10
 packaging==26.0


code-style/action.yml
```diff
@@ -57,13 +58,28 @@ inputs:
     default: true
     type: boolean
 
+  use-prek:
+    description: |
+      Whether to use prek instead of pre-commit. If ``true``, prek will be used
+      (faster alternative). If ``false``, pre-commit will be used (default behavior).
+    required: false
+    default: false
+    type: boolean
```
RobPasMue: Should we change the default to true here? Which one should be the default engine?

SMoraisAnsys:
@RobPasMue Can you have a quick look to the latest changes so that we can merge this PR if no one opposes ?

PR #1159, opened 2026-02-17
ci: use geometry latest images rather than fixing the usage of 25R2

RobPasMue:
As title says.. Otherwise CI/CD will continue to fail as shared here https://github.com/ansys/actions/actions/runs/22067805128/job/63853386883?pr=1158

### DIFF
--- a/doc/source/changelog/1159.maintenance.md
+++ b/doc/source/changelog/1159.maintenance.md
@@ -1,1 +1,1 @@
-Use geometry latest images rather than hardcode 25R2
+Use geometry latest images rather than fixing the usage of 25R2


.github/workflows/ci_cd_pr.yml
```diff
@@ -170,9 +170,9 @@ jobs:
       matrix:
         os:
           - name: windows-latest
-            container-image-tag: ghcr.io/ansys/geometry:core-windows-25.2
+            container-image-tag: ghcr.io/ansys/geometry:core-windows-latest
           - name: ubuntu-latest
-            container-image-tag: ghcr.io/ansys/geometry:core-linux-25.2
+            container-image-tag: ghcr.io/ansys/geometry:core-linux-latest
```
moe-ad:
```suggestion
            container-image-tag: ghcr.io/ansys/geometry:core-linux-latest-26.1
```

RobPasMue:
> @RobPasMue The reason for testing against a stable container version it is to limit external impacts. Latest container builds might not be as stable and pulling that might start impacting the actions pipelines more frequently.
> 
> Can we do 261 instead?

As one of the main maintainers of the PyAnsys Geometry repository, I have to disagree on that one..

The ``*-core-latest`` images are verified against the entire test suite before they are tagged as such. By default the repository receives ``*-core-latest-unstable`` coming from internal sources. If the entire test suite passes, then and only then, is the image promoted to ``*-core-latest``.

In fact, using a "version specific" image is not ensured to be stable. Take ``*-core-271`` for example.. that tag gets pushed every time a commit is done inside the server team. And I can assure you that you will hit problems if you use that one 😄

On the other hand, if you use an old version (such as the 26R1 image), I can also assure you that the documentation might fail to build, as it is happening currently.. because we are constantly adding new examples and some of them might require features from 27R1.

So, all that being said, I suggest we stick to the ``*-core-latest`` image.

PR #1166, opened 2026-02-20
chore: revert to original pip licenses

### LINKED ISSUE #1047: Switch back to original pip-licenses ?
This is an open topic, feel free to disapprove.

Seems like the repo https://github.com/raimon49/pip-licenses is back alive. Didn't have the time to check the changes but from the comments in https://github.com/raimon49/pip-licenses/issues/227, it seems that the latest version could be something we could leverage.

Given the time we have had to update the fork we created, it might be a good option to fall back to the original repo.

moe-ad:
Closes #1047.

I ran tests in the Metapackage against Python 3.9 through 3.13 and the tests pass (see ansys/pyansys#1205).

Since `pip-licenses` officially supports Python 3.9, should we still be enforcing 3.10 as the minimum Python in `ansys/check-licenses` or it is ok to stop doing that like I have already done?
@RobPasMue @SMoraisAnsys @jorgepiloto

### DIFF
--- a/doc/source/changelog/1175.dependencies.md
+++ b/doc/source/changelog/1175.dependencies.md
@@ -1,0 +1,1 @@
+Bump ansys-sphinx-theme from 1.7.0 to 1.7.1 in /requirements
--- a/requirements/requirements_doc.txt
+++ b/requirements/requirements_doc.txt
@@ -1,4 +1,4 @@
-ansys-sphinx-theme==1.7.0
+ansys-sphinx-theme==1.7.1
 jupyter==1.1.1
 pyyaml==6.0.3
 Sphinx==9.1.0


check-licenses/action.yml
```diff
@@ -271,22 +259,15 @@ runs:
...
+    - name: "Update pip and install pip-licenses library"
       shell: bash
       env:
         ACTIVATE_VENV: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
-        INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --no-cache' || 'python -m pip install --no-cache-dir' }}
+        INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python' || 'python -m pip install' }}
       run: |
         ${ACTIVATE_VENV}
-        $INSTALL_COMMAND external/pip-licenses
+        $INSTALL_COMMAND -U pip
```
dipinknair: I think we can remove pip upgrade here since its already done at line 260

SMoraisAnsys (approved):
Thanks for the changes @moe-ad It seems that the repo is indeed back alive and that the changes performed aren't sketchy. Could you have a look at my comment ?

check-licenses/action.yml
```diff
@@ -176,8 +176,15 @@ runs:
       with:
         level: "ERROR"
         message: >
-          The check-licenses action leverages Ansys fork of pip-licenses.
-          This package requires Python 3.10 or higher.
+          ansys/actions/check-licenses requires Python 3.10 or higher. The provided Python version is ${{ inputs.python-version }}.
```
SMoraisAnsys: Just to confirm a statement, I checked and `prettyable==3.17.0` requires Python 3.10+.
Since we'll use the requirements.txt file, we can use that statement :)

check-licenses/action.yml
```diff
@@ -176,8 +176,15 @@ runs:
...
+
+    - name: Announce the switch to original pip-licenses package
+      uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Maintenance of the original pip-licenses package has resumed, and the Ansys fork is being discontinued.
+          Therefore, the check-licenses action has been updated to use the original pip-licenses package.
```
RobPasMue: Let's hope somebody actually reads this.. if not, well, we tried our best

check-licenses/action.yml
```diff
@@ -266,31 +273,15 @@ runs:
...
-
-    - name: Install ansys/pip-licenses main branch
-      shell: bash
-      env:
-        ACTIVATE_VENV: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
-        INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --no-cache' || 'python -m pip install --no-cache-dir' }}
-      run: |
-        ${ACTIVATE_VENV}
-        $INSTALL_COMMAND external/pip-licenses
+        $INSTALL_COMMAND pip-licenses
```
RobPasMue:
```suggestion
        $INSTALL_COMMAND -r ${GITHUB_ACTION_PATH}/requirements.txt
```

Since you added a requirements file you need to install from it!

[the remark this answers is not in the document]
RobPasMue: Same comment - Why not leverage directly ``${GITHUB_ACTION_PATH}``?

PR #1169, opened 2026-02-23
fix: check vulnerability to correctly fail on bandit / safety issue

SMoraisAnsys:
By adding the skip on exit on steps "Run safety and bandit" and "Run safety advisory checks" in #1134 we are now in a situation where even if an issue is detected, the action won't fail but log a message that something went wrong.

This little change ensure that the action fails while keeping the ability to upload the artifacts (if required). See
- expected failure of the job [when a bandit issue is detected](https://github.com/ansys/pyedb/actions/runs/22318990577/job/64571785967)
- passing job [despite the bandit issue detected](https://github.com/ansys/pyedb/actions/runs/22319163647/job/64572396080?pr=1843)
image

### DIFF
--- a/doc/source/changelog/1169.fixed.md
+++ b/doc/source/changelog/1169.fixed.md
@@ -1,1 +1,1 @@
-Check vulnerability
+Check vulnerability to correctly fail on bandit / safety issue


MaxJPRey (approved):
LGTM.
Can you just add more info in the PR title 😄 ?

moe-ad:
Thanks for the fix @SMoraisAnsys. I will make a patch release as soon as this is merged.

PR #1219, opened 2026-03-20
feat: install python with uv

SMoraisAnsys:
As title says. Could be very usefull when one wants to use our actions in containers where python provided by `actions/setup-python` does not work directly.

The draft still needs to be updated to point to the main branch.

Associated to #1218


RobPasMue (changes requested):
I've go my doubts regarding this... why not reuse the "use-uv" input? If we use uv it should be a consistent usage IMO..

Adding optional inputs/flags is just going to introduce confusion

moe-ad:
If we can rename `use-uv` to something that unambiguously reflects that it only installs packages (maybe `use-uv-for-package-installation` ?), I personally wouldn't mind adding another input for python installation.

If not, we probably shouldn't expose another input . We can instead document this as an additional behavior for the `use-uv` option in the migration guide to give the change more visibility. What do you think @SMoraisAnsys?

SMoraisAnsys:
> If we can rename `use-uv` to something that unambiguously reflects that it only installs packages (maybe `use-uv-for-package-installation` ?), I personally wouldn't mind adding another input for python installation.
> 
> If not, we probably shouldn't expose another input . We can instead document this as an additional behavior for the `use-uv` option in the migration guide to give the change more visibility. What do you think @SMoraisAnsys?

Fully moving to `uv` (python provider & package installation) might have side effects on certain workflows. Typically, I'm thinking about this issue https://github.com/astral-sh/python-build-standalone/issues/596 but others might exist. We could probably bring that topic to one of our meeting. Leaving this PR as draft in the meantime.

jorgepiloto:
Closing this as discussed privately

PR #1223, opened 2026-03-23
fix: Do not delete changelog fragments on pre-release

### LINKED ISSUE #1087: Update changelog actions to include pre-release changelog entries in the stable release changelog.
Currently, if a project goes through a pre-release cycle, the changelog action treats the pre-releases like normal releases. This means, for example, changelog entries already in pre-releases `1.1.0a0`, `1.1.0a1`, `...`, `1.1.0rc0`, `...` will not show up in the entries of stable release `1.1.0`. A user consulting the release notes will expect all changes in `1.1.0` to be included in its own changelog, even if they already show up in pre-releases' changelogs.

A concrete example can be found in the release notes of `ansys/actions` [v10.1.0a0](https://actions.docs.ansys.com/version/dev/changelog.html#a0-september-01-2025) and [v10.1.0](https://actions.docs.ansys.com/version/dev/changelog.html#september-11-2025), where `v10.1.0` changelog did not include that of `v10.1.0a0`.

At a very high level, I guess we need to add extra steps to `doc-deploy-changelog` that does the ffl when making a stable release:
- parses `doc/source/changelog.rst` to determine if pre-releases corresponding to the stable release being made exist.
- include their (pre-releases) changelog entries in that of the stable release.

See also https://github.com/ansys/pygranta/issues/113 (the section on release notes)

### LINKED ISSUE #1137: Changelog action shouldn't delete fragments on pre-release

The changelog action currently deletes all consolidated changelog fragments on creating a pre-release changelog. This isn't great, because it means if one does a release candidate followed immediately by a release, there are no changes included in the release because they were all previously released in the candidate. Since most users won't look at the changes for a pre-release build, these are essentially lost.

It would be better if they either weren't deleted for a pre-release build (each pre-release for a given release would contain all changelog fragments up until that point) or were stashed in a separate folder.

A workaround currently used by PyGranta packages is to decline the PR against main and to revert the automatic commit following pre-release publishing on the release branch.

_No response_

_No response_

klmcadams:
As title says. Closes #1087 and Closes #1137. Had to also add pip install tomlkit to actions


RobPasMue (changes requested):
Why is tomlkit needed here? I don't see its usage I believe.

Apart from that, LGTM

SMoraisAnsys (approved):
LGTM except for @RobPasMue's comment. From what I've seen online, we shouldn't be requiring this extra requirement. Can you clarify the need @klmcadams ?

RobPasMue (approved):

moe-ad (approved):
I added https://github.com/ansys/actions/pull/1223/commits/ca08b6892d0cf2e0027854508937c20019d4c58f, looks good to me.

SMoraisAnsys (changes requested):
Some minor comments as the current message and implementation do not seem to align. If it's on purpose, I'd suggest to consider updating it to avoid the "die and retry" approach and to make the changes homogeneous in the repo

SMoraisAnsys (changes requested):
Requesting changes to follow up on my comment.

doc-deploy-changelog/action.yml
```diff
@@ -190,6 +197,18 @@ inputs:
...
+      if: inputs.toml-version != '' || inputs.tomli-version != ''
+      with:
+        level: "WARNING"
+        message: >
+          The ``toml-version`` and ``tomli-version`` inputs are deprecated. With the adoption of
+          dependency groups in PyAnsys packages, the ``toml`` and ``tomli`` libraries are no
+          longer used because they cannot parse ``pyproject.toml`` files where groups include
+          other groups. This action now uses the ``tomlkit`` library instead. Please use the
+          ``tomlkit-version`` input accordingly. The ``toml-version`` and ``tomli-version``
+          inputs will be removed in v11 and v12, respectively.
```
SMoraisAnsys:
```suggestion
          The ``toml-version`` and ``tomli-version`` inputs are deprecated. With the adoption
          of dependency groups in PyAnsys packages, the ``toml`` library is no longer used
          because it cannot parse ``pyproject.toml`` files where groups include other groups.
          While ``tomli`` supports parsing dependency groups, it can lose a lot of information
          during parsing, including formatting and comments, which can negatively impact
          the behavior of some of our actions. To ensure consistent read/write operations and
          preserve this information, this action now uses the ``tomlkit`` library instead. If you
          want to use a specific version of ``tomlkit``, please set the ``tomlkit-version`` input
          accordingly.
```
The content isn't correct, `tomli` can parse the `pyproject.toml` file and retrieve the dependency group information but the "write" operation makes us lose a lot of information. This isn't an impact for this action as it is only reading but, to ensure consistency among our actions, the migration to `tomlkit` is performed.

SMoraisAnsys (changes requested):

MaxJPRey (approved):

moe-ad:
@SMoraisAnsys can you give this one final look?

SMoraisAnsys (approved):

PR #1233, opened 2026-03-25
fix: install MiKTeX at user level

SMoraisAnsys:
# Description

This PR updates the installation of MiKTeX through Chocolatey (Windows only) by installing MiKTeX at user level. The reason for such changement is that, in self-hosted runners, one would end up reinstall MiKTeX at every run of the doc-build action. However, with the recent changes in Chocolatey, one would then hit the following error in CI:

image

After digging into Chocolatey it seems that, when an installation is already existing, the script used to install MiKTeX behaves differently. Indeed, it checks the registry path where MiKTeX is installed and checks if it's under `HKEY_CURRENT_USER`. If it is, it sets a variable `$admin` to `''`, meaning no --admin flag will be used in subsequent commands, otherwise it sets `$admin` to `'--admin'`, so the next MiKTeX commands will run with administrative privileges. This leads to an error when one uses Chocolatey `2.7.0` which was not caught with `2.5.0`.

On top of being cleaner, by modifying the installation to the user level, the error is avoided.

Here is a workflow where the modification worked https://github.com/ansys/pyaedt-toolkits-common/actions/runs/23654090294/job/68908948861
- [Build GH](https://github.com/ansys/pyaedt-toolkits-common/actions/runs/23654090294/job/68906475000) is a call to doc-build on a GitHub runner with `windows-latest`
- [Build SH](https://github.com/ansys/pyaedt-toolkits-common/actions/runs/23654090294/job/68906475245) is a call to doc-build on a self-hosted runner without choco installed
- [Build SH BIS](https://github.com/ansys/pyaedt-toolkits-common/actions/runs/23654090294/job/68908948861)  is a call to doc-build on the same runner as Build SH which leads to not reinstalling Chocolatey and MiKTeX, and does not trigger an error.

# Related issues

Close #1235


doc/source/migrations/index.rst
```diff
@@ -410,7 +410,7 @@ Version ``v6``
   - Inclusion of `canonical` link tags in all HTML files for SEO purposes
 
 - Extend ``ansys/actions/doc-build`` to be able to run in Windows runners.
-  To buid the documentation in a Windows runner, we install ``Chocolatey`` and ``Miktex``.
+  To buid the documentation in a Windows runner, we install ``Chocolatey`` and ``MiKTeX``.
```
da1910:
```suggestion
  To build the documentation in a Windows runner, we install ``Chocolatey`` and ``MiKTeX``.
```

doc/source/changelog/1233.fixed.md
```diff
@@ -0,0 +1 @@
+MiKTeX installation with recent Chocolatey
```
da1910: This changelog entry isn't the most informative, could we make it a bit more illuminating?

SMoraisAnsys: Thanks for pointing that @da1910

da1910 (approved):

Revathyvenugopal162 (approved):
These changes make sense to me. However, I have one question: should we install MiKTeX, or is it acceptable to skip that installation step, by giving a skip-miktex-install input? If we do skip the MiKTeX installation, will everything still work as expected?

moe-ad (approved):

SMoraisAnsys:
> These changes make sense to me. However, I have one question: should we install MiKTeX, or is it acceptable to skip that installation step, by giving a skip-miktex-install input? If we do skip the MiKTeX installation, will everything still work as expected?

On windows you will have an issue unless you work with a self hosted runner where you already have installed miktex or another program that you use behind the scene to build your documentation's PDF. AFAIK Miktex is the default tool used by most of our workflows until now.

PR #1247, opened 2026-04-03
refactor: use actions/attest instead of actions/attest-build-provenance

SMoraisAnsys:
Since version `4.0.0` the action `actions/attest-build-provenance` is simply a wrapper around [actions/attest](https://github.com/actions/attest). Moving to the "main" action should prevent us from missing new updates if the synchronization is not performed on `actions/attest-build-provenance` side.

See https://github.com/actions/attest-build-provenance#usage or https://github.com/actions/attest-build-provenance/pull/835/changes#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6

### DIFF
--- a/build-library/action.yml
+++ b/build-library/action.yml
@@ -216,6 +216,6 @@ runs:
 
     - name: Generate build provenance attestation
       if: inputs.attest-provenance == 'true'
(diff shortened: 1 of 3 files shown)


AlejandroFernandezLuces (approved):
LGTM!

PR #1251, opened 2026-04-07
chore: revert check-environment-approval action deprecation

moe-ad:
This reverts #1232. The dependabot behavior we observed is now gone.


SMoraisAnsys (changes requested):
Echoing on the discussion over the PR where the action was removed. Since the action wasn't used by any ansys repo and will probably not be, we should consider making it private. Could you update that ?

moe-ad:
> Echoing on the discussion over the PR where the action was removed. Since the action wasn't used by any ansys repo and will probably not be, we should consider making it private. Could you update that ?

@SMoraisAnsys handled in https://github.com/ansys/actions/pull/1251/commits/82229971b176c7c38662b1535e7d9b09c1830049.

SMoraisAnsys (changes requested):
In private action we enforce that all inputs should be passed and no default value should be used. Could you update that too ?

moe-ad:
> In private action we enforce that all inputs should be passed and no default value should be used. Could you update that too ?

Thanks for catching that @SMoraisAnsys. Done in https://github.com/ansys/actions/pull/1251/commits/b44c8e5d89f8c269331e4374b42a556652e61ce8.

_check-environment-approval/action.yml
```diff
@@ -0,0 +1,105 @@
...
+  .. important:: **Required GitHub Permissions**
+
+     This action does not require any GitHub permissions as it only evaluates GitHub context
+     and outputs an environment name. The actual environment approval is configured at the
+     workflow level.
+
+inputs:
+
+  # Optional inputs
+
```
SMoraisAnsys:
```suggestion

```

SMoraisAnsys (changes requested):
Can you add back the removed fragment file ?

SMoraisAnsys (approved):

PR #1252, opened 2026-04-08
chore: bump zizmor version

moe-ad:
Bumping the version of zizmor in preparation for `v10.3` release.

We already had a discussion about not enforcing `secrets-outside-env`.

As for `superfluous-actions`, I think disabling it also makes sense, otherwise it forces using cli tools in place of some popular actions. For example, using `gh release create` in place of `softprops/action-gh-release` and using `gh pr/issue comment` in place of `peter-evans/create-or-update-comment`. Enforcing such a shift will cause significant friction (not everyone is familiar with using cli tools for example, among other things).

### DIFF
--- a/check-actions-security/action.yml
+++ b/check-actions-security/action.yml
@@ -122,6 +122,12 @@ runs:
       run: |
         cat << _EOF_ > zizmor.yml
         rules:
+          secrets-outside-env:
+            disable: true
+
+          superfluous-actions:
+            disable: true
+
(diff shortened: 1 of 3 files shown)


ecoussoux-ansys (approved):
Approving the changes. 
I agree with the rationale for disabling the ``superfluous-actions`` rule.

moe-ad:
Thanks @ecoussoux-ansys!

PR #1260, opened 2026-04-14
feat: sbom and wheelhouse filtering on GitHub release

SMoraisAnsys:
This PR adds support for filtering distribution files in the `dist/wheelhouse` and `dist/sbom` directories when releasing on Github. 
The main reason for this change is that the combinatorial associated to the build-wheelhouse action can lead to a releases page heavily loaded, making it hard for users to find the artifact they would like to consume. Below is an example in pyaedt 
image
where not all items are showing AND where we have been discarding the wheelhouses without optional target. This is because most (if not all) of our users request the wheelhouse with `all` target. 

While removing the wheelhouses without optional target reduces the number of artifact, it can also lead to a release that is not working without an optional dependency installed :/ This scenario happened in a release of `pyaedt` where the command `import ansys.aedt.core` was failing due to a graphic dependency being required. This error went through the CI because we removed the target '' from the build-wheelhouse job as we didn't want to many artifacts in the releases page...

The proposed feature would help to keep having the full build-wheelhouse job and filtering what gets into the releases page.

---

[Here](https://github.com/SMoraisAnsys/ansys-tools-visualization-interface/releases) is a successful run where I played with the filter `"*ubuntu-latest*,*windows-latest-3.12*,*-all-wheelhouse-macos-latest-3.14*"` and which worked as expected.

### DIFF
--- a/python-utils/release_github_utils.py
+++ b/python-utils/release_github_utils.py
@@ -20,6 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import fnmatch
 import re
 from pathlib import Path
 
@@ -157,3 +158,29 @@ def get_release_notes(pyproject_path: Path):
 
     # Save the env variable
     save_env_variable("RELEASE_NOTES_BODY", body)
+
+
+def filter_dist_files(dist_filter: str) -> None:
+    """Filter files in wheelhouse and SBOM distribution directories.
+
+    Files that do not match any of the provided glob patterns are deleted.
+
+    Parameters
+    ----------
+    dist_filter: str
+        Comma-separated list of glob patterns to keep (e.g. ``'*all*,*graphics*'``).
+    """
+    directories = [Path("dist/wheelhouse"), Path("dist/sbom")]
+
+    patterns = [p.strip() for p in dist_filter.split(",") if p.strip()]
+
+    for directory in directories:
+        if not directory.is_dir():
+            continue
+        print(f"Filtering in {directory} with patterns: {dist_filter}")
+        for file in directory.rglob("*"):
+            if not file.is_file():
+                continue
+            if not any(fnmatch.fnmatch(file.name, pattern) for pattern in patterns):
+                print(f"  Removing: {file}")
+                file.unlink()
--- a/release-github/action.yml
+++ b/release-github/action.yml
@@ -329,25 +329,6 @@ runs:
 
     # ------------------------------------------------------------------------
 
-    - name: "Filter SBOM or wheelhouse files (if needed)"
-      if: inputs.dist-filter != ''
-      shell: python
-      env:
-        DIST_FILTER: ${{ inputs.dist-filter }}
-      run: |
-        import os
-        import sys
-
-        github_action_path = os.getenv("GITHUB_ACTION_PATH")
-        sys.path.insert(1, f'{github_action_path}/../python-utils/')
-
-        from release_github_utils import filter_dist_files
-
-        dist_filter = os.getenv("DIST_FILTER")
-        filter_dist_files(dist_filter)
-
-    # ------------------------------------------------------------------------
-
     - uses: ansys/actions/_logging@main
       if: inputs.only-code == 'false'
       with:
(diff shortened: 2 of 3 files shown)


moe-ad (approved):
LGTM. Thanks for this @SMoraisAnsys!

release-github/action.yml
```diff
@@ -442,6 +452,23 @@ runs:
         ${INSTALL_COMMAND} --upgrade pip tomlkit=="$TOMLKIT_VERSION"
         ${INSTALL_COMMAND} pypandoc-binary=="$PYPANDOC_BINARY_VERSION"
```
dipinknair, line 453: add logging ?

```
      if: inputs.dist-filter != ''
      with:
        level: "INFO"
        message: >
          Filter dist/wheelhouse and dist/sbom directories using patterns: ${{ inputs.dist-filter }}
```
?

python-utils/release_github_utils.py
```diff
@@ -157,3 +158,29 @@ def get_release_notes(pyproject_path: Path):
...
+    """Filter files in wheelhouse and SBOM distribution directories.
+
+    Files that do not match any of the provided glob patterns are deleted.
+
+    Parameters
+    ----------
+    dist_filter: str
+        Comma-separated list of glob patterns to keep (e.g. ``'*all*,*graphics*'``).
+    """
+    directories = [Path("dist/wheelhouse"), Path("dist/sbom")]
```
dipinknair, line 173:
```suggestion
    if not dist_filter:
       return
    directories = [Path("dist/wheelhouse"), Path("dist/sbom")]
```

python-utils/release_github_utils.py
```diff
@@ -157,3 +158,29 @@ def get_release_notes(pyproject_path: Path):
...
+    for directory in directories:
+        if not directory.is_dir():
+            continue
+        print(f"Filtering in {directory} with patterns: {dist_filter}")
+        for file in directory.rglob("*"):
+            if not file.is_file():
+                continue
+            if not any(fnmatch.fnmatch(file.name, pattern) for pattern in patterns):
+                print(f"  Removing: {file}")
+                file.unlink()
```
dipinknair, line 186: add something like this at the end ?
```python

remaining = [f for f in directory.rglob("*") if f.is_file()]
if not remaining:
   print(f"WARNING: All files in {directory} were removed by the filter. No artifacts will be uploaded from this directory.")
```

release-github/action.yml
```diff
@@ -442,6 +452,23 @@ runs:
...
+        import sys
+
+        github_action_path = os.getenv("GITHUB_ACTION_PATH")
+        sys.path.insert(1, f'{github_action_path}/../python-utils/')
+
+        from release_github_utils import filter_dist_files
+
+        dist_filter = os.getenv("DIST_FILTER")
+        filter_dist_files(dist_filter)
+
```
dipinknair, line 471:
```
    - name: "Display the structure of the 'dist/' folder after filtering"
      if: inputs.dist-filter != ''
      shell: bash
      run: ls -R dist/
```
may be adding this might be a good idea for debugging in future ?

clatapie (approved):

python-utils/release_github_utils.py
SMoraisAnsys: This shouldn't be required as this function expects a string and is only intended to be called if the value is different from `''`.

python-utils/release_github_utils.py
SMoraisAnsys: I don't think it is a good idea to add this kind of check, I'd prefer to add the step you mentioned in the CI.

python-utils/release_github_utils.py
```diff
@@ -157,3 +158,29 @@ def get_release_notes(pyproject_path: Path):
...
+
+
+def filter_dist_files(dist_filter: str) -> None:
+    """Filter files in wheelhouse and SBOM distribution directories.
+
+    Files that do not match any of the provided glob patterns are deleted.
+
+    Parameters
+    ----------
+    dist_filter: str
```
MaxJPRey, line 170: Why not providing a list of string instead of a string made of string?
Would it be possible.

SMoraisAnsys: Sure, but we would have to add another layer to make that transformation from the workflow into that function. This would add complexity and I don't see much gain from it so I'd suggest to leave it as it is :)

PR #1261, opened 2026-04-15
ci: add type checking

SMoraisAnsys:
As title says. Update the CI with `ty` checking and associated fix.

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -134,16 +134,13 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Set up Python
-        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      - name: "Set up Python"
+        uses: ansys/actions/_setup-python@main
         with:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
-
-      - name: Set up uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
-        with:
-          enable-cache: false
-          prune-cache: true
+          use-cache: false
+          provision-uv: true
+          prune-uv-cache: true
 
       - name: Create virtual environment
         run: uv venv .venv
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -131,6 +131,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: Set up Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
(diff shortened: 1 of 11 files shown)


python-utils/versions.py
```diff
@@ -143,7 +146,11 @@ def write_versions_file() -> None:
     # Other versions (including stable)
     full_list = sorted(get_versions_list(), reverse=True)
     for version in full_list[:render_last]:
-        if stable_release and version == Version(stable_release):
+        if (
+            stable_release is not None
+            and version == Version(stable_release)
+            and url_stable is not None
```
moe-ad:
```suggestion
```

If stable release is not None, url stable is already not None.

python-utils/parse_pr.py
```diff
@@ -560,7 +560,9 @@ def sort_towncrier_types(config: dict, changelog_sections: list):
     towncrier_section["type"] = new_types
 
 
-def get_towncrier_config_value(category: str, pyproject_path: str = "pyproject.toml"):
+def get_towncrier_config_value(
+    category: str, pyproject_path: str | Path = "pyproject.toml"
```
moe-ad, line 564: I guess the logic of the function should be modified a bit. A simple `if isinstance(pyproject_path, str)` above line 583 should suffice. The docstring needs to updated too to reflect that `pyproject_path` can now be a `Path` object.

python-utils/parse_doc_conf.py
```diff
@@ -43,8 +43,9 @@ def get_project_name(conf_path):
     with open(conf_path, "r") as conf_file:
         for line in conf_file:
             if line.strip().startswith(("project =", "project=")):
-                print(line)
```
moe-ad, line 46: I guess printing the line was intentional (for debugging purposes?) and we should probably keep.

.github/workflows/ci_cd_pr.yml
```diff
@@ -125,6 +125,40 @@ jobs:
...
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: ${{ env.MAIN_PYTHON_VERSION }}
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        with:
+          enable-cache: false
+          prune-cache: true
```
moe-ad:
```suggestion
      - name: "Set up Python"
        uses: ansys/actions/_setup-python@main
        with:
          python-version: ${{ env.MAIN_PYTHON_VERSION }}
          use-cache: false
          provision-uv: true
          prune-uv-cache: true
```

Maybe we should use this instead?

moe-ad (commented):
Added a few comments.

python-utils/parse_doc_conf.py
SMoraisAnsys: I think it was resulting from failed cleanup. I don't think that there is a need to print the line at all (might have been usefull when coding that though)

python-utils/parse_pr.py
SMoraisAnsys: Handling an input of type `str | Path` and using `path = Path(input)` is a pattern used by a lot of libraries. We could use `isinstance` but I would prefer to avoid doing it. Thanks for catching the miss in the docstring

moe-ad (approved):
LGTM.

RobPasMue (approved):
Overall LGTM!

MaxJPRey (approved):

PR #1262, opened 2026-04-16
refactor: check-vulnerabilities to use sub process and avoid sys.argv manipulation

ansjmoody:
Run Safety as a subprocess because `Safety` scans the parent process, which can fail when the script invokes `safety` directly due to not being able to recognize the argument `--run-local`.
Improved exception handling when invoking `Safety`.
Run Bandit as a subprocess for consistency.


ansjmoody:
This is related to issue #641 but does not fix it.

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -331,36 +331,40 @@ def generate_advisory_files():
...
     # Delete previous advisory files
     if os.path.exists("info_safety.json"):
         os.remove("info_safety.json")
     if os.path.exists("info_bandit.json"):
         os.remove("info_bandit.json")
 
-    # Safety check
+    scripts_dir = os.path.dirname(sys.executable)
+    safety_exe = os.path.join(scripts_dir, "safety.exe")
+    bandit_exe = os.path.join(scripts_dir, "bandit.exe")
```
SMoraisAnsys:
```suggestion
    import shutil
    import subprocess

    # Delete previous advisory files
    if os.path.exists("info_safety.json"):
        os.remove("info_safety.json")
    if os.path.exists("info_bandit.json"):
        os.remove("info_bandit.json")

    safety_exe = shutil.which("safety")
    if safety_exe is None:
        raise FileNotFoundError("safety executable not found")

    bandit_exe = shutil.which("bandit")
    if bandit_exe is None:
        raise FileNotFoundError("bandit executable not found")
```
The use of EXE might be tightly coupled to windows and not other OS. I would avoid to leverage `shutil` here.

Also, adding the check on bandit / safety is required as we no longer import them. I've a feeling that the current implementation could lead to a silent error if safety or bandit is not installed.

SMoraisAnsys (changes requested):
Hi @ansjmoody, thanks for the changes proposition. I think that not playing with `sys.argv` is a good idea, could you try your changes on a example to ensure that everything is correctly working ?

check-vulnerabilities/check_vulnerabilities.py
ansjmoody: @SMoraisAnsys That makes sense. These changes worked on Windows 11. It would help if someone else could test on Linux.

ansjmoody: I applied the suggestion in a new commit. I had issues applying the suggestion inline in this PR.

SMoraisAnsys: @ansjmoody you could have just accepted the proposition instead of performing the changes manually. The idea of proposing the changes is to demonstrate and for the changes to be accepted unless there is some concern with them.

ansjmoody: @SMoraisAnsys Yeah, I tried to accept the proposition, but it did not work for some reason. That's what my previous comment was trying to say.

SMoraisAnsys:
@ansjmoody Could you run the changes to illustrate that they work both for windows and linux ?

ansjmoody:
> @ansjmoody Could you run the changes to illustrate that they work both for windows and linux ?

@SMoraisAnsys I ran this on Windows and Python 3.13.11 with no issues. I don't have a Linux environment that I can use for testing. Can someone else test this on Linux?

SMoraisAnsys:
I've performed some tests using this branch through [PyAEDT's CI](https://github.com/ansys/pyaedt/pull/7576) and the changes seem to work:
- direct use of the actions in [windows](https://github.com/ansys/pyaedt/actions/runs/24883357353/job/72857133046) and [linux](https://github.com/ansys/pyaedt/actions/runs/24883357353/job/72857133030)
- manual use of the modified files in [windows](https://github.com/ansys/pyaedt/actions/runs/24883357353/job/72857133029) and [linux](https://github.com/ansys/pyaedt/actions/runs/24883357353/job/72857133024)

SMoraisAnsys (approved):
LGTM
This modifications should be followed by additional updates (cf #1281)  but I'd suggest to do that in another PR

Revathyvenugopal162 (approved):

RobPasMue (approved):
LGTM!

PR #1265, opened 2026-04-17
refactor: doc deploy actions

### LINKED ISSUE #1103: Further improvements to how pull request documentation gets cleaned up.
I think this is best explained with an example.

Suppose **PR-1** and **PR-2** both have deployed documentation in some repo. Consider four points in time: **u** < **x** < **y** < **z**.

- At time u, PR-1 triggers the `doc-deploy-pr` workflow. At this moment, the deployment contains documentation for both PR-1 and PR-2 (i.e. `pull/1` and `pull/2` directories exist on the gh-pages branch).
- At time x, PR-2 is merged, which triggers the `doc-deploy-dev` workflow.
- That workflow finishes at time y, removing the deployed documentation for PR-2 (the pull/2 directory is removed).
- Later, at time z, the workflow that started at time u finishes.

The problem is that the workflow started at u still “thinks” PR-2 exists, because it captured the state before PR-2 was merged (i.e. had information about `pull/2`). When it finishes at z, it redeploys documentation for PR-2, effectively undoing the removal that happened at time y.

Is there a way we can avoid this?

### LINKED ISSUE #1104: Avoid single-commit strategy in doc-deploy actions
**Problem**

As per the issue raised in #1103, out `doc-deploy` actions must:

- Ensure that the deploy branch is an orphan branch
- Avoid having a unique commit in this branch

Previous issues introduce race conditions, large commit sizes, and long build times.

Furthermore, having multiple commits would allow us to roll-back a documentation in case we lost some data.

moe-ad:
Close #1104. Close #1103.

I went down the rabbit hole with this one, but the discovery was worth it in the end.

The gist of issues #1104 and #1103 is that we only need to update `doc-deploy-pr` and `doc-deploy-dev`, `doc-deploy-stable` stays the same. One way we could have handled things is as follows:
  - For `doc-deploy-pr`, we set `publish_dir` and `destination_dir` to the target PR directory. This will solve the race condition described in #1103, which is really my major concern. Normally, the `keep-files` option having a value of `true` "ideally" means that all other files on the `gh-pages` branch will still be kept, so this change should "ideally" be enough.
  - Regarding committing diffs for both `doc-deploy-pr` and `doc-deploy-dev`, we can either choose to leave things the way they currently are, or start committing diffs by setting `force-orphan` to false by default. Projects can already decide if they want the latter (since the option is already exposed, see [pydpf-composites](https://github.com/ansys/pydpf-composites/blob/a1fc81d680afab59ba3ac32a830129c6a57b0f99/.github/workflows/ci_cd.yml#L257) for example), so I was initially leaning towards leaving things the way they are for both actions.
 
The problem I discovered was the way `peaceiris/action-gh-pages` treats other options when `force-orphan` is set to `true`, most especially the `keep-files` option. In summary, `keep-files` option is not respected when `force-orphan` is `true` (see [this issue](https://github.com/peaceiris/actions-gh-pages/issues/455)) and this means that the described solution for stoping the race condition with `doc-deploy-pr` would wipe every other files besides what is in `pull/{XXXX}` when `force-orphan` is true.

So the solution I am proposing in this PR is to:
  - Stop exposing `force-orphan` option for `doc-deploy-pr` and set it to false internally. I have checked both `ansys` and `ansys-internal` orgs and I didn't find a single instance where the option was explicitly used, so we can just deprecate it silently like I have done.
  - Not change the default behavior of `doc-deploy-dev`. It is nice to still have it keep cleaning the `gh-pages` branch history when it runs.


moe-ad:
Tests:
https://github.com/ansys-internal/test-repo-moe/pull/36
https://github.com/ansys-internal/test-repo-moe/pull/37

doc-deploy-pr/action.yml
```diff
@@ -130,7 +122,7 @@ runs:
   using: "composite"
   steps:
 
-    - uses: ansys/actions/_pr-doc-deployment@main
+    - uses: ansys/actions/_pr-doc-deployment@refactor/doc-deploy-actions
```
moe-ad:
```suggestion
    - uses: ansys/actions/_pr-doc-deployment@main
```

To be reverted before merging the PR.

doc-deploy-pr/action.yml
```diff
@@ -142,10 +134,9 @@ runs:
         repository: ${{ inputs.repository }}
         branch: ${{ inputs.branch }}
         commit-message: ${{ inputs.commit-message }}
-        force-orphan: ${{ inputs.force-orphan }}
         maximum-pr-doc-deployments: ${{ inputs.maximum-pr-doc-deployments }}
 
-    - uses: ansys/actions/_pr-doc-clean@main
+    - uses: ansys/actions/_pr-doc-clean@refactor/doc-deploy-actions
```
moe-ad:
```suggestion
    - uses: ansys/actions/_pr-doc-clean@main
```

To be reverted before merging the PR.

doc-deploy-pr/action.yml
```diff
@@ -142,10 +134,9 @@ runs:
...
       if: github.event.pull_request.state == 'closed'
```
SMoraisAnsys, line 140: Could we leave the warning message here ?

```
          Starting with version v10.2, the doc-deploy-pr action no longer requires you to add the "closed" GitHub event to "on.pull_request.types" for
          documentation cleanup when PRs are closed. Documentation for closed PRs is now cleaned up
          automatically in the background without requiring an explicit event trigger or additional configuration. For more details, see
          https://actions.docs.ansys.com/version/stable/migrations/docs-deploy-pr-setup.html
```

_pr-doc-clean/action.yml
```diff
@@ -60,37 +60,24 @@ inputs:
...
-  force-orphan:
-    description: |
-      Whether to force the deployment branch to be orphan or not. Default value
-      is ``true``.
-    required: false
-    default: true
+    required: true
     type: string
 
 runs:
```
SMoraisAnsys: Shouldn't we just remove this action ? It's internal, doesn't contain any code any more and would avoid "handling" this kind of changes in a file with limited use.

SMoraisAnsys (approved):
This seems like a solution to the problem exposed. I'm not sure if it's needed given that no one was leveraging the orphan input but maybe we should also add documentation on the change of behavior associated to the cleanup logic. Since there was a recent change associated to `github.event.pull_request.state` in the past weeks, users might get confused with the cleanup management. Could be highlighted in the doc-deploy-dev action which still exposes `force-orphan` ?

I left a minor comment that isn't related to the logic of this PR but was highlighted through it's review :D

_pr-doc-clean/action.yml
```diff
@@ -60,37 +60,24 @@ inputs:
...
   repository:
     description: |
       Repository name in the form of ``username/repository`` to be used for
       deploying the documentation. The current repository is assumed by default.
-    required: false
-    default: 'current'
+    required: true
     type: string
 
   branch:
```
clatapie:
```suggestion
  deploy-branch:
```

_pr-doc-deployment/action.yml
```diff
@@ -62,62 +62,46 @@ inputs:
...
   repository:
     description: |
       Repository name in the form of ``username/repository`` to be used for
       deploying the documentation. The current repository is assumed by default.
-    required: false
-    default: 'current'
+    required: true
     type: string
 
   branch:
```
clatapie, line 88:
```suggestion
  deploy-branch:
```

_pr-doc-deployment/action.yml
clatapie, line 131: Is there a reason not to keep the default value of the ``commit-message`` argument?

_pr-doc-clean/action.yml
moe-ad: Great idea. I have removed the file in https://github.com/ansys/actions/pull/1265/commits/0df91d70956045ca02caae127cbbd13eed687818.

doc-deploy-pr/action.yml
moe-ad: I added the message as info to `_pr-doc-deployment` since that file is more tightly related to the changes made. Let me know what you think about the [wording](https://github.com/ansys/actions/pull/1265/changes/0df91d70956045ca02caae127cbbd13eed687818#diff-cebda1bf37df927e20df32bd0d26ed6b126af4f4e004b906f1f7dfb77c71954cR111) of the message.

clatapie (approved):
Apart from those questions, it looks good to me!

Revathyvenugopal162 (approved):
LGTM

_pr-doc-deployment/action.yml
```diff
@@ -62,68 +62,63 @@ inputs:
...
+    - uses: ansys/actions/_logging@main
+      with:
+        level: "INFO"
+        message: >
+          Starting with v10.3, doc-deploy-pr action doesn't force the deployment branch to be an orphan with
+          only the latest commit (as opposed doc-deploy-dev and doc-deploy-stable which force the deployment
+          branch to be an orphan with only the latest commit by default). This helps to prevent race conditions in very active
+          repositories (see https://github.com/ansys/actions/issues/1103 for more information). This means PR documentation are
+          added to the deployment branch as separate commits, and only a run of doc-deploy-dev or doc-deploy-stable will consolidate
+          (into a single commit) the linear history that results from a sequence PR documentation deployments.
```
SMoraisAnsys:
```suggestion
          repositories (see https://github.com/ansys/actions/issues/1103 for more information). This means PR documentation are
          added to the deployment branch as separate commits. The branch history is only rewritten into a single commit when
          doc-deploy-dev or doc-deploy-stable is run with ``force-orphan: true`` (which is the default value).
```
I'm not sure to catch the meaning of consolidate or the need to mention a linear history (I might be missing some information here though). Would that be a good alternative ?

_pr-doc-deployment/action.yml
SMoraisAnsys: We enforce that inner actions (starting with `_...`) are only composed of required inputs => no default value should be provided. This should be added to the public action.

doc-deploy-pr/action.yml
SMoraisAnsys: resolving following my comment

_pr-doc-clean/action.yml
moe-ad: This file has been deleted in https://github.com/ansys/actions/commit/0df91d70956045ca02caae127cbbd13eed687818.

_pr-doc-deployment/action.yml
moe-ad: This input name is consistent across all the `doc-deploy-*` actions. We can revisit if we want to change the input name in the future (for `v11`?) across all of them, but I would rather keep it as is for now since changing input names requires a planned deprecation period for the former input name. Thanks for the suggestion.

_pr-doc-deployment/action.yml
moe-ad: Yes, it is a good alternative.

PR #1268, opened 2026-04-20
feat(build-library): use uv build when use-uv is true

viseshrp:
When `use-uv: true` (the default), the build step now uses `uv build` instead of `python -m build`, avoiding the need to install the `build` package as a separate dependency and also speeding up builds at the same time. The `python -m build` path is fully preserved when `use-uv: false`.

Changes:

Skip installing the build package when `use-uv: true` (only `twine` is needed for validation)
Run `uv build --out-dir dist` instead of `python -m build --outdir dist` when `use-uv: true`


klmcadams (approved):
This makes sense to me, thanks!

SMoraisAnsys (approved):
LGTM, if work associated to https://github.com/astral-sh/uv/issues/8641 get handled, we might even get rid of twine on this workflow

MaxJPRey (approved):
LGTM. Thanks @viseshrp for the contribution.

SMoraisAnsys (approved):

PR #1280, opened 2026-04-22
refactor(build-library)!: remove checkout opt-in option

SMoraisAnsys:
Follow up of the discussion in https://github.com/ansys/actions/pull/1269

The rationale behind these changes is that:
- these options haven't been used in our repos despite being public for multiple months. In fact, even the maintainer that was suggesting such changes doesn't seem to be using them, see this [ref](https://github.com/ansys/pydynamicreporting/blob/main/.github/workflows/ci_cd.yml#L106-L120)
- since we allow one to disable `actions/checkout` to perform a custom checkout, the options under consideration have no use. In fact, that's how things are being handled in the link above...

@moe-ad Depending on how we want to release those changes, we might have to wait for the next major release.

### DIFF
--- a/build-library/action.yml
+++ b/build-library/action.yml
@@ -121,20 +121,6 @@ inputs:
     required: false
     type: boolean
 
-  checkout-fetch-depth:
-    description: |
-      Set fetch-depth option for actions/checkout. Default value is ``1``.
-    required: false
-    default: '1'
-    type: string
-
-  checkout-fetch-tags:
-    description: |
-      Set fetch-tags option for actions/checkout. Default value is ``false``.
-    required: false
-    default: false
-    type: boolean
-
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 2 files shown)


moe-ad:
If we are sure the input is not being used, we can just deprecate it silently and not have to wait for a major release.

Waiting for a major release necessitates deprecating the inputs the usual way, i.e. not removing the inputs yet from the action itself and logging a warning about the expected deprecation if the input usage is detected. Then we remove completely just before a major release.

If we agree on the first approach, we can just merge. I favor the first approach, I would like to avoid deprecation housekeeping if no-one has started using the input... 😄

Revathyvenugopal162 (approved):

SMoraisAnsys:
I check on our orgs, here is the result for ansys https://github.com/search?q=org%3Aansys%20checkout-fetch-depth&type=code

That's also how I found out that the options (previously requested) weren't used ...

SMoraisAnsys:
@MaxJPRey @RobPasMue Do you also agree with not waiting for a major release ?

RobPasMue (approved):

RobPasMue:
> @MaxJPRey @RobPasMue Do you also agree with not waiting for a major release ?

Agreed - let's release this asap

moe-ad (approved):

PR #1287, opened 2026-04-28
fix: ouch decompression

RobPasMue:
As title says... reported in PyCFX. Resolution attempt in https://github.com/ansys/pycfx/pull/65

### DIFF
--- a/_pr-doc-deployment/action.yml
+++ b/_pr-doc-deployment/action.yml
@@ -270,7 +270,7 @@ runs:
       env:
         PR_NUMBER: ${{ github.event.pull_request.number }}
       run: |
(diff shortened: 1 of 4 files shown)


SMoraisAnsys (approved):
LGTM

@RobPasMue Do we have a use case where hidden files are included in artifacts upload ?

RobPasMue:
FYI - worked in https://github.com/ansys/pycfx/actions/runs/25045168421/job/73361358865

RobPasMue:
> @RobPasMue Do we have a use case where hidden files are included in artifacts upload ?

Good question.. but that's going to be on how teams decide to compress their doc artifacts (i.e. including hidden files or not). My modifications would simply extract whatever content is available inside the zip.

PR #1305, opened 2026-05-08
fix: if miktex is available then make use of it

RobPasMue:
Reported in https://github.com/ansys/pysherlock/pull/778

If MikTex is installed already at system level, the new action implementation causes it to fail since we simply take it from the local installation. Let's make it more "dynamic"

1) Verify if MikTex is already there - if so, use it
2) If the available installs of MikTex are corrupt - reinstall at local user level
2) If MikTex is not installed - then use the local user install

### DIFF
--- a/_doc-build-windows/action.yml
+++ b/_doc-build-windows/action.yml
@@ -218,6 +218,15 @@ runs:
         } elseif (Test-Path $systemMiKTeX) {
           Write-Host "MiKTeX already installed at system level, using existing installation."
           echo "$systemMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          # Check if MikTex works by requesting its version
+          $miktexVersion = & "$systemMiKTeX\miktex.exe" --version
+          if ($LASTEXITCODE -ne 0) {
+            Write-Host "Existing MiKTeX installation at system level is not working properly, installing it at user level..."
+            choco install miktex --params '"/Set:basic /ThisUser"' -y
+            echo "$userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          } else {
+            Write-Host "Existing MiKTeX installation at system level is working properly."
+          }
         } else {
           # NOTE: Install at user level to avoid errors when running on self-hosted runners
           # where choco was already installed at system level.
--- a/_doc-build-windows/action.yml
+++ b/_doc-build-windows/action.yml
@@ -202,34 +202,35 @@ runs:
     - name: Install MiKTeX and update PATH with MiKTeX binaries
       shell: powershell
       run: | # zizmor: ignore[github-env]
+        # NOTE: Declared at script scope so it can be accessed inside function
+        $script:userMiKTeX = "$env:LocalAppData\Programs\MiKTeX\miktex\bin\x64"
+
         function Invoke-MiKTeXCheck {
           param(
             [string]$InstallPath,
             [string]$Level,
-            [string]$UserInstallPath
           )
           Write-Host "MiKTeX already installed at $Level level, using existing installation."
           $miktexVersion = & "$InstallPath\miktex.exe" --version
           if ($LASTEXITCODE -ne 0) {
             Write-Host "Existing MiKTeX installation at $Level level is not working properly, reinstalling it at user level..."
             choco install miktex --params '"/Set:basic /ThisUser"' -y --force
-            echo "$UserInstallPath" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+            echo "$script:userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
           } else {
             Write-Host "Existing MiKTeX installation at $Level level is working properly (version: $miktexVersion)."
             echo "$InstallPath" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
           }
         }
 
-        $userMiKTeX = "$env:LocalAppData\Programs\MiKTeX\miktex\bin\x64"
         $systemMiKTeX = "C:\Program Files\MiKTeX\miktex\bin\x64"
-        if (Test-Path "$userMiKTeX\miktex.exe") {
-          Invoke-MiKTeXCheck -InstallPath $userMiKTeX -Level "user" -UserInstallPath $userMiKTeX
+        if (Test-Path "$script:userMiKTeX\miktex.exe") {
+          Invoke-MiKTeXCheck -InstallPath $script:userMiKTeX -Level "user"
         } elseif (Test-Path "$systemMiKTeX\miktex.exe") {
-          Invoke-MiKTeXCheck -InstallPath $systemMiKTeX -Level "system" -UserInstallPath $userMiKTeX
+          Invoke-MiKTeXCheck -InstallPath $systemMiKTeX -Level "system"
         } else {
           Write-Host "MiKTeX is not installed, installing it at user level..."
           choco install miktex --params '"/Set:basic /ThisUser"' -y
-          echo "$userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          echo "$script:userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
         }
 
     - name: "Install Poppler for PDF to PNG conversion"
--- a/_doc-build-windows/action.yml
+++ b/_doc-build-windows/action.yml
@@ -202,35 +202,32 @@ runs:
     - name: Install MiKTeX and update PATH with MiKTeX binaries
       shell: powershell
       run: | # zizmor: ignore[github-env]
+        function Invoke-MiKTeXCheck {
+          param(
+            [string]$InstallPath,
+            [string]$Level,
+            [string]$UserInstallPath
+          )
+          Write-Host "MiKTeX already installed at $Level level, using existing installation."
+          $miktexVersion = & "$InstallPath\miktex.exe" --version
+          if ($LASTEXITCODE -ne 0) {
+            Write-Host "Existing MiKTeX installation at $Level level is not working properly, reinstalling it at user level..."
+            choco install miktex --params '"/Set:basic /ThisUser"' -y --force
+            echo "$UserInstallPath" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          } else {
+            Write-Host "Existing MiKTeX installation at $Level level is working properly (version: $miktexVersion)."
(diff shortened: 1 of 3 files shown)


doc-build/action.yml
```diff
@@ -265,7 +265,7 @@ runs:
 
     - name: Documentation build (Windows)
       if: ${{ runner.os == 'Windows' }}
-      uses: ansys/actions/_doc-build-windows@main
+      uses: ansys/actions/_doc-build-windows@fix/use-miktex-if-available-on-windows
```
RobPasMue:
```suggestion
      uses: ansys/actions/_doc-build-windows@main
```
To be reverted after testing

RobPasMue:
Worked in https://github.com/ansys/pysherlock/actions/runs/25545137307/job/74980228985

AlejandroFernandezLuces (approved):
LGTM if it fixes the issue. Just an opinionated suggestion. Is it possible to handle the ifs logic with github workflow syntax and keep powershell minimal? I really hate powershell it is so hard to read.

I'm thinking of something like: try to use user installation. If it fails, proceed to next approach.

RobPasMue:
> I'm thinking of something like: try to use user installation. If it fails, proceed to next approach.

We will hardly get around the powershell syntax - especially for the Windows-based action 😄 . I agree with you, it's not my favourite either... but even if we split it into GitHub workflow steps, we will still need to verify the exit codes from previous powershell operations, store paths into the env, etc. -- meaning that we might be able to reduce 20% of the PowerShell syntax at the expense of increasing a lot the number of steps ( I envision at least 3 new workflow steps in that case)...

So, overall, the gain is not too much 😄 

Let's discuss this altogether in the next team meeting (getting rid of PowerShell syntax as much as possible) - but if we attempt this, it will have to be done at a larger scale. And we would have to ensure that all runners have a bash terminal (Windows runners included, since sometimes the self-hosted ones might be missing it). But I'm all for it!

_doc-build-windows/action.yml
```diff
@@ -202,11 +202,38 @@ runs:
...
+          echo "$systemMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          # Check if MikTex works by requesting its version
+          $miktexVersion = & "$systemMiKTeX\miktex.exe" --version
+          if ($LASTEXITCODE -ne 0) {
+            Write-Host "Existing MiKTeX installation at system level is not working properly, installing it at user level..."
+            choco install miktex --params '"/Set:basic /ThisUser"' -y
+            echo "$userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          } else {
+            Write-Host "Existing MiKTeX installation at system level is working properly."
+          }
```
klmcadams: Could you make a function for the content of the if $userMiKTeX and else if $systemMiKTeX conditions? It seems like the logic is repeated in there. 

I noticed in $userMiKTeX `echo "$userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8` happens after the exit code check, but $systemMiKTeX does this before the last exit code check. Is there a reason for that?

_doc-build-windows/action.yml
```diff
@@ -202,11 +202,38 @@ runs:
...
       shell: powershell
       run: | # zizmor: ignore[github-env]
-        # NOTE: Install at user level to avoid errors when running on self-hosted runners
-        # where choco was already installed at system level.
-        # For more information, see https://github.com/ansys/actions/pull/1233
-        choco install miktex --params '"/Set:basic /ThisUser"' -y
-        echo "$env:LocalAppData\Programs\MiKTeX\miktex\bin\x64" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+        $userMiKTeX = "$env:LocalAppData\Programs\MiKTeX\miktex\bin\x64"
+        $systemMiKTeX = "C:\Program Files\MiKTeX\miktex\bin\x64"
+        if (Test-Path $userMiKTeX) {
```
MaxJPRey: Maybe we could check if the exe is there in addition of the directory . That more accurately represents MiKTeX is installed.
```suggestion
        if (Test-Path "$userMiKTeX\miktex.exe") {
```

_doc-build-windows/action.yml
```diff
@@ -202,11 +202,38 @@ runs:
...
+            choco install miktex --params '"/Set:basic /ThisUser"' -y --force
+          } else {
+            Write-Host "Existing MiKTeX installation at user level is working properly."
+          }
+          echo "$userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+        } elseif (Test-Path $systemMiKTeX) {
+          Write-Host "MiKTeX already installed at system level, using existing installation."
+          echo "$systemMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          # Check if MikTex works by requesting its version
+          $miktexVersion = & "$systemMiKTeX\miktex.exe" --version
```
MaxJPRey: We could first check the version and then add it to the path.
Otherwise, if `--version` does not work , the path will still be modified with a corrupted value.

But I got why you set the path to leverage it immediately after.

MaxJPRey (approved):
Great! I was rusty with PowerShell though.

_doc-build-windows/action.yml
RobPasMue: I'll clean that up

_doc-build-windows/action.yml
```diff
@@ -202,11 +202,35 @@ runs:
...
+        $systemMiKTeX = "C:\Program Files\MiKTeX\miktex\bin\x64"
+        if (Test-Path "$userMiKTeX\miktex.exe") {
+          Invoke-MiKTeXCheck -InstallPath $userMiKTeX -Level "user" -UserInstallPath $userMiKTeX
+        } elseif (Test-Path "$systemMiKTeX\miktex.exe") {
+          Invoke-MiKTeXCheck -InstallPath $systemMiKTeX -Level "system" -UserInstallPath $userMiKTeX
+        } else {
+          Write-Host "MiKTeX is not installed, installing it at user level..."
+          choco install miktex --params '"/Set:basic /ThisUser"' -y
+          echo "$userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+        }
```
SMoraisAnsys, line 234:
```suggestion
        # NOTE: Declared at script scope so it can be accessed inside function
        $script:userMiKTeX = "$env:LocalAppData\Programs\MiKTeX\miktex\bin\x64"

        function Invoke-MiKTeXCheck {
          param(
            [string]$InstallPath,
            [string]$Level,
          )
          Write-Host "MiKTeX already installed at $Level level, using existing installation."
          $miktexVersion = & "$InstallPath\miktex.exe" --version
          if ($LASTEXITCODE -ne 0) {
            Write-Host "Existing MiKTeX installation at $Level level is not working properly, reinstalling it at user level..."
            choco install miktex --params '"/Set:basic /ThisUser"' -y --force
            echo "$script:userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
          } else {
            Write-Host "Existing MiKTeX installation at $Level level is working properly (version: $miktexVersion)."
            echo "$InstallPath" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
          }
        }

        $systemMiKTeX = "C:\Program Files\MiKTeX\miktex\bin\x64"
        if (Test-Path "$script:userMiKTeX\miktex.exe") {
          Invoke-MiKTeXCheck -InstallPath $script:userMiKTeX -Level "user"
        } elseif (Test-Path "$systemMiKTeX\miktex.exe") {
          Invoke-MiKTeXCheck -InstallPath $systemMiKTeX -Level "system"
        } else {
          Write-Host "MiKTeX is not installed, installing it at user level..."
          choco install miktex --params '"/Set:basic /ThisUser"' -y
          echo "$script:userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
        }
```

SMoraisAnsys (commented):
LGTM, thanks for handling that @RobPasMue. I left a minor comment to remove the extra argument (and had a great time discovering coding in powershell... \o/)

Given the problem that happened in the past with choco and potential future updates, I would highly recommend any installation at system level to be removed and reinstalled at user level but if this. But, that's another siubject :)

RobPasMue:
> Given the problem that happened in the past with choco and potential future updates, I would highly recommend any installation at system level to be removed and reinstalled at user level but if this. But, that's another siubject :)

Let's leave that for another day 😄 - I think that's a broader discussion we should have for all choco-related installs we perform. And we should handle the changes all at the same time throughout our actions

RobPasMue:
@MaxJPRey @SMoraisAnsys @klmcadams - since your comments implied modifications, please review the PR again before we proceed to merge

_doc-build-windows/action.yml
```diff
@@ -202,11 +202,36 @@ runs:
...
+          Write-Host "MiKTeX already installed at $Level level, using existing installation."
+          $miktexVersion = & "$InstallPath\miktex.exe" --version
+          if ($LASTEXITCODE -ne 0) {
+            Write-Host "Existing MiKTeX installation at $Level level is not working properly, reinstalling it at user level..."
+            choco install miktex --params '"/Set:basic /ThisUser"' -y --force
+            echo "$script:userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          } else {
+            Write-Host "Existing MiKTeX installation at $Level level is working properly (version: $miktexVersion)."
+            echo "$InstallPath" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
+          }
```
Revathyvenugopal162, line 222:
```suggestion
            echo "$script:userMiKTeX" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
            return
          } 
          Write-Host "Existing MiKTeX installation at $Level level is working properly (version: $miktexVersion)."
          echo "$InstallPath" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8

```
Since this is an error case, can we use an early return instead of else? I’m not expert in PowerShell, so please confirm 🙂

RobPasMue: It's not an actual error case - we are trying to verify that the package is properly installed. We are not trying to simply check that an error occurred and return, we are attempting to reinstall properly and fix the install.

Thus, it's clearer from a logic perspective if we have the two if-branches separated IMO

SMoraisAnsys (approved):
LGTM, just revert the pinned branch and all good :)

RobPasMue:
@moe-ad  - could you please trigger a patch release with these changes?

moe-ad:
Release completed.

RobPasMue:
Thanks @moe-ad !!

PR #1310, opened 2026-05-12
feat: allow skipping python setup

SMoraisAnsys:
This pull request adds a skip-python-setup input to the affected composite actions. The default value is `false`, which leads to keeping the same behavior as before. When set to `true`:

- The _setup-python step is skipped entirely, leaving $GITHUB_PATH untouched.
- A validation step checks that the Python already on PATH matches the version specified in the python-version input (supports both major.minor and major.minor.micro precision), failing fast with a clear error message if there is a mismatch.

This provides flexibility for container-based environments with a pre-installed Python when the current implementation does not. As a matter of fact, the latest release of `ansys/actions` can't be used to build wheelhouse in a rocky container due to `glibc` incompatibilities, even if one does manually install python before calling `ansys/actions/build-wheelhouse`.

The affected actions are currently limited to `build-library`, `build-wheelhouse`, `doc-build` and `tests-pytest` as they are the ones most likely benefiting from it. The `build` ones would allow to build on specific environments, e.g. rocky; the `doc` and `tests` ones are used in some of our self-hosted runners.

### DIFF
--- a/doc-build/action.yml
+++ b/doc-build/action.yml
@@ -223,9 +223,8 @@ inputs:
 
       .. warning::
 
-          The python version available in the environment must match the
-          version specified in the ``python-version`` input. If not, the
-          action will fail.
+          When this option is enabled, the ``python-version`` input is ignored.
+          The Python version already available in the environment is used as-is.
 
     required: false
     default: false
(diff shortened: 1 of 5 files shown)


SMoraisAnsys:
I tested the build-wheelhouse changes and it:
- worked [here](https://github.com/ansys/pyaedt/actions/runs/25727805595/job/75545021041) (wheelhouse for rocky)
- failed as expected [here](https://github.com/ansys/pyaedt/actions/runs/25728076688) (wrong python checking)

doc-build/action.yml
```diff
@@ -225,14 +243,36 @@ runs:
...
       with:
         python-version: ${{ inputs.python-version }}
         use-cache: ${{ inputs.use-python-cache }}
         provision-uv: ${{ inputs.use-uv }}
         prune-uv-cache: ${{ inputs.use-python-cache != 'true' }}
 
+    - name: "Check Python version matches expected version"
+      if: ${{ inputs.skip-python-setup == 'true' }}
+      shell: bash
+      env:
```
RobPasMue: Question...

If you are passing the ``skip-python-setup`` as ``true``.. as a user I wouldn't expect this to fail just because the expected Python version is not matching my actual version.. 

In fact, as a user, I wouldn't be passing the ``python-version`` input probably... which means that it would simply pick-up the default value (i.e. ``3.11``)... And I might have installed myself 3.12 or 3.13... causing the workflow to fail "weirdly".

I would honestly recommend to remove it - it is up to the workflow owner to make sure that they are using the proper/expected Python version in case they decide to skip it.

RobPasMue (changes requested):
Big pitfall IMO with this check - I'm fine with the skipping logic but we should definitely remove the check introduced to match the ``python-version`` input

doc-build/action.yml
SMoraisAnsys: Following offline discussions, updates have been performed in 60a6738 to remove the checking & log the skip of python setup to use system python.

RobPasMue (approved):
LGTM! Thanks @SMoraisAnsys

MaxJPRey (approved):
Nice. I would be curious to see how much time we can save at large in the ecosystem with this option set to false 😃

PR #1313, opened 2026-05-13
docs: update local usage of check vulnerabilities

### LINKED ISSUE #1281: Local use of check-vulnerability does not align with the action behavior
Following the changes in #920 we should also update the `--run-local` behavior to expect a few changes to be performed before hand (extra venv and dump of the package dependencies into a requirement file). This should be clearly stated in the note explaining how to run the check-vulnerabilities.py file.

Note: this changes should be performed after https://github.com/ansys/actions/pull/1262 is completed

SMoraisAnsys:
Following #1262, this PR aligns the check vulnerability usage in local mode.

Close #1281

### DIFF
--- a/check-vulnerabilities/check_vulnerabilities.py
+++ b/check-vulnerabilities/check_vulnerabilities.py
@@ -41,6 +41,7 @@ from github.AdvisoryVulnerability import (
     SimpleAdvisoryVulnerability,
     SimpleAdvisoryVulnerabilityPackage,
 )
+from requests.exceptions import SSLError
 
 TOKEN = os.environ.get("DEPENDENCY_CHECK_TOKEN", None)
 PACKAGE = os.environ.get("DEPENDENCY_CHECK_PACKAGE_NAME", None)
@@ -114,7 +115,15 @@ def check_vulnerabilities():
     g = github.Github(auth=github.Auth.Token(TOKEN))
 
     # Get the repository
-    repo = g.get_repo(REPOSITORY)
+    try:
+        repo = g.get_repo(REPOSITORY)
+    except SSLError as e:
+        raise RuntimeError(
+            "SSL error occurred while trying to access the repository. "
+            "If this error occured while running on a corporate network, "
+            "this may be due to a corporate SSL inspection proxy. "
+            "Ensure that REQUESTS_CA_BUNDLE is set correctly."
+        ) from e
 
     # Get the available security advisories
     existing_advisories = {}
@@ -365,7 +374,7 @@ def generate_advisory_files():
     # Safety check - invoke the safety executable directly to avoid Safety reading
     # the parent process argv (a Safety 3.x bug when called via `python -m safety`)
     try:
-        subprocess.run(
+        result = subprocess.run(
             [
                 safety_exe,
                 "check",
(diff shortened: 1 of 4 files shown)


RobPasMue (approved):

MaxJPRey (approved):
LGTM.

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -355,6 +355,9 @@ def generate_advisory_files():
     if bandit_exe is None:
         raise FileNotFoundError("bandit executable not found")
 
+    if not os.path.exists("requirements-for-safety.txt"):
+        raise FileNotFoundError("Expected requirements-for-safety.txt not found.")
```
MaxJPRey: Here we could add an additional info in the error to mention why this file is needed.
Even though people can still refer to the doc content that you added in this PR.

SMoraisAnsys: Let me know if 283ef0c suits you :)

doc/source/vulnerability-actions/index.rst
```diff
@@ -18,10 +18,32 @@ Check vulnerabilities action
...
+
+   #. Create a first virtual environment for the project and install the repository:
+
+      .. code-block:: bash
+
+         python -m venv .venv
+         source .venv/bin/activate  # On Windows: .venv\Scripts\activate
+         pip install -e .
+
+   #. Dump the project dependencies to a file. **This step is required** to isolate the
```
clatapie, line 30: This approach is fine as it enables to separate the potential dependency issues coming from ``safety`` or ``bandit``. However, it implies the user will need to delete the `requirements-for-safety.txt` and rerun the `pip freeze > requirements-for-safety.txt` command to see the changes they did from their `pyproject.toml` file taken into account in the safety outputs.

This should be mentioned somewhere either as a warning here or as another step right after "#. Run the script by running ``python check_vulnerabilities.py --run-local``."

SMoraisAnsys: I've added 42a639e to address your point @clatapie Let me know if it's enough :)

clatapie: Yes it is, thanks @SMoraisAnsys!

clatapie (approved):
LGTM!

AlejandroFernandezLuces (approved):
LGTM!

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -355,20 +364,50 @@ def generate_advisory_files():
...
+            raise RuntimeError(
+                "Safety could not reach the vulnerability database. "
+                "This typically happens on corporate networks where an SSL inspection proxy "
+                "intercepts HTTPS connections and presents its own certificate, which Python "
+                "does not trust by default. "
+                "To fix this, set REQUESTS_CA_BUNDLE to a combined CA bundle that includes "
+                "both the corporate root CA (exported as PEM from the OS certificate store) "
+                "and the standard certifi bundle. Setting it to the corporate CA alone is not "
+                "sufficient: it replaces certifi entirely, causing connections to domains not "
+                "intercepted by the proxy (e.g. api.github.com) to fail as well. "
```
RobPasMue: Shouldn't the error message be similar for both cases? I believe the first error is slightly less informative. Let's create a shared error message and make use of it in both errors raised.

SMoraisAnsys: I split it into two because using the corporate CA would fix the issue previously encountered when using safety (the DB is reached). However it's not enough for working with `github` afterward. I'll define a single message containing both explanation.

RobPasMue: But it is more than possible that if they hit the first one, they also hit the second one I guess.

RobPasMue: Raising it in 2 different places is fine - I was just referring to the fact that we could reuse parts (if not all) of the same message in both places (through string composition)

RobPasMue (approved):

check-vulnerabilities/check_vulnerabilities.py
RobPasMue: Thanks @SMoraisAnsys !

MaxJPRey (approved):

PR #1321, opened 2026-05-20
chore: pin actions dependencies

### LINKED ISSUE #985: Pin version of dependencies for each action
With recent events breaking our actions, we really should start pinning python dependencies of our actions.

Unless I'm wrong, we are only using `uv` or `pip` for installing action's related dependencies so we could move to `requirements.txt` or `pylock.toml` files. The first option is probably the best for now as the latter doesn't seem to be supported by `dependabot`, see https://github.com/dependabot/dependabot-core/issues/12094.

Since we are compatible with multiple versions of python, we might end up having multiple requirements file or using environment markers in the same file.

### LINKED ISSUE #1185: Input versions frozen and usage of locked environments
We are freezing a lot of versions lately on our ansys/actions.. but whenever they are input-based, we are forgetting to update them regularly so that they keep up-to-date.

Whenever possible, we need to move to requirements.txt based approaches where we freeze deps (or lock files, I don't care).
 
And if input-version is still desired, we need to set up mechanisms to search for updates as well and modify the *.yml files accordingly.

moe-ad:
Closes #985. See #991 for full context.

Closes #1185 as well because of [this commit](https://github.com/ansys/actions/pull/1321/commits/9b6477bf095f9df2c1663940490474948b488727).


moe-ad:
@SMoraisAnsys I have confirmed [this](https://github.com/ansys/actions/pull/991#issuecomment-3773058818) to be true:
  - https://github.com/ansys-internal/test-repo-moe/pull/38
  - https://github.com/ansys-internal/test-repo-moe/pull/39
  - https://github.com/ansys-internal/test-repo-moe/pull/40

The lock files and pyproject.toml are being updated, meaning we can remove the "uv lock" of [this workflow](https://github.com/ansys/actions/pull/1321/changes/3a0e76b0688d0321c582c859e875f570b7d458e3#diff-ae319f93885777278340290cf61ecf33d9a7220775f6244b952e41cd809236d2R1).

moe-ad:
@RobPasMue @SMoraisAnsys you can start reviewing this. All the discussions we had in #991 has been taken into account, except the provision of an input to opt out of pinned version installation. Since this is intended to ship with a major release, should I still proceed to provide such input?

.github/workflows/ci_cd_dependabot.yml
```diff
@@ -0,0 +1,77 @@
...
+
+          # Exit if there are no changes to commit
+          if git diff --quiet --exit-code; then
+            echo "No updates to requirements files detected, no changes to commit."
+            exit 0
+          else
+            echo "Requirements files update detected, preparing to commit."
+            git commit -am "fix: update requirements files [skip ci]"
+          fi
+          git push origin "HEAD:${REF_NAME}"
```
RobPasMue:
```suggestion
          git push
```
Shouldn't this be enough?

code-style/action.yml
```diff
@@ -283,12 +283,29 @@ runs:
         PRE_COMMIT_VERSION: ${{ inputs.pre-commit-version }}
         ACTIVATE_VENV: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
         BUILD_BACKEND: ${{ steps.github-environment-variables.outputs.BUILD_BACKEND }}
+        SKIP_INSTALL: ${{ inputs.skip-install }}
       run: |
         ${ACTIVATE_VENV}
-        if [[ "${BUILD_BACKEND}" == 'uv' ]]; then
-          uv pip install "pre-commit==${PRE_COMMIT_VERSION}"
+        if [[ "${PRE_COMMIT_VERSION}" == "4.6.0" ]]; then
```
RobPasMue: What's the reason for checking the version number? We should really move towards empty inputs now that we are pinning down the deps in the reqs files. Thus, if a certain version is passed in, then we leverage it. Otherwise we assume it's default behavior.

Also I'd probably remove the version inputs altogether. Users shouldn't care which version we are leveraging. We should move towards a ``use-prek`` / ``use-pre-commit`` input type

doc-changelog/action.yml
```diff
@@ -211,7 +211,12 @@ runs:
         TOMLKIT_VERSION: ${{ inputs.tomlkit-version }}
         INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --system' || 'python -m pip install' }}
       run: |
-        ${INSTALL_COMMAND} --upgrade pip towncrier==${TOWNCRIER_VERSION} tomlkit==${TOMLKIT_VERSION}
+        ${INSTALL_COMMAND} --upgrade pip
+        if [[ "${TOWNCRIER_VERSION}" == "24.8.0" && "${TOMLKIT_VERSION}" == "0.14.0" ]]; then
```
RobPasMue: Same comment as with pre-commit. We should no longer expose the version control IMO... we are pinning down deps because we know they work. We should consider actions as "applications" rather than libraries

doc-deploy-changelog/action.yml
```diff
@@ -250,7 +250,12 @@ runs:
         TOMLKIT_VERSION: ${{ inputs.tomlkit-version }}
       shell: bash
       run: |
-        python -m pip install --upgrade pip towncrier=="${TOWNCRIER_VERSION}" tomlkit=="${TOMLKIT_VERSION}"
+        python -m pip install --upgrade pip
+        if [[ "${TOWNCRIER_VERSION}" == "24.8.0" && "${TOMLKIT_VERSION}" == "0.14.0" ]]; then
```
RobPasMue: Same

doc-style/action.yml
```diff
@@ -184,17 +184,17 @@ runs:
...
        TOMLKIT_VERSION: ${{ inputs.tomlkit-version }}
       run: |
-        python -m pip install --upgrade pip tomlkit==${TOMLKIT_VERSION}
-
-    - name: "Install docutils"
-      shell: bash
-      run: |
-        python -m pip install --upgrade docutils
+        python -m pip install --upgrade pip
+        if [[ "${TOMLKIT_VERSION}" == "0.14.0" ]]; then
```
RobPasMue: Same

release-github/action.yml
```diff
@@ -449,8 +449,13 @@ runs:
         PYPANDOC_BINARY_VERSION: ${{ inputs.pypandoc-binary-version }}
         INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --system' || 'python -m pip install' }}
       run: |
-        ${INSTALL_COMMAND} --upgrade pip tomlkit=="$TOMLKIT_VERSION"
-        ${INSTALL_COMMAND} pypandoc-binary=="$PYPANDOC_BINARY_VERSION"
+        ${INSTALL_COMMAND} --upgrade pip
+        if [[ "${PYPANDOC_BINARY_VERSION}" == "1.15" && "${TOMLKIT_VERSION}" == "0.14.0" ]]; then
```
RobPasMue: Same

code-style/action.yml
moe-ad: Thanks for the suggestion. This and similar comments have been addressed in https://github.com/ansys/actions/pull/1321/commits/9b6477bf095f9df2c1663940490474948b488727. Let me know what you think.

RobPasMue: giphy

Looks just like I wanted =) great job @moe-ad

moe-ad: Thanks!

.github/workflows/ci_cd_dependabot.yml
```diff
@@ -0,0 +1,77 @@
...
+
+name: Update requirement.txt files
+
+on:
+  push:
+    branches:
+      - "dependabot/uv/**"
+
+env:
+  MAIN_PYTHON_VERSION: '3.13'
```
RobPasMue:
```suggestion
  MAIN_PYTHON_VERSION: '3.14'
```
Since you are already aligning in #1337 - let's not forget this

RobPasMue (approved):
LGTM

.github/workflows/ci_cd_dependabot.yml
moe-ad: Done in https://github.com/ansys/actions/pull/1321/commits/dbd19fa079358bd150d360d0711c2d195bd4acdf.

.github/dependabot.yml
```diff
@@ -3,10 +3,6 @@ updates:
   - package-ecosystem: "pip"
     directories:
       - "/requirements"
-      - "/check-vulnerabilities"
-      - "/hk-package-clean-except"
-      - "/hk-package-clean-untagged"
-      - "/check-licenses"
     schedule:
       interval: "daily"
```
SMoraisAnsys: Given how often we get updates on the repo, I'd suggest to update this configuration (and the others) to be weekly.

.github/dependabot.yml
```diff
@@ -20,6 +16,21 @@ updates:
...
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+      include:
+        - "*"  # Include all dependencies in cooldown
+    exclude-paths:
+      - "**/requirements.txt"
+    commit-message:
+      prefix: "build(actions-dependencies)"
```
SMoraisAnsys:
```suggestion
      prefix: "build(uv)"
```

build-library/action.yml
```diff
@@ -200,13 +200,8 @@ runs:
...
       env:
         INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --system' || 'python -m pip install' }}
-        USE_UV: ${{ inputs.use-uv }}
       run: |
-        if [ "${USE_UV}" == "true" ]; then
-          ${INSTALL_COMMAND} twine
-        else
-          ${INSTALL_COMMAND} build twine
-        fi
+        ${INSTALL_COMMAND} -r ${GITHUB_ACTION_PATH}/requirements.txt
```
SMoraisAnsys: Open question: should we keep the previous logic and have two requirements file here ? One for uv and without `build`, one for pip and with `build`. This will likely require extra changes in other locations so it might lead to a trade off between clarity/simplicity and practical need.

.github/workflows/ci_cd_pr_poetry.yml
```diff
@@ -142,8 +143,6 @@ jobs:
             library: pytwin
           - name: pyacp
             library: ansys-acp-core
-          - name: pygranta
```
SMoraisAnsys, line 145: Why are some repos removed ?

code-style/action.yml
```diff
@@ -276,19 +284,23 @@ runs:
           python -m pip install .
         fi
 
-    - name: "Install pre-commit v${{ inputs.pre-commit-version }}"
-      if: inputs.use-prek == 'false'
+    - name: "Install action dependencies"
```
SMoraisAnsys, line 287: I would advocate for 2 requirements file here. Pre-commit comes with multiple dependencies, prek is standalone.

.github/dependabot.yml
```diff
@@ -20,6 +16,21 @@ updates:
...
+  - package-ecosystem: "uv"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+      include:
+        - "*"  # Include all dependencies in cooldown
+    exclude-paths:
+      - "**/requirements.txt"
```
SMoraisAnsys: This might need to be updated if multiple requirements file are created in some actions (cf incoming comments)

SMoraisAnsys (changes requested):
Good job @moe-ad I'm not that familiar with `poetry` but your changes seem to be as fine as possible considering what we want to achieve. Maybe it could be worth to reach maintainers leveraging `poetry` to get their feedback ? Tbh I would have loved to be able to simply use the uv.lock file and not using the new added workflow but... it is very convenient given that we work with different eco systems.

I've left a few minor comments. I think the one on installing `build` could have been "discarded" but the one on `prek` / `pre-commit` shouldn't. And since we open the floor with the later, I would do both :D

.github/workflows/ci_cd_pr_poetry.yml
moe-ad: `pygranta` switched from poetry to uv. Others like `pymechanical` also switched to some other backend.

.github/dependabot.yml
moe-ad: Taken care of in https://github.com/ansys/actions/pull/1321/commits/218f12edeb92229bb0bf915ff8ae1025bc1bfdcb.

code-style/action.yml
moe-ad: Taken care of in https://github.com/ansys/actions/pull/1321/commits/218f12edeb92229bb0bf915ff8ae1025bc1bfdcb.

.github/dependabot.yml
moe-ad: Taken care of in https://github.com/ansys/actions/pull/1321/commits/218f12edeb92229bb0bf915ff8ae1025bc1bfdcb.

build-library/action.yml
moe-ad: I think having a single requirement file in this case is ok.

moe-ad:
@SMoraisAnsys thanks for the review. I have addressed your feedbacks. Please have another look.

.github/workflows/ci_cd_dependabot.yml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+            python-version: ${{ env.MAIN_PYTHON_VERSION }}
+            use-cache: false
+            provision-uv: true
+            prune-uv-cache: false
+
+      - name: "Update requirement.txt files"
+        run: |
+          echo "Directories containing requirements files:"
+          echo "==========================================="
```
SMoraisAnsys:
```suggestion
          echo "=========================================="
```

.github/workflows/ci_cd_dependabot.yml
```diff
@@ -0,0 +1,96 @@
...
+
+              echo "  Processing: $file_cleaned (group: $group_name)"
+
+              # Run uv export command
+              uv export --format requirements.txt --group "$group_name" --output-file "$file_cleaned" --no-hashes
+          done
+
+      - name: Commit and push requirements files if changed
+        env:
+          REF_NAME: ${{ github.ref_name }}
```
SMoraisAnsys:
```suggestion
```
Not sure that this is needed, is it ?

build-library/action.yml
SMoraisAnsys: Since the changes have already been adapted to accomodate precommit/prek requirements logic, I would use two separate files here too. Here, we could have requirements-base.txt (with all the required dependencies excecpt `build`) and requirements-pip.txt (with only `build`). But I would prefer to have things in a single file so requirements-pip and requirements-uv.

That would make uv work with its own logic and that's also why we did split the step in a recent commit.
@ansys/pyansys-core Thoughts on that ?

RobPasMue: If we decide to split them.. I would create the files package-manager based.. so requirements-pip and requirements-uv - and make them independent/standalone.

But yeah... is it practical? 😆

.github/workflows/ci_cd_dependabot.yml
moe-ad: Thanks for catching that.

build-library/action.yml
moe-ad: Added in https://github.com/ansys/actions/pull/1321/commits/9d461195e16a678e7c43094d89cf2a46a38f6946.

FYI I didn't want to split because there are too many dependencies in common to both cases (32 out of 36), which is a different scenario to the prek/pre-commit case. That's not to say I do not appreciate the benefits of splitting.

MaxJPRey (approved):

SMoraisAnsys (approved):
LGTM, @moe-ad can you give it a last try since there were a few changes added ?

moe-ad:
> LGTM, @moe-ad can you give it a last try since there were a few changes added ?

I just added https://github.com/ansys/actions/pull/1321/commits/55004351a78d4d80f6d0bad681e7e7c0a7f37c1c to extend our testing and https://github.com/ansys-internal/test-repo-moe/pull/49 was also triggered and modified using the latest version of ci_cd_dependabot.yml.

JSYK I had initially extended build-wheelhouse testing in https://github.com/ansys/actions/pull/1321/commits/1b6ae5bc5f1504d58cfacb6d759db38a4ee95dee and code-style testing in https://github.com/ansys/actions/pull/1321/commits/4959e7c95a73fee0c7ff3f7d4649487172318cc7.

moe-ad:
Current pyansys-geometry code-style failure is due to https://github.com/tox-dev/toml-fmt/issues/355. @RobPasMue I have opened https://github.com/ansys/pyansys-geometry/pull/2813.

Merging this.

PR #1325, opened 2026-05-21
fix: update input tomli-version default value to empty string

SMoraisAnsys:
That input value is not used in the action AND it is showing a warning which doesn't make much sense from the user perspective, see below

```
Warning: The ``tomli-version`` input is deprecated and will be removed in v12. This action now uses the ``tomlkit`` library instead. Please use the ``tomlkit-version`` input accordingly.
```

Changing the default value to `''` should remove the warning and have zero impact on users :)


RobPasMue (approved):
LGTM!

RobPasMue:
We should probably patch-release this - the warning is reaally annoying TBH. @moe-ad could you handle this?

PR #1327, opened 2026-05-22
fix: geometry doc build

RobPasMue:
Might be related to a change I had to do... https://github.com/ansys/pyansys-geometry/commit/a6c49c186234800b1f34e2460f8c34e270f33190

This change fixes the doc build issues but.. do we want to open the door to installing more miktex packages? I know this one specifically might be  needed for many other libs

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -253,7 +253,7 @@ jobs:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           add-pdf-html-docs-as-assets: true
           needs-quarto: true
-          dependencies: 'pandoc texlive-science'
+          dependencies: 'pandoc'
           # sphinxopts: '-j 1 -W --color'
           # TODO: Re-enable warnings as errors after numpydoc issues are fixed
           sphinxopts: '-j 1'
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -253,7 +253,7 @@ jobs:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           add-pdf-html-docs-as-assets: true
           needs-quarto: true
-          dependencies: 'pandoc'
+          dependencies: 'pandoc texlive-science'
           # sphinxopts: '-j 1 -W --color'
           # TODO: Re-enable warnings as errors after numpydoc issues are fixed
           sphinxopts: '-j 1'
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_cd_pr.yml
```diff
@@ -253,7 +253,7 @@ jobs:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           add-pdf-html-docs-as-assets: true
           needs-quarto: true
-          dependencies: 'pandoc'
+          dependencies: 'pandoc texlive-science'
```
RobPasMue:
```suggestion
          dependencies: 'pandoc'
```
D*** it... texlive* packages are not available on choco...

doc-build/action.yml
```diff
@@ -290,7 +290,7 @@ runs:
 
     - name: Documentation build (Windows)
       if: ${{ runner.os == 'Windows' }}
-      uses: ansys/actions/_doc-build-windows@main
+      uses: ./../_doc-build-windows
```
RobPasMue:
```suggestion
      uses: ansys/actions/_doc-build-windows@fix/geom-doc-build
```

doc-build/action.yml
```diff
@@ -290,7 +290,7 @@ runs:
 
     - name: Documentation build (Windows)
       if: ${{ runner.os == 'Windows' }}
-      uses: ansys/actions/_doc-build-windows@main
+      uses: ansys/actions/_doc-build-windows@fix/geom-doc-build
```
RobPasMue:
```suggestion
      uses: ansys/actions/_doc-build-windows@main
```
To be reverted before merging

AlejandroFernandezLuces (commented):
We could expose a parameter that allows users to install the "optional" dependencies. I prefer the opt in version, specially to reduce action run time.

RobPasMue:
> We could expose a parameter that allows users to install the "optional" dependencies. I prefer the opt in version, specially to reduce action run time.

Yeah works for me - I can expose something for this... but let's keep in mind that this only affects Windows doc-builds. I'll have to be explicit on the naming

RobPasMue:
I think https://github.com/ansys/actions/pull/1327/commits/472f10588782648c3e3c2f8281a958c7df22e5d4 might be enough. This font is only needed for quarto builds in fact so... it might do the trick. If somebody needs it for something else we can expose a new parameter. But I'd avoid adding a new one... it's getting quite crowded already the inputs section :lau

AlejandroFernandezLuces (approved):
Works for me!

PR #1329, opened 2026-05-25
feat: pre-commit automatic update using dependabot

RobPasMue:
As title says - also took advantage to upgrade the deps and freeze the commits

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -119,3 +119,20 @@ updates:
           - dependabot/fetch-metadata
         exclude-patterns:
           - "ansys/actions*"
+
+  - package-ecosystem: "pre-commit"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7 # semver-* options are not applicable for pre-commit
+      exclude:
+        - "ansys/pre-commit-hooks/*"
+    labels:
+      - "dependencies:pre-commit"
+    commit-message:
+      prefix: "build(pre-commit)"
+    groups:
+      pre-commit-hooks:
+        patterns:
+          - "*"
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -119,3 +119,20 @@ updates:
...
         exclude-patterns:
           - "ansys/actions*"
+
+  - package-ecosystem: "pre-commit"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7 # semver-* options are not applicable for pre-commit
+      exclude:
```
SMoraisAnsys, line 131:
```suggestion
      include:
        - "*"  # Include all dependencies in cooldown
      exclude:
```
Not sure if it's necessary but I would keep things consistent over ecosystems (both actions and pip use it)

.pre-commit-config.yaml
```diff
@@ -1,23 +1,23 @@
...
 
 - repo: https://github.com/psf/black
-  rev: 25.1.0
+  rev: 87928e6d6761a4a6d22250e1fee5601b3998086e  # frozen: 26.5.1
   hooks:
   - id: black
 
 - repo: https://github.com/pycqa/isort
-  rev: 6.0.1
+  rev: dac090ce4d9ee313d086e2e89ab1acb8c2664fa1  # frozen: 9.0.0a3
```
SMoraisAnsys:
```suggestion
  rev: a333737ed43df02b18e6c95477ea1b285b3de15a  # frozen: 8.0.1
```
I'd prefer to use a stable version instead

SMoraisAnsys (approved):
LGTM

MaxJPRey (approved):

.github/dependabot.yml
RobPasMue: That's the default I believe - but all good

PR #1339, opened 2026-05-28
fix: announcement reset on stable release

SMoraisAnsys:
Instead of recreating the announcement.html file from scratch, this PR aims at keeping it's existing content and only update a specific part of the file if it exists (both the file & the targeted part).

The reason for this change is that every stable release removes the banner "This version supports [secure transport modes](https://dev.docs.pyansys.com/how-to/grpc-api-packages.html#securing-connections)." and this gets only added every Saturday through automation workflow.

If you think that's enough and a "hole" of a week (maximum) isn't an issue then we can keep things as they are and close this PR but if we ever have to add new announcement, this mechanism might be good to have as it leaves user with the ability to keep some of their own announcements.

### DIFF
--- a/doc-deploy-stable/action.yml
+++ b/doc-deploy-stable/action.yml
@@ -413,12 +413,12 @@ runs:
       shell: bash
       run: |
         cat > announcement-managed.html <<EOF
-        <!-- BEGIN ANSYS_LATEST_STABLE_ANNOUNCEMENT -->
+        <!-- BEGIN PROJECT_LATEST_STABLE_ANNOUNCEMENT -->
         <p style="background-color: #F5B0AE; width: 100%; padding: 1rem 0rem;">
           You are not viewing the most recent version of this documentation.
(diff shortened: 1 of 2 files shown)


moe-ad:
Sounds like a great idea to me.

RobPasMue:
So the banner addition was done in that intent on purpose...

We might decide, at some point in time, that announcing the "secure" connection banner is not needed anymore (since it has gone from "latest news" to something taken for granted).

The current approach consequently allows teams to continue to do releases and if at some point we disable the banner addition, they don't need to do anything. The "secure connection" banner doesn't get added and life moves on. Just like we want it to be.

However, with your approach... the stable banner will continue to show up "forever" unless we actively go to the released gh-pages and remove it ourselves... which is going to be a pain.

I agree that we should have a way of adding banners consistently for projects... but the way the projects should leverage it is via: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html

SMoraisAnsys:
Closing following https://github.com/ansys/actions/pull/1339#issuecomment-4590930501

PR #1341, opened 2026-05-28
chore: remove deprecation notes from safety scan output

### LINKED ISSUE #641: Use scan command in favor of check one for bandit

moe-ad:
Closes #641.


moe-ad:
Tested here: https://github.com/ansys-internal/test-repo-moe/actions/runs/26634222768/job/78490312364?pr=52

check-vulnerabilities/action.yml
```diff
@@ -540,8 +540,20 @@ runs:
...
+        /\+==.*=+\+/ {
+            if (buffering) { for (i=0;i<nb;i++) print buf[i] }
+            buf[0]=$0; nb=1; buffering=1; next
+        }
+        buffering && /^[[:space:]]*$/ { buf[nb++]=$0; next }
+        buffering && /DEPRECATED:.*has been DEPRECATED/ { buffering=0; skipping=1; next }
+        buffering { for (i=0;i<nb;i++) print buf[i]; buffering=0; print; next }
+        { print }
+        END { if (buffering) for (i=0;i<nb;i++) print buf[i] }
+        '
```
RobPasMue: What an awk command... hard to follow to be honest. But I can see the output works.

RobPasMue (approved):
Despite the fact that the awk command seems like black magic, LGTM!

check-vulnerabilities/action.yml
moe-ad: > What an awk command... hard to follow to be honest. But I can see the output works.

... missed the opportunity to use 'awkward'.

I didn't craft this just by myself incase you were wondering... :)
But it was the simplest solution that caught both deprecation blocks (at the beginning and at the end) while leaving the formatting of the rest of the output intact.

RobPasMue: > ... missed the opportunity to use 'awkward'.

Should've thought about it 😆 

> But it was the simplest solution that caught both deprecation blocks (at the beginning and at the end) while leaving the formatting of the rest of the output intact.

All good!

moe-ad: Actually I just noticed something because your comment made me think twice. I was testing locally without using the `-o bare` CLI flag, which was what necessitated such a convoluted awk command. I should be able to simplify things, let me have another go at it.

moe-ad: Added https://github.com/ansys/actions/pull/1341/changes/629f3a23cc9d68a01e727eee020f602fec006b99 which is much better. Tested here: https://github.com/ansys-internal/test-repo-moe/actions/runs/26640470043/job/78511844481?pr=52.

RobPasMue: Oh my - joy teardrops running down my face! Thanks @moe-ad !

PR #1347, opened 2026-05-29
refactor: update project+targets installation logic in check-licenses action

### LINKED ISSUE #1293: Use poetry for installation in `ansys/actions/check-licenses` for poetry-based projects.
As the title states. Project and target installation in the check-licenses action should be consistent with other actions wrt to the poetry backend.

moe-ad:
Closes #1293.

### DIFF
--- a/.github/workflows/ci_cd_pr_flit.yml
+++ b/.github/workflows/ci_cd_pr_flit.yml
@@ -179,7 +179,6 @@ jobs:
           operating-system: ${{ matrix.os }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           attest-provenance: true
-          check-licenses: false
 
   test-build-library-flit-projects:
     name: "Test build-library action for ${{ matrix.repository.name }} project"
--- a/.github/workflows/ci_cd_pr_flit.yml
+++ b/.github/workflows/ci_cd_pr_flit.yml
@@ -179,6 +179,7 @@ jobs:
           operating-system: ${{ matrix.os }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           attest-provenance: true
+          check-licenses: false
 
   test-build-library-flit-projects:
     name: "Test build-library action for ${{ matrix.repository.name }} project"
@@ -222,6 +223,43 @@ jobs:
           library-name: ${{ matrix.repository.library }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           attest-provenance: true
+
+  test-check-licenses-flit-projects:
+    name: "Test check-licenses action for ${{ matrix.repository.name }} project"
+    runs-on: ubuntu-latest
+    needs: test-build-library-flit-projects
+    strategy:
+      matrix:
+        repository:
+          - name: pyansys-geometry
+            library: ansys-geometry-core
+          - name: pymapdl
+            library: ansys-mapdl-core
+          - name: pyfluent
+            library: ansys-fluent-core
+          - name: pymechanical
+            library: ansys-mechanical-core
+      fail-fast: false
+    steps:
+
+      - name: "Install Git and clone ${{ matrix.repository.name }} repository"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+          repository: ansys/${{ matrix.repository.name }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Clone ansys/actions into a different path"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+          path: actions
+
+      - name: "Check license validity for ${{ matrix.repository.name }}"
+        uses: ./actions/check-licenses
+        with:
+          python-version: ${{ env.MAIN_PYTHON_VERSION }}
+          checkout: false
 
   test-code-style-flit-projects:
     name: "Test code-style action (use-prek: ${{ matrix.use-prek }}) for ${{ matrix.repository.name }} project"
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -459,7 +459,7 @@ runs:
         subject-path: ${{ steps.sanitize.outputs.sanitized_library_name }}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}.zip
 
     - name: Check library's dependencies license
-      uses: ansys/actions/check-licenses@main
+      uses: ansys/actions/check-licenses@refactor/check-licenses-action
       if: ${{ inputs.check-licenses == 'true' }}
       with:
         python-version: ${{ inputs.python-version }}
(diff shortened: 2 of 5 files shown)


moe-ad:
To-do: test changes.

build-wheelhouse/action.yml
```diff
@@ -459,7 +459,7 @@ runs:
         subject-path: ${{ steps.sanitize.outputs.sanitized_library_name }}-v${{ steps.library-version.outputs.library_version }}-${{ steps.specific-target-requested.outputs.wheelhouse_target }}-${{ inputs.operating-system }}-${{ inputs.python-version }}.zip
 
     - name: Check library's dependencies license
-      uses: ansys/actions/check-licenses@main
+      uses: ansys/actions/check-licenses@refactor/check-licenses-action
```
moe-ad:
```suggestion
      uses: ansys/actions/check-licenses@main
```

To be reverted prior to merging.

check-licenses/action.yml
```diff
@@ -272,43 +276,46 @@ runs:
...
         ACTIVATE_VENV: ${{ steps.virtual-environment-activation-command.outputs.ACTIVATE_VENV }}
         BUILD_BACKEND: ${{ steps.build-system.outputs.BUILD_BACKEND }}
         SKIP_INSTALL: ${{ inputs.skip-install }}
-        INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python' || 'python -m pip install' }}
+        TARGET: ${{ inputs.target }}
       run: |
         ${ACTIVATE_VENV}
+        if [[ "${TARGET}" != '' ]]; then
+          if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
+            echo "Installing extra targets with poetry: ${EXTRA_TARGETS}"
```
SMoraisAnsys: I think `EXTRA_TARGETS`'s definition is missing here, do you mean `TARGET` ? Same comment below

SMoraisAnsys (changes requested):

check-licenses/action.yml
moe-ad: Corrected in https://github.com/ansys/actions/pull/1347/commits/8d5c1304129aa5b335398a077ebd9396deae5a30.

.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -179,6 +179,7 @@ jobs:
           operating-system: ${{ matrix.os }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           attest-provenance: true
+          check-licenses: false
```
RobPasMue: Why are we not testing it here too? It's a different entry point but we should make sure it works as well.. right?

RobPasMue (approved):

.github/workflows/ci_cd_pr_flit.yml
moe-ad: The reason why I stopped testing through here is that `build-wheelhouse` passes `skip-install=true` to `check-licenses`(since the action itself installs the project already), so the testing never hits the core logic we should actually be testing.

It is the main reason why I had to add https://github.com/ansys/actions/pull/1347/commits/efdf6f060879e609cfb9c6e4f0535d08d4fa2f73.

check-licenses/action.yml
```diff
@@ -272,43 +276,46 @@ runs:
...
-        INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python' || 'python -m pip install' }}
+        TARGET: ${{ inputs.target }}
       run: |
         ${ACTIVATE_VENV}
+        if [[ "${TARGET}" != '' ]]; then
+          if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
+            echo "Installing extra targets with poetry: ${TARGET}"
+            EXTRA_TARGETS=${TARGET}
+          else
+            echo "Installing extra targets with pip: ${TARGET}"
```
SMoraisAnsys, line 296:
```suggestion
            echo "Installing extra targets with pip/uv: ${TARGET}"
```

SMoraisAnsys (approved):
LGTM, just left a minor comment

.github/workflows/ci_cd_pr_flit.yml
SMoraisAnsys: ~Can you add a comment on that then ? If we come back here in a year, it would be great to understand why this is programmed in such a way and not having to crawl back commit history :)~
In fact, I think that we should keep the `check-licenses` to true because that's mainly how most of our users use that action. Keeping the direct test to `check-licenses` is fine but this action is not that often directly leveraged.

check-licenses/action.yml
```diff
@@ -272,43 +276,46 @@ runs:
           fi
         fi
 
-    - name: "Update pip and install pip-licenses library"
+    - name: "Install project and pip-licenses"
```
SMoraisAnsys, line 279: Shouldn't the install project and pip-licenses installation be splitted ?

I mean, the step consisting in installing the project (and target) should be dependent on `skip-install`. However, that's not the case of `pip-licenses`.

SMoraisAnsys (changes requested):

check-licenses/action.yml
moe-ad: Combining the steps was intentional. That's why there is still a skip-install check in the logic of the combined step:
image

The reason for the combination is that `poetry install` command is now being run multiple times (up to 3 times in some cases) in actions that have to support poetry based projects, and this is not ideal. Besides unnecessary installation of the project multiple times during action runs, the existing situation also means any poetry related change needs to be done across multiple steps and you also need to track how what you are changing in one step could affect the others.

With the changes I made here, you can easily follow the installation logic, and the poetry projects are only installed once.

I do agree that the name of the step can be improved. Something like "Install pip-licenses and optionally install the project" should be ok?

FYI I plan to extend the changes to other actions in the future.

check-licenses/action.yml
moe-ad: Updated in https://github.com/ansys/actions/pull/1347/commits/84956c823a4cf5e4e52329be56028eaca95d1412.

.github/workflows/ci_cd_pr_flit.yml
moe-ad: Reverted in https://github.com/ansys/actions/pull/1347/commits/84956c823a4cf5e4e52329be56028eaca95d1412.

check-licenses/action.yml
SMoraisAnsys: I get that and maybe I'm missing something but if skip install is true, then you don't end up using poetry to install pip-licenses and other requirements, right ?

moe-ad: 💡 
I see what you mean.
https://github.com/ansys/actions/pull/1347/commits/75ca1625b6f8f13c552b19f359f6cb130aa62480 addresses it.

check-licenses/action.yml
```diff
@@ -272,43 +276,53 @@ runs:
...
+        if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
+          # Add pip-licenses to poetry lockfile, then install everything in one go
           grep -v '^\s*#' "${GITHUB_ACTION_PATH}/requirements.txt" | grep -v '^\s*$' | grep -v '^\s' | while read -r line; do
             poetry add "${line//\'/\"}" --lock
           done
-          poetry install
+          if [[ "${SKIP_INSTALL}" == 'false' ]]; then
+            poetry install --extras "${EXTRA_TARGETS}"
+          else
+            poetry install --no-root --extras "${EXTRA_TARGETS}"
```
SMoraisAnsys, line 312: I think this will fail if no input target is provided because `--extras` should be expecting a value

SMoraisAnsys (changes requested):

check-licenses/action.yml
moe-ad: No it won't, I am 100% sure about this behavior of poetry.
You can also test locally to be convinced.

FYI we already use this pattern in other actions too.

SMoraisAnsys (approved):

moe-ad:
Thanks for your reviews @SMoraisAnsys and @RobPasMue. Much appreciated.

PR #1350, opened 2026-06-01
refactor: updates to group/optional dependencies installation in doc-build action

### LINKED ISSUE #1109: Change default behavior of ``group-dependencies-name`` and ``optional-dependencies-name``
See:

https://github.com/ansys/actions/blob/5eda12d2eb970d38b5f29cf7352872761ab72656/doc-build/action.yml#L183-L202

The problem is that if a user defines only ``group-dependencies-name`` and does not have the optional dependencies for docs.. (i.e. they do not ship the doc target), the workflow fails..

The workflow currently tries to install the optional dependencies target (i.e. by default, doc) no matter what.

We are asking our repos to move away from optional dependencies to groups.. and when they do so, I would expect them to eventually remove the "doc" target.. but in no case I would expect our actions to continue to use the optional-dependency target if the group was defined.

My proposal:

- Change both defaults to empty
- If both of them are kept empty by users, then we default inside our action to the current behavior: ``optional-dependencies-name=doc``
- If any of them have content, that's the one we take.
- If both of them have content, so be it. We continue to use both of them

moe-ad:
Closes #1109.

### DIFF
--- a/doc-build/action.yml
+++ b/doc-build/action.yml
@@ -189,8 +189,9 @@ inputs:
       Any valid install targets/extras defined in the ``pyproject.toml`` file, or the suffix
       of a requirement file. Multiple targets can be specified as a comma-separated
       list (space-separated for poetry-based projects). The associated dependencies are installed
-      before building the documentation. Default value is ``doc``.
-    default: 'doc'
+      before building the documentation. Default value is ``''``. If neither this nor
+      ``group-dependencies-name`` is provided, the action defaults to ``doc``.
+    default: ''
     required: false
     type: string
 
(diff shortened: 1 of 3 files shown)


RobPasMue (approved):
We should probably document this in the migration guide too just in case - other than that LGTM

moe-ad:
@RobPasMue thanks for the review. Have a look at https://github.com/ansys/actions/pull/1350/commits/c9745b1f382421d6f24c7a390027f6beaec08f8e for the updated migration guide.

RobPasMue (approved):

PR #1357, opened 2026-06-05
feat: add uv lock check in build-library

SMoraisAnsys:
This PR enhances the build verification process for Python libraries by adding a check to ensure that the `uv.lock` file, if present, is up to date with the project's dependencies. This helps maintain consistency between the lock file and actual dependencies.

The idea if very open and could be rejected / updated to work with an input / ... Let me know if you think it make sense to add it here :)

Successful workflow here https://github.com/ansys/pyaedt-toolkits-common/actions/runs/27014823110/job/79727057353
Expected failing workflow here https://github.com/ansys/pyaedt-toolkits-common/actions/runs/27014432436/job/79725757899

The context of the failing workflow is the addition of `ansys-tools-common` to the main dependencies and the failure to remember about updating the lock file. The CI worked because the package was previously a transitive dependency so we ended up installing the package anyways but the lock file wasn't sync.

### DIFF
--- a/build-library/action.yml
+++ b/build-library/action.yml
@@ -26,7 +26,9 @@ name: |
 description: |
   Verifies if a Python library builds properly. As a result of a successful
   building process, wheel and source distribution artifacts are generated. This
-  action is expected to be used as an independent job.
+  action is expected to be used as an independent job. When a project contains
+  a uv lock file, it is also verified during the build process to ensure it is up to
+  date with the project dependencies.
 
   .. note::
 
@@ -226,6 +228,13 @@ runs:
       if: inputs.validate-build == 'true'
       run: |
         python -m twine check dist/*
+
+    - name: "Check if the uv lock file is up to date with respect to project dependencies (if it exists)"
+      if: inputs.use-uv == 'true' && hashFiles(format('{0}/uv.lock', inputs.working-directory)) != ''
+      shell: bash
+      env:
+        WORKING_DIRECTORY: ${{ inputs.working-directory }}
+      run: uv lock --check --project "${WORKING_DIRECTORY}"
 
     - name: "Upload distribution artifacts to GitHub artifacts"
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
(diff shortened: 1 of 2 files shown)


RobPasMue (changes requested):
Shouldn't this be the responsibility of the project maintainers to keep it up to date? Also, shouldn't dependabot handle this as part of the upgrade checks? I'm confused by the need for this change...

RobPasMue:
My point is that if a project is using uv to manage their dependencies - because that's their build system - they should make sure that their lock file is up to date and in a healthy state. I don't see why we should be checking this in an action... the build-library action should be in charge of building the library distributable (wheel).

Otherwise, by your same reasoning, we would have to do the same for poetry.. and for any other build system that has a lock file.

SMoraisAnsys:
> Shouldn't this be the responsibility of the project maintainers to keep it up to date? Also, shouldn't dependabot handle this as part of the upgrade checks? I'm confused by the need for this change...

I'm not sure if dependabot can catch this kind of behavior. It depends on how it works behind the scenes but you might be right. Also, I agree that this requirement should be delegated to maintainers.

I just remembered that we have a "Check build health" step so I though about adding that other step too. Note that I wouldn't mind about the same check for poetry :)

RobPasMue:
> I'm not sure if dependabot can catch this kind of behavior. It depends on how it works behind the scenes but you might be right. Also, I agree that this requirement should be delegated to maintainers.

When your dependabot config is pointing to the ``uv`` or ``poetry`` environments it  updates the lock files - I can share offline some examples

> I just remembered that we have a "Check build health" step so I though about adding that other step too. Note that I wouldn't mind about the same check for poetry :)

The build health step that we already have is to check that the distributed wheel has been successfully generated. We perform a ``twine check`` operation - which is intended for this action since we are building the library. I think the health check you are trying to implement is really that the repo's lock file (which is used for development purposes really) is healthy.. and has little to do with the distribution of the library IMO.

But I'm happy to debate it

SMoraisAnsys:
No debat required, I wasn't too convinced myself which is why I mentioned to reject the idea :D I could have opened a discussion / issue but the changes were trivial so I went with that approach.

Since I'm getting convinced that dependabot should do the trick (even if it consist of iterating packages on at a time) we should be good. Closing this :)

PR #1364, opened 2026-06-16
feat: allow us uv lock file

SMoraisAnsys:
This PR can be removed at any moment and the approach followed in #1088 can be used (leveraging lock file if detected).

However, leveraging uv lock file has some points to adress and I'd like to leave that PR as draft (unless we want to keep it) to illustrate a few things.


jorgepiloto:
Seems like a spurious issue with Windows. Thanks for fixing this, @SMoraisAnsys.

PR #1374, opened 2026-06-22
feat: doc-deploy-custom-path action

### LINKED ISSUE #921: Allow user to deploy docs to subdirectory
This feature has been requested by multiple users, including docs team and solutions. Right now, we are always publishing to the root directory of the `gh-pages` branch.

moe-ad:
Closes #921.

This PR adds a new action that allows users to deploy to custom directories. It doesn't do much other the following:
- Exposes an input through which a custom deployment path can be passed.
- Validates this input to ensure it can't conflict with what already exists on the gh-pages branch in any way. This basically isolates the files of other doc-deploy actions from potential conflicts.
- Call peaceiris action with a sane set of inputs

If you think there are other functionalities the action should expose, please feedback is welcome @ansys/pyansys-core.


moe-ad:
Tested here: https://github.com/ansys-internal/test-repo-moe/commit/fe1aeecd32fc616a16e71863a48a7ebb81d5bc00
And deployed successfully: https://glowing-adventure-e4rq83p.pages.github.io/nightly/

doc-deploy-custom-path/action.yml
```diff
@@ -0,0 +1,329 @@
...
+
+    # ------------------------------------------------------------------------
+
+    - name: "Validate 'custom-path' input"
+      shell: bash
+      env:
+        CUSTOM_PATH: ${{ inputs.custom-path }}
+      run: |
+        if [[ -z "${CUSTOM_PATH}" || "${CUSTOM_PATH}" == "." ]]; then
+          echo "Error: 'custom-path' must not be empty or '.'." >&2
```
moe-ad:
```suggestion
          echo "Error: 'custom-path' must not be empty or '.'."
```

doc-deploy-custom-path/action.yml
```diff
@@ -0,0 +1,329 @@
...
+          exit 1
+        fi
+        if [[ "${CUSTOM_PATH}" == /* ]]; then
+          echo "Error: 'custom-path' must not be an absolute path." >&2
```
moe-ad:
```suggestion
          echo "Error: 'custom-path' must not be an absolute path."
```

doc-deploy-custom-path/action.yml
```diff
@@ -0,0 +1,329 @@
...
+          exit 1
+        fi
+        if [[ "${CUSTOM_PATH}" == *..* ]]; then
+          echo "Error: 'custom-path' must not contain '..'." >&2
```
moe-ad:
```suggestion
          echo "Error: 'custom-path' must not contain '..'."
```

doc-deploy-custom-path/action.yml
```diff
@@ -0,0 +1,329 @@
...
+          exit 1
+        fi
+        reserved=("version" "pull")
+        for r in "${reserved[@]}"; do
+          if [[ "${CUSTOM_PATH}" == "${r}" || "${CUSTOM_PATH}" == "${r}/"* ]]; then
+            echo "Error: 'custom-path' must not use the reserved path '${r}'." >&2
```
moe-ad:
```suggestion
            echo "Error: 'custom-path' must not use the reserved path '${r}'."
```

RobPasMue (approved):
I struggle to understand the need for the action - given the situation that one would want to deploy to a specific directory and simply ignore all the additional tooling... in my case I would simply use peaceiris TBH. But fine by me.

moe-ad:
@RobPasMue I don't have a fixed opinion on having a new action. An alternative could be adding some instruction somewhere instructing people on how to use peaceiris directly in a way that won't conflict with existing documentation actions. Maybe in the dev guide? Or in the actions docs?

PR #1401, opened 2026-07-01
chore: ignore patch updates for uv package ecosystem

moe-ad:
We do not want dependabot to change the pinned versions in **pyproject.toml** from `major.minor` to `major.minor.patch` (see [1](https://github.com/ansys/actions/pull/1395) and [2](https://github.com/ansys/actions/pull/1368) as examples). With the help of @SMoraisAnsys, we came across the configuration option proposed in this PR, which seems to address our concern.

The hope is that dependency resolution still takes place and patch updates of dependencies still happens for the lock file (and consequently the `requirements.txt` files). We will need to monitor this in the coming weeks.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,6 +32,9 @@ updates:
       prefix: "build(uv)"
     labels:
       - "dependencies:requirements"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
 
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):

MaxJPRey (approved):
I think it is the right approach.

klmcadams (approved):

SMoraisAnsys (approved):

PR #1402, opened 2026-07-07
fix: improvements to hk-migrate-fork-pr

moe-ad:
Feedback from @lukequeenan-ansys after trying out the `hk-migrate-fork-pr` action on https://github.com/ansys/pylumerical/pull/69:

> I got an opportunity to try out the migration feature - it's mostly working great, although I noticed two things worth documenting for next time:
>- The branch it tries to create uses the migration prefix which doesn't match the conventional commit rules we had in place - that was minor and I updated the rules to allow it, but wondering if it wouldn't be better to add a bypass for the bot so people can't also use that branch name?
>- When making the PR, it also fails to run since the PR name doesn't match either, I also fixed this by putting the migrated part at the end, docs: example- HEAT thermal waveguide tuner getting started example (migrated (PR 69)). But maybe it would be best to allow migration in the PR name as well?
>- It also looks like the bot ends up telling itself to perform a sync accidentally and then fails because it's not in the team

My bad for the first two, I should have taken that into consideration during the initial implementation. For the third, the cause is the presence of `@pyansys-ci-bot sync` in [the comment](https://github.com/ansys/pylumerical/pull/69#issuecomment-4898023554) the bot posts upon successful migration (this issue wasn't caught during testing because `pyansys-core` was the team-slug used, which `pyansys-ci-bot` is a member of).

This PR fixes these issues, while maintaining backwards compatibility so this can go into the next patch release.

WIP. Needs some form of testing before we can merge.


moe-ad:
#### Test

To reproduce a scenario where the backwards compatibility logic can be tested, I first migrated a PR ([original](https://github.com/ansys/ansys-all-members/pull/55) --> [migrated](https://github.com/ansys/ansys-all-members/pull/56)) with the current release i.e. `v10.3.3`:
image

Followed by editing the migrated PR title manually:
image

Upon resyncing using the updates on this PR, the behavior I expected was that the branch and PR title will be renamed but that wasn't the case. Instead, the initial PR got closed automatically upon branch rename and a [new PR](https://github.com/ansys/ansys-all-members/pull/58) was opened:
image

I looked into this and it is indeed how GH behaves when branches are renamed (see https://github.com/orgs/community/discussions/4453).

moe-ad:
Backwards compatibility is proving difficult to achieve. The best case scenario will be for the action logic to only rename existing PR titles and keep working with existing branches, but use the new branch naming structure for new PRs.

I also think a much easier approach will be to sync with @lukequeenan-ansys. He is the first (and currently the only) person leveraging this action. Once he merges https://github.com/ansys/pylumerical/pull/69, we can quickly release a patch afterwards so that there will be no impact. That will also help us avoid adding new logic to the action and having to remove them in the near future. What do you think @RobPasMue @SMoraisAnsys?

lukequeenan-ansys (approved):

hk-migrate-fork-pr/action.yml
```diff
@@ -364,6 +380,62 @@ runs:
         fetch-depth: 0
         persist-credentials: true # zizmor: ignore[artipacked]
 
+    - name: "Migrate legacy naming to conventional commits"
+      # TODO: Remove this step once all legacy migration/pr-N branches have been renamed.
```
lukequeenan-ansys: I've merged the PR we had open using this method, so perhaps safe to remove now

moe-ad:
Reverted backwards compatibility logic, merging, and performing a patch release with this.

klmcadams (approved):

PR #1409, opened 2026-07-10
chore: update release testing

### LINKED ISSUE #1407: Update upload artifacts to test PyPI
A mechanism has been added into PyPI to check that after a certain amount of time, future uploads to an existing release are disabled. The current value is 14 days, see this [PR](https://github.com/pypi/warehouse/pull/19727).

This has an impact on our CI workflows since we have test for releasing our dummy package using trusted publisher approach. [Here](https://github.com/ansys/actions/actions/runs/29001908522/job/86072635706?pr=1402#step:3:207) is an example of such a failure, see below 

Image

I've checked https://github.com/pypi/warehouse and there is no internal logic to split behavior depending on the domain used so I'm assuming there won't be a pass for the test.pypi.org domain. A solution would be to rework how we are testing. For example, we could inspire from https://github.com/pypa/gh-action-pypi-publish which performs tests using `devpi`, see this [workflow](https://github.com/pypa/gh-action-pypi-publish/blob/unstable/v1/.github/workflows/reusable-smoke-test.yml#L138), which seems to be a python server that can be used as a package registry.

---

Related issue(s): https://github.com/pypi/warehouse/issues/20251

moe-ad:
Closes #1407 (see for full description of the issue this PR solves).

Summary of tasks completed:
  - [x] License metadata of test packages has been updated.
  - [x] A `devpi` directory has been added (to `.ci` directory), containing necessary docker files to start a devpi server. I decided to create our own files and not leverage [muccg/devpi](https://github.com/muccg/docker-devpi) because:
    - [x] The contents of the docker files in [muccg/devpi](https://github.com/muccg/docker-devpi) are quite outdated. The published image on ghcr is correspondingly old.
    - [x] The docker file contains extra setup that I consider unnecessary for our use case.
    - [x] The repo has been archived, so there is no way for things to get updated. Moreover, we are easily able to modify things by having our own files
  - [x] Updated the workflow files to leverage devpi for testing

### DIFF
--- a/.github/workflows/ci_cd_pr_flit.yml
+++ b/.github/workflows/ci_cd_pr_flit.yml
@@ -113,13 +113,22 @@ jobs:
           checkout: false
 
   test-release-flit:
-    name: "Test releasing ansys-actions-flit package using trusted publishing"
+    name: "Test releasing ansys-actions-flit package to local devpi server"
     runs-on: ubuntu-latest
     needs: test-build-library-flit
     permissions:
-      id-token: write # Needed for trusted publishing OIDC
       contents: read
     steps:
+
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        persist-credentials: false
+
+    - name: "Start devpi server"
+      run: docker compose -f .ci/devpi/docker-compose.yml up -d --build
+
+    - name: "Wait for devpi server to be ready"
+      run: timeout 120 bash -c 'until curl -sf http://localhost:3141/; do sleep 2; done'
 
     - name: "Download distribution artifacts"
       uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
@@ -127,13 +136,17 @@ jobs:
         name: ${{ env.LIBRARY_NAME }}-artifacts
         path: dist
 
-    - name: "Upload artifacts to test PyPI using trusted publisher"
-      uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
-      with:
-        repository-url: "https://test.pypi.org/legacy/"
-        print-hash: true
-        skip-existing: true
-        verbose: true
+    - name: "Upload artifacts to devpi"
+      env:
+        TWINE_USERNAME: root
+        TWINE_PASSWORD: devpipass
+        TWINE_REPOSITORY_URL: http://localhost:3141/root/public/
+      run: |
+        pip install twine
+        twine upload dist/*
+
+    - name: "Verify package is installable from devpi"
+      run: pip install --index-url http://localhost:3141/root/public/+simple/ --no-deps ${{ env.LIBRARY_NAME }}
 
   test-build-wheelhouse-flit-projects:
     name: "Test build-wheelhouse action for ${{ matrix.repository.name }} project on ${{ matrix.os }}"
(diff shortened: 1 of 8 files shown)


.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -113,27 +113,42 @@ jobs:
...
-      with:
-        repository-url: "https://test.pypi.org/legacy/"
-        print-hash: true
-        skip-existing: true
-        verbose: true
+    - name: "Upload artifacts to devpi"
+      env:
+        TWINE_USERNAME: root
+        TWINE_PASSWORD: devpipass
+        TWINE_REPOSITORY_URL: http://localhost:3141/root/public/
```
klmcadams, line 143: Is this okay to have in plaintext?

moe-ad: Yes, since it is a transient server whose lifetime is only the job duration and only accessible from the runner. So no issues.

clatapie (approved):

ecoussoux-ansys (approved):
LGTM

.github/workflows/ci_cd_pr_poetry.yml
```diff
@@ -114,27 +114,42 @@ jobs:
...
+        TWINE_PASSWORD: devpipass
+        TWINE_REPOSITORY_URL: http://localhost:3141/root/public/
+      run: | # zizmor: ignore[use-trusted-publishing]
+        pip install -U twine packaging
+        twine upload dist/*
+
+    - name: "Verify package is installable from devpi"
+      env:
+        LIBRARY: ${{ env.LIBRARY_NAME }}
+      run: pip install --index-url http://localhost:3141/root/public/+simple/ --no-deps "${LIBRARY}"
```
MaxJPRey, line 152: `--no-deps` skips resolving` importlib-metadata >=4.0`, so a broken dependency spec would sail through. Why not use `--extra-index-url` pointing at PyPI (so real deps still resolve) and drop --no-deps?

MaxJPRey:
Regarding the duplication between the two workflows (`ci_cd_pr_flit.yml` and `ci_cd_pr_poetry.yml`) The test-release-* jobs are similars between flit and poetry except for the job name / needs. 

Could we either:

Extract into a **reusable workflow** (.github/workflows/_test-release-devpi.yml) called with uses: + with: { library-name, needs-job }.
**Or**
**composite action** under .ci/devpi/action.yml that encapsulates "checkout devpi files → start server → wait → upload → verify".

.github/workflows/ci_cd_pr_poetry.yml
moe-ad: Good catch. The problem is that there is no way to specify a search prioritization of the indexes (as documented [here](https://pip.pypa.io/en/latest/cli/pip_install/#finding-packages)). We already have a `10.4.dev0` version on test PyPI, so pip might pull that instead.

We can update it when we a doing a bump to `10.5.dev0`. I have opened https://github.com/ansys/actions/issues/1411 so we do not forget that

moe-ad:
> Regarding the duplication between the two workflows (`ci_cd_pr_flit.yml` and `ci_cd_pr_poetry.yml`) The test-release-* jobs are similars between flit and poetry except for the job name / needs.
> 
> Could we either:
> 
> Extract into a **reusable workflow** (.github/workflows/_test-release-devpi.yml) called with uses: + with: { library-name, needs-job }. **Or** **composite action** under .ci/devpi/action.yml that encapsulates "checkout devpi files → start server → wait → upload → verify".

Thanks for the suggestion, I have opened https://github.com/ansys/actions/pull/1412 to address it.

Issue #1391, opened 2026-06-30
deploy-docs-dev fails silently


inimaz:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

I think this is a corner case. The moment I added the `deploy-docs-pr` to the ci, the `deploy-docs-dev` fails with `Error: Process completed with exit code 1.` and no logs as for why. 

### 📝 Steps to reproduce

- `doc-build` + `deploy-docs-dev` works. 
- `doc-build` + `deploy-docs-pr`  +  `deploy-docs-dev`  does not work.

# Script to reproduce the issue
```yaml
name: GitHub CI
on:
  pull_request:
  workflow_dispatch:
  push:
    tags:
      - "*"
    branches:
      - main
env:
  MAIN_PYTHON_VERSION: '3.11'
  DOCUMENTATION_CNAME: example.com

jobs:

  docs-build:
    name: Building Documentation
    runs-on: ubuntu-latest

    steps:
      - name: Build documentation
        uses: ansys/actions/doc-build@a5b0351dc277d68c2dc1beca2d2674c7dfa4fc55 # v10.3.3
        with:
          python-version: ${{ env.MAIN_PYTHON_VERSION }}
          group-dependencies-name: "doc"
          skip-dependencies-cache: true

  deploy-docs-pr:
    name: Deploy PR docs
    runs-on: ubuntu-latest
    needs: docs-build
    if: github.event_name == 'pull_request'
    permissions:
      contents: write
      pull-requests: write
    steps:
      - name: Checkout repository
        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
      - name: Deploy PR documentation
        uses: ansys/actions/doc-deploy-pr@a5b0351dc277d68c2dc1beca2d2674c7dfa4fc55 # v10.3.3
        with:
          cname: ${{ env.DOCUMENTATION_CNAME }}
          token: ${{ secrets.GITHUB_TOKEN }}
          bot-user: ${{ secrets.PYANSYS_CI_BOT_USERNAME }}
          bot-email: ${{ secrets.PYANSYS_CI_BOT_EMAIL }}

  deploy-docs-dev:
    name: Deploy development docs (on main branch)
    runs-on: ubuntu-latest
    needs: docs-build
    if: github.ref == 'refs/heads/main'
    permissions:
      contents: write
    steps:
[...]

PR #1410, opened 2026-07-10
feat: check job permissions and warn if not enough

inimaz:
Related to #1391.

- Goal is to warn if any of the known needed permissions are not there when starting to run the workflow. And if so, add a warning.
- `probe` function tests the token against an action in the API that uses the given permission.
- `write` permissions not checked for now... Since one has to find a harmless action to run per each permission. We can think of something in future PRs if needed.
- Adding it to `doc-deploy-dev` as an example.


moe-ad:
@inimaz can you confirm that this works through some minimal testing?

FYI, to test:
- Change the `_check-github-permissions` call on this branch to `ansys/actions/_check-github-permissions@feat/check-permissions`
- Call `ansys/actions/doc-deploy-dev@feat/check-permissions` from a test repo (without enough permissions of course).

inimaz:
> * Call `ansys/actions/doc-deploy-dev@feat/check-permissions` from a test repo (without enough permissions of course).

Nice! I didn't know this was possible, thanks on it!

_check-github-permissions/action.yml
```diff
@@ -0,0 +1,131 @@
...
+
+        IFS=',' read -ra PERMS <<< "${PERMISSIONS}"
+        for perm in "${PERMS[@]}"; do
+          perm=$(echo "${perm}" | xargs)
+          scope=$(echo "${perm}" | cut -d: -f1 | xargs)
+
+          case "${scope}" in
+            contents)
+              probe "repos/${REPO}/contents/" || \
+                echo "::warning::Missing permission '${perm}'. Add 'contents: read' (or higher) to your workflow's permissions block."
```
moe-ad, line 95: I was wondering how easy it would be to print an exact warning message i.e. instead of "contents: read (or higher)", "contents: write" is printed (which we could detect from `inputs.permissions`).

_check-github-permissions/action.yml
```diff
@@ -0,0 +1,131 @@
...
+              fi
+              ;;
+            attestations)
+              if [ -z "${ACTIONS_ID_TOKEN_REQUEST_URL:-}" ]; then
+                echo "::warning::Missing permission '${perm}'. Add 'attestations: write' and 'id-token: write' to your workflow's permissions block."
+              fi
+              ;;
+            *)
+              echo "::notice::Unknown permission scope '${scope}' — skipping verification."
+              ;;
```
moe-ad, line 129: If any of the checks fail, we can fail the action early by emitting exit code 1, since the action won't work correctly anyways if the permissions aren't set properly.

_check-github-permissions/action.yml
inimaz: Indeed! The issue is that we cannot distinguish for now `read` and `write` permissions. Since normally `read` is a prerequisite for `write` I only test `read` things in the `probe()` function, like this I do not write anything. For future versions we can try to find write actions that are harmless per each permission?

_check-github-permissions/action.yml
inimaz: Sure! 
Up to us to decide. Maybe there are some actions that work with a subset of the permissions?
For example I was using the `doc-deploy-dev` without the `pull-requests:write` permission and it was working fine on its own, until I added the `doc-deploy-pr` that creates comments on prs --> then `doc-deploy-dev` needs the permission to comment that on a pr to close previously-opened comments .

**TLDR**:  some jobs still work without the permissions because they do just a subset of all the actions, should we fail them and tell them to add the proper permission?

_check-github-permissions/action.yml
moe-ad: Maybe we should change the wording of the warning message then. For instance, we can add a link pointing to the documentation of the action (which already contains information about the minimum permissions required).

_check-github-permissions/action.yml
moe-ad: Based on your explanation in the other comment, until we are able to test write permissions, we cannot fail all workflows.

It can however be done for workflows that require only read permissions, so maybe we do that first and re-iterate in the future like you suggested.

On the `doc-deploy-dev` case, your observation is correct. We will have to treat it as a special case. For example, we can split [this step](https://github.com/ansys/actions/blob/198e33d7f55c5a283b6610e8333ae06f39a5da91/doc-deploy-dev/action.yml#L398) into multiple steps: the first checks if there is something to be removed and sets a flag, the next then checks for write permissions, and the last does the actually removal. That way, we are able to check for write permissions only when needed. You see what I mean?
But of course, that still depends on being able to test write permissions in the first place.

AlejandroFernandezLuces (approved):

_check-github-permissions/action.yml
```diff
@@ -0,0 +1,131 @@
...
+        probe() {
+          local endpoint="$1"
+          local err
+          if ! err=$(gh api "${endpoint}" 2>&1 >/dev/null); then
+            echo "${err}" | grep -qiE "HTTP 403|403|forbidden|not accessible by integration|insufficient permission" && return 1
+          fi
+          return 0
+        }
+
+        IFS=',' read -ra PERMS <<< "${PERMISSIONS}"
```
klmcadams, line 87: To condense the repeating echo statements, could you do something like this? I understand if the case structure is more readable though

```
# Define the endpoints and suggested permissions as associative arrays
declare -A endpoints=(
  [contents]="repos/${REPO}/contents/"
  [pull-requests]="repos/${REPO}/pulls?per_page=1"
  [issues]="repos/${REPO}/issues?per_page=1"
  [actions]="repos/${REPO}/actions/runs?per_page=1"
  [packages]="repos/${REPO}/packages?package_type=container"
  [pages]="repos/${REPO}/pages"
)

declare -A suggestions=(
  [contents]="contents: read"
  [pull-requests]="pull-requests: read"
  [issues]="issues: read"
  [actions]="actions: read"
  [packages]="packages: read"
  [pages]="pages: read"
)

IFS=',' read -ra PERMS <<< "${PERMISSIONS}"
for perm in "${PERMS[@]}"; do
  perm=$(echo "${perm}" | xargs)
  scope=$(echo "${perm}" | cut -d: -f1 | xargs)

  if [[ -n "${endpoints[$scope]}" ]]; then
    probe "${endpoints[$scope]}" || \
      echo "::warning::Missing permission '${perm}'. Add '${suggestions[$scope]}' (or higher) to your workflow's permissions block."
  elif [[ "$scope" == "id-token" ]]; then
    [ -z "${ACTIONS_ID_TOKEN_REQUEST_URL:-}" ] && \
      echo "::warning::Missing permission '${perm}'. Add 'id-token: write' to your workflow's permissions block."
  elif [[ "$scope" == "attestations" ]]; then
    [ -z "${ACTIONS_ID_TOKEN_REQUEST_URL:-}" ] && \
      echo "::warning::Missing permission '${perm}'. Add 'attestations: write' and 'id-token: write' to your workflow's permissions block."
  else
    echo "::notice::Unknown permission scope '${scope}' — skipping verification."
  fi
done
```

inigoimaz: Thanks for this @klmcadams , I missed the ping because I changed git user. Implemented your improvement here --> [3365ac2](https://github.com/ansys/actions/pull/1410/commits/3365ac2d3b05d3f222d78d4d97efd07e5292445d)

_check-github-permissions/action.yml
inimaz: Good idea thanks @moe-ad. Done here --> [66771a4](https://github.com/ansys/actions/pull/1410/commits/66771a403a796ce7ba19b473b15b113dd6443f5c)
if no `doc-url` it will give the url by default of github and permissions on actions.

jorgepiloto (commented):
Thanks for adding this, @inigoimaz. Is it something we can expand to other actions?

inigoimaz:
> Thanks for adding this, @inigoimaz. Is it something we can expand to other actions?

We can indeed @jorgepiloto ! Shall I do it in a separate PR or in this one? Both approaches are fine by me.

jorgepiloto:
> > Thanks for adding this, @inigoimaz. Is it something we can expand to other actions?
> 
> We can indeed @jorgepiloto ! Shall I do it in a separate PR or in this one? Both approaches are fine by me.

Let's do it in this same pull-request if that works for you, @inigoimaz

inigoimaz:
> Let's do it in this same pull-request if that works for you, @inigoimaz

Sounds good @jorgepiloto ! Done here --> [c1cae88](https://github.com/ansys/actions/pull/1410/commits/c1cae88168cb268edcf40c2608b65475555b7159). 

I did not add it to the ones deprecated (e.g. `release-pypi-*`). I set the branch to `feat/check-permissions` for now just to see if the ci passes ([b1386c8](https://github.com/ansys/actions/pull/1410/commits/b1386c8600d2fa80c2e2e6c693dc8a0544ad31b5)).
 Once approved this PR I can change the references back to `main` before merging.

jorgepiloto (approved):
Thanks for adding this. Approved.

_check-github-permissions/action.yml
inigoimaz: Yep thanks! We can upgrade this later on to distinguish read from write if needed.

MaxJPRey (approved):

_check-github-permissions/action.yml
```diff
@@ -0,0 +1,127 @@
...
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        PERMISSIONS: ${{ inputs.permissions }}
+        REPO: ${{ github.repository }}
+        DOC_URL: ${{ inputs.doc-url }}
+      run: |
+        probe() {
+          local endpoint="$1"
+          local err
+          if ! err=$(gh api "${endpoint}" 2>&1 >/dev/null); then
```
MaxJPRey, line 92: Just to be sure, are we checking at any point in our workflows that `gh` exists on the runner, for instance self hosted runners.

inigoimaz: Good catch! There is no check on missing `gh`. I added it in a separate PR (see #1577)

RobPasMue: Should we actually check this? GitHub runners come by default with it - it's one of their requirements. The same way we have to set up self-hosted runners whenever we create them, this might be a requirement for us next time...

I don't see the need on adding this check to be honest. We are starting to implement many checks which maybe shouldn't be on our actions (i.e. dependency-wise). We don't make sure that ``g++`` or the MS Build tools are installed on the ``build-package`` or ``build-wheelhouse`` action.. and if a library has to compile its sources (e.g. C-based extensions) the actions will also fail.

I know in this case we are actively using the ``gh`` CLI tool - but it should be installed by default on the nominal scenarios. Self-hosted runners should set up their workspace properly.

inigoimaz: Both options fine by me. If users do not have gh on the machine and we don't do the check on`gh`, in the worst-case scenario they will get a warning on the job. So not a big deal.

RobPasMue: I'm open to both - but I have my own preference. I just wanted to question whether that's the route we want to take or not

inigoimaz: if we take the route of checking tooling, we have as you mentioned several tools to check... Plus python, plus nodejs in some cases... Maybe not worth the effort?

clatapie (approved):
LGTM!

PR #1422, opened 2026-07-22
docs: disclaimer on security checks

RobPasMue:
As title says - follow up from situations seen in the ecosystem


MaxJPRey (approved):
Thanks for the quick update.

PR #1425, opened 2026-07-23
chore: update the handling of uv dependabot updates

moe-ad:
The changes made in #1401 did not work as expected, examples:
  - https://github.com/ansys/actions/pull/1419
  - https://github.com/ansys/actions/pull/1413
  - https://github.com/ansys/actions/pull/1420

So in this PR, I am:
  - [x] Reverting the changes made in #1401.
  - [x] Updating all the undesirable `pyproject.toml` changes that have been merged.
  - [x] Adding a job to `ci_cd_dependabot.yml` to add a note to the end of `uv` dependabot PR description so that we are reminded to verify any `pyproject.toml` changes in subsequent PRs before merging. The note will look [like this](https://github.com/ansys-internal/test-repo-moe/pull/57/#issue-4957265765). Already discussed this with @SMoraisAnsys as a fallback approach in the past.

We can also decide to stop worrying about restricting the bounds in the `pyproject.toml` altogether. We can then observe by how much the CI load would actually increase and then decide if we are fine with that.

### DIFF
--- a/.github/workflows/ci_cd_dependabot.yml
+++ b/.github/workflows/ci_cd_dependabot.yml
@@ -93,3 +93,58 @@ jobs:
             git commit -am "fix: update requirements files [skip ci]"
           fi
           git push
+
+  append-pr-note:
+    name: Append pyproject.toml note to dependabot PR
+    permissions:
+      contents: read
+      pull-requests: write # Needed to edit the dependabot PR body
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Append warning note to the dependabot PR body"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          BRANCH: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+
+          # Resolve the PR number associated with the pushed dependabot branch.
+          PR_NUMBER="$(gh pr list --head "${BRANCH}" --state open --json number -q '.[0].number' || true)"
+          if [ -z "${PR_NUMBER}" ]; then
+            echo "::warning::No open PR found for branch '${BRANCH}' yet; skipping PR body update."
+            exit 0
+          fi
+          echo "Found PR #${PR_NUMBER} for branch '${BRANCH}'."
+
+          # Hidden marker to prevent duplicate appends.
+          MARKER="<!-- ansys-actions:pyproject-note -->"
+
+          # Fetch the current PR body.
+          CURRENT_BODY="$(gh pr view "${PR_NUMBER}" --json body -q .body)"
+
+          if printf '%s' "${CURRENT_BODY}" | grep -qF "${MARKER}"; then
+            echo "Marker already present in PR #${PR_NUMBER} body; nothing to do."
+            exit 0
+          fi
+
+          # Build the new body: existing body + blank line + marked note block.
+          NEW_BODY_FILE="$(mktemp)"
+          {
+            printf '%s\n\n' "${CURRENT_BODY}"
+            printf '%s\n' "${MARKER}"
+            cat <<'EOF'
+          > [!WARNING]
+          > If this dependabot update includes any changes to the `pyproject.toml`, ensure the version specifiers in the `pyproject.toml` are reverted to a valid `major.minor` range by taking the following steps:
+          >   - Clone the branch locally.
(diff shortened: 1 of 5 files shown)


RobPasMue:
@moe-ad - watch out with #1419 ...

I had to change the behavior of the scripts to align with ghapi 2.X (which is not a patch update - it's a major bump!)

RobPasMue:
Looking at your changes it doesn't seem to be reverted so I am fine with it

klmcadams (approved):

.github/workflows/ci_cd_dependabot.yml
```diff
@@ -93,3 +93,58 @@ jobs:
...
+          # Build the new body: existing body + blank line + marked note block.
+          NEW_BODY_FILE="$(mktemp)"
+          {
+            printf '%s\n\n' "${CURRENT_BODY}"
+            printf '%s\n' "${MARKER}"
+            cat <<'EOF'
+          > [!WARNING]
+          > If this dependabot update includes any changes to the `pyproject.toml`, ensure the version specifiers in the `pyproject.toml` are reverted to a valid `major.minor` range by taking the following steps:
+          >   - Clone the branch locally.
+          >   - Update the pinned versions in the `pyproject.toml` as needed. For example, say dependabot proposes a change from "wcwidth<=0.7" to "wcwidth<=0.8.1", this must be updated to "wcwidth<=0.9". Another example is a change from "safety>=3.7,<4" to "safety>=3.8.0,<4", this must be updated to "safety>=3.8,<4". No changes should be made to the `requirement.txt` files.
```
Revathyvenugopal162:
```suggestion
          >   - Update any pinned versions modified by Dependabot in `pyproject.toml` to the appropriate `major.minor` version range. For example, if Dependabot changes `wcwidth<=0.7` to `wcwidth<=0.8.1`, it should be updated to `wcwidth<=0.9`. Likewise, if Dependabot changes `safety>=3.7,<4` to `safety>=3.8.0,<4`, it should be updated to `safety>=3.8,<4`. Do **not** modify any `requirement.txt` files.
```

.github/workflows/ci_cd_dependabot.yml
```diff
@@ -93,3 +93,58 @@ jobs:
...
+          >   - Run `uv lock` to update the lock file.
+          >   - Commit your changes to the PR and proceed to merge.
```
Revathyvenugopal162:
```suggestion
          >   - Commit the changes to this PR and proceed to merge.
```

Revathyvenugopal162 (approved):

SMoraisAnsys (commented):
@moe-ad can you remind us what is the benefit of working with X.Y specificly ? When we discussed about that in the past, the goal was to reduce the number of PR opened by dependabot to lessen the load associated to CI bumps.
AFAIK this is working fine currently and allows us to update packages less often. The proposed changes are going to add an extra manual layer. Isn't that counter productive ?

SMoraisAnsys (changes requested):
Adding a request for changes review to ensure a response to my previous comment before any merge :)

moe-ad:
> @moe-ad can you remind us what is the benefit of working with X.Y specificly ? When we discussed about that in the past, the goal was to reduce the number of PR opened by dependabot to lessen the load associated to CI bumps. AFAIK this is working fine currently and allows us to update packages less often. The proposed changes are going to add an extra manual layer. Isn't that counter productive ?

Hi @SMoraisAnsys. If you read the PR description, I already made a comment about this:
> We can also decide to stop worrying about restricting the bounds in the pyproject.toml altogether. We can then observe by how much the CI load would actually increase and then decide if we are fine with that.

FYI that this is the [feedback](https://github.com/ansys/actions/pull/991#issuecomment-3600869508) that led to us taking this approach initially and I am personally fine with either approach (the manual step isn't much work and it doesn't happen often).

Though I must say that given that we do not yet know every possible way the ongoing effort to add more uv support might be impacted by the action dependencies, meaning we might require more iterations of the implementation in the future (overhaul how we are consuming dependencies for example), it might make sense to retain the flexibility pinning to `major.minor` in the `pyproject.toml` offers us.

SMoraisAnsys:
As long as we keep using pinned dependencies and have a valid and secure way to update the dependencies on a regular base, I don't have any strong opinion. If you think that this won't add too much on your plate then you can test things and give us your feedback later.

SMoraisAnsys (approved):

PR #1427, opened 2026-07-25
feat: use uv lock file

SMoraisAnsys:
Follow up of #1364 but with a clean branch.

Related to https://github.com/ansys/actions/pull/1088, in particular [this comment](https://github.com/ansys/actions/pull/1088#issuecomment-4716877380), and #1088.

The use of `uv add --group ansys-actions -r "${REQUIREMENTS_FILE}"` has been removed in 10ddec7a6d31adcaaef6b9f71e38b39680e5f174 because some hooks might check the pyproject.toml / uv.lock files and lead to CI failure. That is the case with `pyansys-geometry` which leverages `pyproject-fmt`. 

The smoke_test.py file has been updated in [6932c1a](https://github.com/ansys/actions/pull/1427/commits/6932c1a73bcf28020e2ba7ad4841b3179c08c993) to be compatible with `uv`'s architecture to manage packages. I assume that we we weren't affected previously because `uv pip install` mimics `pip`'s behavior. But when using `uv sync` that behavior changes and we can't use `CWD` to retrieve the dist files. Levering relative location solves this issue and seems to also work with other package managers.

>[!NOTE]
To align the lockfile logic to the artificial project `ansys-actions-ci-dependencies`, the `doc` dependency group has been added and the minimum version has been bumped. This aligns with our will to follow [SPEC 0](https://scientific-python.org/specs/spec-0000/) and ensure that we keep using the same dependencies has the one that where defined in `requirements/requirements_doc.txt` **which has been removed** ! Each action has it's requirements.txt file regenerated and `docutils` has been pinned back to `0.22.4` instead of `0.23` (automatic generation).

>[!WARNING]
On top of the breaking change associated to the use of dependencies, the project has been upgraded to python3.12+ compatibility. This means that from this point we might define (if not yet with this PR) dependencies that are not compatible for downstream workflows. I'm fine with it but we have to make it clear to users. Pinging @moe-ad for visibility.


.github/workflows/ci_cd_pr.yml
```diff
@@ -124,6 +124,7 @@ jobs:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           use-python-cache: false
           checkout: false
+          skip-install: false
```
SMoraisAnsys: Adding a comment to remind the need to revert some CI changes before any merge

.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -173,11 +173,12 @@ jobs:
           operating-system: ${{ matrix.os }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           attest-provenance: true
+          use-uv-lockfile: ${{ matrix.repository.name == 'pyansys-geometry' }}
```
moe-ad: I suggest creating a separate workflow (`ci_cd_pr_uv` ?) for testing projects that have started moving to `uv` to keep things clean and separated, and to have a natural place for testing more `uv` projects in the future (quite [a number of them](https://github.com/search?q=org%3Aansys+uv_build&type=code) now). What do you think?

Although `pyansys-geometry` is using a [`hatchling`](https://github.com/ansys/pyansys-geometry/blob/9aa2c263385424b20d17fc7484eab4491ac25e51/pyproject.toml#L2) build backend instead of `uv_build`, we could still test all projects leveraging uv lock files but not using any of `flit_core.buildapi`/`poetry.core.masonry.api` build backends in `ci_cd_pr_uv`.

build-library/action.yml
```diff
@@ -164,7 +176,46 @@ runs:
...
+
+    - name: "Install dependencies from uv lockfile"
+      if: ${{ inputs.use-uv == 'true' && inputs.use-uv-lockfile == 'true' }}
+      shell: bash
+      env:
+        WORKING_DIRECTORY: ${{ inputs.working-directory }}
+      run: |
+        cd ${WORKING_DIRECTORY}
+        uv export --frozen --no-emit-project --no-hashes -o /tmp/requirements-locked.txt
+        uv pip install --no-managed-python --system -r /tmp/requirements-locked.txt
```
moe-ad: Why not use `uv sync ...` like you did in the other actions? Is there a particular reason?

build-wheelhouse/action.yml
```diff
@@ -290,10 +302,33 @@ runs:
...
+      with:
+        level: "ERROR"
+        message: >
+          uv lockfile (uv.lock) not found in repository root
+
     - name: "Install the library with the desired targets"
       env:
         INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python' || 'python -m pip install' }}
         INSTALL_TARGET: ${{ steps.specific-target-requested.outputs.install_target }}
+        TARGET: ${{ inputs.target }}
```
moe-ad, line 310: `INSTALL_TARGET` can be used directly rather than creating a separate env variable.

build-wheelhouse/smoke_test.py
```diff
@@ -70,7 +70,7 @@ def find_module_from_dist(pkg_name: str, attr: str):
         if pth_file:
             source_code_folder = None
             # Read the .pth file to find the source code path
-            with Path(pth_file).open("r") as file:
+            with Path(str(dist.locate_file(pth_file))).open("r") as file:
```
moe-ad, line 73: Out of curiosity, why this change? Did the previous implementation start failing?

.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -173,11 +173,12 @@ jobs:
...
+          use-uv: ${{ matrix.repository.name == 'pyansys-geometry' }}
```
moe-ad:
```suggestion
```

`use-uv` is true by default in the actions, which means we were always testing this execution path for every project. This change now means it will be only be tested for `pyansys-geometry`, which is not what you intended I suppose.

.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -216,6 +216,8 @@ jobs:
           library-name: ${{ matrix.repository.library }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           attest-provenance: true
+          use-uv-lockfile: ${{ matrix.repository.name == 'pyansys-geometry' }}
+          use-uv: ${{ matrix.repository.name == 'pyansys-geometry' }}
```
moe-ad:
```suggestion
```
See the reason for removal from the initial comment on this.

.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -216,6 +216,8 @@ jobs:
           library-name: ${{ matrix.repository.library }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           attest-provenance: true
+          use-uv-lockfile: ${{ matrix.repository.name == 'pyansys-geometry' }}
```
moe-ad: See the first comment on this.

.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -291,6 +293,9 @@ jobs:
           use-python-cache: false
           use-prek: ${{ matrix.use-prek }}
           checkout: false
+          use-uv-lockfile: ${{ matrix.repository.name == 'pyansys-geometry' }}
+          use-uv: ${{ matrix.repository.name == 'pyansys-geometry' }}
```
moe-ad:
```suggestion
```

See the reason for removal from the initial comment on this.

.github/workflows/ci_cd_pr_flit.yml
```diff
@@ -291,6 +293,9 @@ jobs:
           use-python-cache: false
           use-prek: ${{ matrix.use-prek }}
           checkout: false
+          use-uv-lockfile: ${{ matrix.repository.name == 'pyansys-geometry' }}
```
moe-ad: See the first comment on this.

check-vulnerabilities/action.yml
```diff
@@ -376,29 +406,41 @@ runs:
...
+          fi
+          # 2. Trigger the installation command based on the build backend
+          if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
+            poetry install --extras "${extra_targets}"
+          elif [[ "${BUILD_BACKEND}" == 'uv' ]]; then
+            uv pip install ."${extra_targets}"
+          else
+            python -m pip install ."${extra_targets}"
+          fi
         fi
```
moe-ad: I am not sold on having `uv` related logic split across multiple sections TBH. I suggest something like:

**BEFORE**
```
${ACTIVATE_VENV}
if [[ "${EXTRA_TARGETS}" != '' ]]; then
  if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
    echo "Installing extra targets with poetry: ${EXTRA_TARGETS}"
    extra_targets=${EXTRA_TARGETS}
  else
    echo "Installing extra targets with pip: ${EXTRA_TARGETS}"
    extra_targets=[${EXTRA_TARGETS}]
  fi
else
  echo "No extra targets to install"
  extra_targets=""
fi

if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
  poetry install --extras "${extra_targets}"
elif [[ "${BUILD_BACKEND}" == 'uv' ]]; then
  uv pip install ."${extra_targets}"
else
  python -m pip install ."${extra_targets}"
fi
```

**AFTER**
```
${ACTIVATE_VENV}
if [[ "${EXTRA_TARGETS}" != '' ]]; then
  if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
    echo "Installing extra targets with poetry: ${EXTRA_TARGETS}"
    extra_targets=${EXTRA_TARGETS}
  else
    echo "Installing extra targets with pip: ${EXTRA_TARGETS}"
    extra_targets=[${EXTRA_TARGETS}]
  fi
else
  echo "No extra targets to install"
  extra_targets=""
fi

if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
  poetry install --extras "${extra_targets}"
elif [[ "${BUILD_BACKEND}" == 'uv' ]]; then
  if [[ ${USE_UV_LOCKFILE}" == 'true' ]]; then
    if [[ "${EXTRA_TARGETS}" != '' ]]; then
      uv sync --frozen --extra "${EXTRA_TARGETS}" --no-dev
    else
      uv sync --frozen --no-dev
    fi
  else
    uv pip install ."${extra_targets}"
  fi
else
  python -m pip install ."${extra_targets}"
fi
```

check-vulnerabilities/action.yml
```diff
@@ -410,6 +452,17 @@ runs:
...
+    - uses: ansys/actions/_logging@main
+      if: ${{ inputs.skip-install == 'true' && inputs.use-uv-lockfile == 'true' && steps.check-exists-requirements-file.outputs.EXISTS_REQUIREMENTS == 'true' }}
+      with:
+        level: "ERROR"
+        message: >
+          ``use-uv-lockfile`` is enabled, but ``skip-install`` is also enabled and
+          a requirements file exists. Please, either use the lock file or the requirements
+          file, but not both. If you want to use the lock file, remove the requirements file
+          from the repository. If you want to use the requirements file, disable
+          ``use-uv-lockfile``.
```
moe-ad: I think this step should come way earlier in the action rather than being here.

moe-ad (commented):
Thanks for the changes @SMoraisAnsys, I left a couple of comments.

There is one other thing I have been thinking about, why can't we just use the `uv` lock files automatically once detected rather than exposing an explicit input for it (similar to how there is no input for `poetry`). IMO, a project that already uses a lock file would expect its dependencies to be installed from it.

.github/workflows/ci_cd_pr_flit.yml
SMoraisAnsys: That was just a CI config to illustrate that the changes work. Having another test workflow is a good idea but shouldn't we wait for [uv_build](https://docs.astral.sh/uv/concepts/build-backend/#the-uv-build-backend) to be used ?

build-wheelhouse/smoke_test.py
SMoraisAnsys: This is described in the PR description. Let me know if that's not clear enoguh

.github/workflows/ci_cd_pr_flit.yml
SMoraisAnsys: This is a temporary ci config and planned to be reverted (cf https://github.com/ansys/actions/pull/1427#discussion_r3671958669)

SMoraisAnsys:
> Thanks for the changes @SMoraisAnsys, I left a couple of comments.
> 
> There is one other thing I have been thinking about, why can't we just use the `uv` lock files automatically once detected rather than exposing an explicit input for it (similar to how there is no input for `poetry`). IMO, a project that already uses a lock file would expect its dependencies to be install from it.

I'm fine with detecting it and activating it by default. My only concern is the sudden change of behavior for people that still work with version tag and/or don't read every release note carefully.

build-wheelhouse/action.yml
SMoraisAnsys: Unfortunately this is not possible because `INSTALL_TARGET` would be `.` or `.[my-target]` (I just tested the changes :') ) Also, this env var is used when creating the wheel so better keep things as they are.

.github/workflows/ci_cd_pr_flit.yml
moe-ad: My thought is that projects that have since switched to `hatchling` + `uv` do not fit under flit/poetry anymore, and we might as well group them with uv related projects since the important thing for us to test atm is how project management related `uv` commands interact with those projects.

build-wheelhouse/smoke_test.py
moe-ad: I missed that. Understood... 👍

moe-ad:
> > Thanks for the changes @SMoraisAnsys, I left a couple of comments.
> > There is one other thing I have been thinking about, why can't we just use the `uv` lock files automatically once detected rather than exposing an explicit input for it (similar to how there is no input for `poetry`). IMO, a project that already uses a lock file would expect its dependencies to be install from it.
> 
> I'm fine with detecting it and activating it by default. My only concern is the sudden change of behavior for people that still work with version tag and/or don't read every release note carefully.

I could wait for this to get merged before v11 release, that way we have a graceful way of introducing this change. We can clearly document behavior change in the migration guide and those using tags won't be affected until they explicitly bump. What do you think?

check-vulnerabilities/action.yml
```diff
@@ -392,24 +446,23 @@ runs:
           echo "No extra targets to install"
           extra_targets=""
         fi
-
+        # 2. Trigger the installation command based on the build backend
         if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
           poetry install --extras "${extra_targets}"
         elif [[ "${BUILD_BACKEND}" == 'uv' ]]; then
-          uv pip install ."${extra_targets}"
+          if [[ ${USE_UV_LOCKFILE}" == 'true' ]]; then
```
Revathyvenugopal162:
```suggestion
          if [[ "${USE_UV_LOCKFILE}" == 'true' ]]; then
```
typo, just encounter this issue in another repo :)

check-vulnerabilities/action.yml
```diff
@@ -251,6 +263,28 @@ runs:
...
+    - uses: ansys/actions/_logging@main
+      if: ${{ inputs.skip-install == 'true' && inputs.use-uv-lockfile == 'true' && steps.check-exists-requirements-file.outputs.EXISTS_REQUIREMENTS == 'true' }}
+      with:
+        level: "ERROR"
+        message: >
+          ``use-uv-lockfile`` is enabled, but ``skip-install`` is also enabled and
+          a requirements file exists. Please, either use the lock file or the requirements
+          file, but not both. If you want to use the lock file, remove the requirements file
+          from the repository. If you want to use the requirements file, disable
+          ``use-uv-lockfile``.
```
Revathyvenugopal162:
```suggestion
          ``use-uv-lockfile`` is enabled, but ``skip-install`` is also enabled and a
          requirements file was found. This configuration is not supported. Choose a
          single dependency source: either ``uv.lock`` or the requirements file.
          Disable ``use-uv-lockfile`` or remove the requirements file..
```

.github/workflows/ci_cd_pr_flit.yml
Revathyvenugopal162: i agree with @moe-ad  creating a separate workflow. Testing uv.lock functionality under the flit CI/CD workflow does not really fit and keeping it separate would make things cleaner.

But the name ci_cd_pr_uv makes the idea we are using uv_build. as it may suggest that we are testing projects using `uv_build`. Unlike ci_cd_pr_flit and ci_cd_pr_poetry, uv is not necessarily the build backend here. 

Maybe we could use a more specific name, such as `ci_cd_pr_uv_lock`, `ci_cd_pr_uv_lockfiles`, or `ci_cd_pr_uv_projects` to better reflect what is actually being tested for now?

That said, this is just a suggestion and I'm fine with either approach. If the team prefers ci_cd_pr_uv/ keep it here, I have no strong objection to it.

check-vulnerabilities/action.yml
MaxJPRey:
```suggestion
          if [[ "${USE_UV_LOCKFILE}" == 'true' ]]; then
```

doc/source/changelog/1427.added.md
```diff
@@ -0,0 +1 @@
+Use uv lock file
```
SMoraisAnsys, line 1: pytest action still requires to be updated

build-library/action.yml
```diff
@@ -164,7 +164,38 @@ runs:
         message: >
           Skipping Python setup to use system Python.
 
+    # ------------------------------------------------------------------------
+    # Install package by leveraging uv lockfile if requested and available
+    # ------------------------------------------------------------------------
```
moe-ad:
```suggestion
    # --------------------------------------------------------
    # Install package by leveraging uv lockfile if detected
    # --------------------------------------------------------
```

moe-ad (approved):
I think out of all the actions this has been extended to, `_doc-build-windows`, `check-vulnerabilities` and `tests-pytest` are the only ones not being tested directly in the actions pipelines. It will be nice if we can test those prior to merging, just to be certain things run fine.

Thanks for the major changes @SMoraisAnsys!

SMoraisAnsys:
> I think out of all the actions this has been extended to, `_doc-build-windows`, `check-vulnerabilities` and `tests-pytest` are the only ones not being tested directly in the actions pipelines. It will be nice if we can test those prior to merging, just to be certain things run fine.
> 
> Thanks for the major changes @SMoraisAnsys!

I was working on adding those tests in the other PR. ~If the changes aren't pushed, I'll update that (miss on my side)~ (it was pushed).

SMoraisAnsys:
Converted to draft until testing is fully performed (still some misses)

requirements/requirements_doc.txt
jorgepiloto, line 1: Thanks for removing this file. The `tox.ini` must be updated so that the `doc-html` and `doc-links` environment work properly.

I think this wasn't caught by the CI/CD because we're not running Tox there.

SMoraisAnsys: I'm thinking of adding a new dev dependency group with `tox` and [tox-uv](https://github.com/tox-dev/tox-uv) which seems to be aligning with the use of lockfile. I'm not familiar with the use of tox, do you think this should be the path or another approach would be wiser ?

jorgepiloto: There shouldn't be a dependency group for `tox`. The reason is that it would install `tox` inside `tox`.

Replacing `requirements` table by `dependency_groups` should be enough.

SMoraisAnsys: Not sure if I'm missing something or if I was not clear in my sentence. Can you have a look in [8d34315](https://github.com/ansys/actions/pull/1427/commits/8d343154261eadf153a694bba12f24afb74378fc) ? I did some changes to address your comment.

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
 
 [testenv]
 description = Base test environment
+runner = uv-venv-runner
```
moe-ad:
```suggestion
```
Not needed, this will be the default for all tox environments unless overidden.

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
...
 passenv =
     PIP_EXTRA_INDEX_URL
 
 [testenv:code-style]
+runner = uv-venv-lock-runner
 description = Run code style checks via prek
 skip_install = true
-deps =
-    prek
+package = skip
```
moe-ad:
```suggestion
```

`skip_install = true` already takes care of this.

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
...
+runner = uv-venv-lock-runner
 description =
     Environment for
     html: html documentation generation
     clean: cleaning previously generated html documentation
     links: verifying the integrity of external links within the documentation
 skip_install = true
-deps =
-    links,html: -r{toxinidir}/requirements/requirements_doc.txt
+package = skip
```
moe-ad:
```suggestion
```

tox.ini
```diff
@@ -53,9 +64,13 @@ commands =
     python {toxinidir}/python-utils/update_tag_references.py {posargs}
 
 [testenv:tests]
+runner = uv-venv-lock-runner
 description = Run Python utility tests
 skip_install = true
-deps =
-    pytest
+package = skip
```
moe-ad:
```suggestion
```

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
 
 [testenv]
 description = Base test environment
+runner = uv-venv-runner
 basepython = python3
 setenv =
     PYTHONUNBUFFERED = yes
```
moe-ad, line 16:
```suggestion
    PYTHONUNBUFFERED = yes
    UV_FROZEN = 1
```

Let's set this here and we can avoid having to specify `uv_sync_flags = --frozen` everywhere else.

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
...
+runner = uv-venv-lock-runner
 description = Run code style checks via prek
 skip_install = true
-deps =
-    prek
+package = skip
+only_groups =
+    code-style-prek
+uv_sync_flags =
+    --frozen
```
moe-ad:
```suggestion
```

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
...
     clean: cleaning previously generated html documentation
     links: verifying the integrity of external links within the documentation
 skip_install = true
-deps =
-    links,html: -r{toxinidir}/requirements/requirements_doc.txt
+package = skip
+only_groups =
+    links,html: doc
+uv_sync_flags =
+    --frozen
```
moe-ad:
```suggestion
```

tox.ini
```diff
@@ -53,9 +64,13 @@ commands =
...
+runner = uv-venv-lock-runner
 description = Run Python utility tests
 skip_install = true
-deps =
-    pytest
+package = skip
+only_groups =
+    tests-pytest
+uv_sync_flags =
+    --frozen
```
moe-ad:
```suggestion
```

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
...
 skip_install = true
-deps =
-    links,html: -r{toxinidir}/requirements/requirements_doc.txt
+package = skip
+only_groups =
+    links,html: doc
+uv_sync_flags =
+    --frozen
 setenv =
     SOURCE_DIR = doc/source
```
moe-ad, line 42:
```suggestion
setenv =
    {[testenv]setenv}
    SOURCE_DIR = doc/source
```

So that the `UV_FROZEN` gets carried over here.

tox.ini
```diff
@@ -9,38 +11,47 @@ isolated_build = false
...
     BUILD_DIR = doc/_build
     BUILDER_OPTS = --color -v -j auto -W --keep-going
     links: BUILDER = linkcheck
     html: BUILDER = html
 commands =
-    clean: python -c "import shutil, sys; shutil.rmtree(sys.argv[1], ignore_errors=True)" "{toxinidir}/doc/_build"
-    html,links: sphinx-build -d "{toxworkdir}/doc_doctree" {env:SOURCE_DIR} "{toxinidir}/{env:BUILD_DIR}/{env:BUILDER}" {env:BUILDER_OPTS} -b {env:BUILDER}
+    clean: uv run --no-sync python -c "import shutil, sys; shutil.rmtree(sys.argv[1], ignore_errors=True)" "{toxinidir}/doc/_build"
+    html,links: uv run --no-sync sphinx-build -d "{toxworkdir}/doc_doctree" {env:SOURCE_DIR} "{toxinidir}/{env:BUILD_DIR}/{env:BUILDER}" {env:BUILDER_OPTS} -b {env:BUILDER}
```
moe-ad:
```suggestion
    clean: python -c "import shutil, sys; shutil.rmtree(sys.argv[1], ignore_errors=True)" "{toxinidir}/doc/_build"
    html,links: sphinx-build -d "{toxworkdir}/doc_doctree" {env:SOURCE_DIR} "{toxinidir}/{env:BUILD_DIR}/{env:BUILDER}" {env:BUILDER_OPTS} -b {env:BUILDER}
```

Not needed, Tox will automatically run the command in the provisioned venv. We should let tox do its thing and not hand the commands to `uv run ...` again.

tox.ini
```diff
@@ -53,9 +64,13 @@ commands =
...
-deps =
-    pytest
+package = skip
+only_groups =
+    tests-pytest
+uv_sync_flags =
+    --frozen
 commands =
-    pytest -vv {toxinidir}/python-utils {posargs}
+    uv run --no-sync pytest -vv {toxinidir}/python-utils {posargs}
```
moe-ad:
```suggestion
    pytest -vv {toxinidir}/python-utils {posargs}
```

moe-ad (commented):
Thanks for the new changes @SMoraisAnsys, I left a couple of comments.

tox.ini
SMoraisAnsys: ~AFAIK tox’s built-in default runner is not uv so it is required here (and could be removed in other locations), right ?~
I think I understand your comment now. I'll apply you changes but it might have been worth to set the runner here, no ?

tests-pytest/action.yml
```diff
@@ -346,7 +371,19 @@ runs:
           # Optional dependencies
           optional_dependencies_flag=""
           if [ -n "${OPTIONAL_DEPENDENCIES_NAME}" ]; then
-            optional_dependencies_flag="[${OPTIONAL_DEPENDENCIES_NAME}]"
+            if [[ "${BUILD_BACKEND}" == 'uv' && "${USE_UV_LOCKFILE}" == 'true' ]]; then
+              if [[ -n "${OPTIONAL_DEPENDENCIES_NAME}" ]]; then
```
Revathyvenugopal162:
```suggestion
```
i think this is redundant, remove the inner check in favour of L#373

tests-pytest/action.yml
```diff
@@ -346,7 +371,19 @@ runs:
...
+                for extra in "${extras_list[@]}"; do
+                  trimmed_extra="$(echo "${extra}" | xargs)"
+                  if [[ -n "${trimmed_extra}" ]]; then
+                    optional_dependencies_flag="$optional_dependencies_flag --extra ${trimmed_extra}"
+                  fi
+                done
+              fi
+            else
+              optional_dependencies_flag="[${OPTIONAL_DEPENDENCIES_NAME}]"
+            fi
```
Revathyvenugopal162, line 384:
```suggestion
            if [[ "${BUILD_BACKEND}" == 'uv' && "${USE_UV_LOCKFILE}" == 'true' ]]; then
                IFS=',' read -ra extras_list <<< "${OPTIONAL_DEPENDENCIES_NAME}"
                for extra in "${extras_list[@]}"; do
                  trimmed_extra="$(echo "${extra}" | xargs)"
                  if [[ -n "${trimmed_extra}" ]]; then
                    optional_dependencies_flag="$optional_dependencies_flag --extra ${trimmed_extra}"
                  fi
                done
            else
              optional_dependencies_flag="[${OPTIONAL_DEPENDENCIES_NAME}]"
            fi
```

jorgepiloto (commented):
Thanks for the big work in here , @SMoraisAnsys. As soon as the concerns raised by @moe-ad in #1439 get solved, references in here can be updated and we can merge things.

build-library/action.yml
```diff
@@ -164,7 +164,38 @@ runs:
...
+      if: ${{ inputs.use-uv == 'true' }}
+      shell: bash
+      env:
+        WORKING_DIRECTORY: ${{ inputs.working-directory }}
+      run: |
+        if [[ -f "${WORKING_DIRECTORY}/uv.lock" ]]; then
+          echo "USE_UV_LOCKFILE=true" >> ${GITHUB_OUTPUT}
+        else
+          echo "USE_UV_LOCKFILE=false" >> ${GITHUB_OUTPUT}
+        fi
```
Revathyvenugopal162, line 182: I guess this can be done in a dedicated composite action, e.g. `_detect-uv-lockfile/action.yml` and reduce the duplicated code

tests-pytest/action.yml
Revathyvenugopal162: @SMoraisAnsys

doc-build/action.yml
```diff
@@ -289,7 +289,7 @@ runs:
 
     - name: Documentation build (Linux)
       if: ${{ runner.os == 'Linux' }}
-      uses: ansys/actions/_doc-build-linux@main
+      uses: ansys/actions/_doc-build-linux@feat/use-uv-lock-file
```
klmcadams: Don't forget to change this before merging into main or right after

klmcadams (approved):
Looks good to me! I agree with Revathy's comment about creating a composite action to reduce redundant code

build-library/action.yml
SMoraisAnsys: I'll leave that for a future refactoring because this would also be the place where we would add build backend detection, dependency groups usage, extra install option handling, ... But that's a very good idea and would help to reduce many actions complexity.

tests-pytest/action.yml
SMoraisAnsys: Very good catch, I can't add your whole proposition due to indent issue so I'll take the first one and handle things manually :)

build-library/action.yml
SMoraisAnsys: Resolving this comment, follow up discussion in https://github.com/ansys/actions/issues/1457

AlejandroFernandezLuces (approved):
LGTM!

Revathyvenugopal162 (approved):

ecoussoux-ansys (approved):

PR #1439, opened 2026-07-31
ci: add workflow to test uv lockfiles

SMoraisAnsys:
Extend the CI testing with a workflow focusing on actions that can leverage uv lockfiles.


moe-ad (commented):
My main concern here is making things consistent, especially for the future. I am proposing some changes via this single comment so that my thought process is presented more clearly in a single place:
  - Let's simply name the new workflow `ci_cd_pr_uv.yml`.
  - We should create another small `ansys-actions-uv` package in `.ci/`. We can anticipate future trends with this package by using `uv_build` as the backend, even though projects currently going all-in with uv couple it with `hatchling.build` (examples [1](https://github.com/ansys/pyacp/pull/988) and [2](https://github.com/ansys/pyansys-geometry/pull/2872)). Creating this package will help address the following:
    - The new workflow tests directly with `ansys-actions-flit` package by adding a lock file. But `use-uv` is true by default in all actions and there is autodetection of uv lock files, meaning there are now lots of duplicated test jobs between `ci_cd_pr_flit.yml` and this workflow.
    - A `test-release-uv` test job can be subsequently added to the new workflow, serving a distinct purpose.
    - We start preparing `ansys/actions` for when projects start switching to `uv_build` as the build backend.
  - `ci_cd_pr.yml` serves the purpose of collecting test jobs that won't appear in all `ci_cd_pr_*.yml` files. So let's update the `test-doc-build-pyansys-geometry` job that already exists in that workflow to avoid a duplicated test job and also move `test_doc_build-pyaedt` to that workflow.
  - All `pyansys-geometry` test jobs in `ci_cd_pr_flit.yml` needs to be removed in favor of the new workflow as the new location to avoid duplicated jobs.
  - All `pyacp` test jobs in `ci_cd_pr_poetry.yml` also needs to be moved to the new workflow because it now falls in the same category as `pyansys-geometry`.

.github/workflows/ci_cd_pr_uv.yml
```diff
@@ -0,0 +1,146 @@
...
+        uses: ./build-library
+        with:
+          library-name: ${{ env.LIBRARY_NAME }}
+          attest-provenance: true
+          python-version: ${{ env.MAIN_PYTHON_VERSION }}
+          working-directory: .ci/${{ env.LIBRARY_NAME }}
+          use-uv: true
+          checkout: false
+
+  test-check-vulnerabilities:
```
SMoraisAnsys, line 113: @moe-ad Just wanted to highlight this way of testing for our dummy repos. I think it might be worth to also apply that logic with other packages.

moe-ad (approved):

PR #1469, opened 2026-08-10
chore: bump default python version

### LINKED ISSUE #1379: Update default Python version to `3.12+`
With dropping  Python `3.11` and `3.10`, it would be beneficial to update the default Python version used in several actions (https://github.com/ansys/actions/blob/main/build-library/action.yml) to Python 3.12 and above.

Since the Python version is already configurable as an input parameter, this change is expected to have minimal impact on existing workflows. 
Updating the default will help keep the setup aligned with the currently supported Python versions while still allowing flexibility when needed.

moe-ad:
As the title states. Closes #1379.

### DIFF
--- a/build-library/action.yml
+++ b/build-library/action.yml
@@ -62,7 +62,7 @@ inputs:
   python-version:
     description: |
       Python version used for installing and execution of the build process.
-    default: '3.11'
+    default: '3.12'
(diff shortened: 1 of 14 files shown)


jorgepiloto (approved):
Good catch!

PR #1474, opened 2026-08-10
fix: dependabot(uv) update issues

### LINKED ISSUE #1473: Investigate what caused a change in the behavior of dependabot update scripts for `uv`
In #1461, a lot of the updates made to the requirement files just made no sense, which we reverted in #1468.
The same pattern can be seen in other currently opened PRs like #1462, #1463, and #1464. We should investigate and fix the root cause.

moe-ad:
Closes #1473.

The root cause was not adding `--no-dev` flag as I have done (uv automatically includes the dev group in every export unless explictly disabled), and the reason the issue is just revealing itself now (and not in the past) is because the dev dependency group was recently just added (in https://github.com/ansys/actions/pull/1427).

### DIFF
--- a/.github/workflows/ci_cd_dependabot.yml
+++ b/.github/workflows/ci_cd_dependabot.yml
@@ -72,7 +72,7 @@ jobs:
               echo "  Processing: $file_cleaned (group: $group_name)"
 
               # Run uv export command
-              uv export --format requirements.txt --group "$group_name" --output-file "$file_cleaned" --no-hashes
+              uv export --format requirements.txt --group "$group_name" --output-file "$file_cleaned" --no-hashes --no-dev
           done
 
       - name: Commit and push requirements files if changed
--- a/doc/source/changelog/1474.fixed.md
+++ b/doc/source/changelog/1474.fixed.md
@@ -1,0 +1,1 @@
+Dependabot(uv) update issues


moe-ad:
Tested locally btw. If the reviewer also wants to test locally:
  - Checkout main branch
  - Update `uv.lock` as dependabot proposed in https://github.com/ansys/actions/pull/1461.
  - Run the following shell script:
     ```bash
      echo "Directories containing requirements files:"
      echo "==========================================="
      
      # Process standard requirements.txt files
      find . -maxdepth 2 -type f -name "requirements.txt" -exec dirname {} \; | sort -u | while read dir; do
          # Remove leading ./
          dir_cleaned="${dir#./}"
          # Get the directory name without leading underscore for the group name
          group_name="${dir_cleaned#_}"
      
          echo "  Processing: $dir_cleaned/requirements.txt (group: $group_name)"
      
          # Run uv export command
          uv export --format requirements.txt --group "$group_name" --output-file "$dir_cleaned/requirements.txt" --no-hashes --no-dev
      done
      
      # Process requirements-<suffix>.txt files (e.g., code-style/requirements-prek.txt)
      find . -maxdepth 2 -type f -name "requirements-*.txt" | sort | while read file; do
          # Remove leading ./
          file_cleaned="${file#./}"
          dir_cleaned="$(dirname "$file_cleaned")"
          # Extract suffix from filename (e.g., "prek" from "requirements-prek.txt")
          basename_file="$(basename "$file_cleaned")"
          suffix="${basename_file#requirements-}"
          suffix="${suffix%.txt}"
          # Group name is dir (without underscore) + "-" + suffix
          group_name="${dir_cleaned#_}-${suffix}"
      
          echo "  Processing: $file_cleaned (group: $group_name)"
      
          # Run uv export command
          uv export --format requirements.txt --group "$group_name" --output-file "$file_cleaned" --no-hashes
      done
      ```
    
    Notice the changes to the requirement files are like in #1461.
  - Discard the changes to the requirements files, add `--no-dev` to the script as I have done here, and run again.
  - Things should be fine and dandy afterwards.

ecoussoux-ansys (approved):
Good catch !
Thanks for providing the detailed instructions to test locally - I performed the test as you described and everything looks good now 👍

RobPasMue (approved):
Well that explains it!

PR #1476, opened 2026-08-11
fix: update handling of `optional-dependencies-name` and `group-dependencies-name` in tests-pytest action

### LINKED ISSUE #1475: Issues with v11 and tests-pytest action
The following has to be adapted so that the behavior is the same as the doc-build action:

https://github.com/ansys/actions/blob/5d71ba8e7baec55197d6531b3d666809b4224695/tests-pytest/action.yml#L101-L121

Currently we are defaulting the value of ``optional-dependencies-name`` to ``tests``. If a project does NOT define an optional target for tests (which is becoming more common) this will fail. We need to adapt this in line with what was performed at the ``doc-build`` action.

moe-ad:
Closes #1475.

This PR aligns how **ansys/actions/tests-pytest** handles `optional-dependencies-name` and `group-dependencies-name` inputs with how they are handled in **ansys/actions/doc-build** (see #1350).

I have also added two tests for tests-pytest (for PyAEDT and PyAnsys Geometry).

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -343,7 +343,7 @@ jobs:
   test-doc-build-pyaedt:
     name: "Test doc-build action for PyAEDT on ${{ matrix.os.name }}"
     runs-on: ${{ matrix.os.name }}
-    needs: doc-build
+    needs: doc-style
     permissions:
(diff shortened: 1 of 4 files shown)


RobPasMue (approved):
LGTM!

PR #1485, opened 2026-08-13
fix: ignore .python-version file when detected

moe-ad:
@RobPasMue reported an issue regarding how actions interact with uv-based projects that have a .python-version file defined.

The problem is that the presence of `.python-version` file can result in uv trying to create a new virtual environment when executing a `uv sync ...` command, which results in failure since all other relevant actions tooling will not be present in this new virtual environment. This happens when the Python version specified in the file is different from the version provisioned by the action itself. See this [workflow run](https://github.com/ansys-internal/ansys-filetransfer-service/actions/runs/31669357193?pr=57) as one example, specifically what happened [in this step](https://github.com/ansys-internal/ansys-filetransfer-service/actions/runs/31669357193/job/94362161959?pr=57#step:2:447).

The simplest and straight forward implementation is to simply ignore the .python-version file as I have done in this PR. Note that I already explored alternatives like:
   - Passing the `--no-config` flag to the `uv sync` command, but that has the extra effect of ignoring other files like the `pyproject.toml` too, so not suitable.
   - Pointing the `UV_PYTHON` environment variable to the path of the Python executable an action provisions. My expectation is that this will take precedence over whatever is in `.python-version` but that doesn't seem to be the case.


moe-ad:
Fix tested in the same workflow linked in the PR description: [passing workflow](https://github.com/ansys-internal/ansys-filetransfer-service/actions/runs/31697670627?pr=57).

RobPasMue (approved):
From my perspective it's a valid approach.. but I wish the other options you explored had worked! 😄

PR #1497, opened 2026-08-19
feat: allow to skip latex install

vgelbgras:
This pull request introduces a new input option to the documentation build GitHub Actions, allowing users to skip the installation of LaTeX during the build process.


_doc-build-linux/action.yml
```diff
@@ -155,6 +155,13 @@ inputs:
     required: true
     type: string
 
+  skip-latex-install:
+    description: |
+      Whether to skip the installation of LaTeX. Default value is ``false``.
+    required: false
```
RobPasMue:
```suggestion
    required: true
```
On the private actions, we always require the inputs

_doc-build-windows/action.yml
```diff
@@ -147,6 +147,13 @@ inputs:
     required: true
     type: string
 
+  skip-latex-install:
+    description: |
+      Whether to skip the installation of LaTeX. Default value is ``false``.
+    required: false
```
RobPasMue:
```suggestion
    required: true
```

RobPasMue (approved):
I'm fine with this change - given the suggestions are merfed! =)

Issue #1342, opened 2026-05-28
Replace safety with an alternative solution


moe-ad:
As the title states. See #641 for more information.

PR #1498, opened 2026-08-24
feat: add uv audit in check vulnerability

### LINKED ISSUE #1458: extend check-vulnerarabilities with uv audit (experimental)
[Here](https://astral.sh/blog/uv-audit) is a blog post on `uv audit` and the associated roadmap can be found [here](https://github.com/astral-sh/uv/issues/18506). 

Given the status of `uv audit` (experimental) this might be something to opt-in.

>[!NOTE]
When I tried it, it was able to provide feedback on vulnerability that safety wasn't providing.

Extra: we should share uv sync operations can perform lightweight OSV-based lookup for previously-resolved malware. This can now be configured with `pyproject.toml` settings since [this PR](https://github.com/astral-sh/uv/pull/20587) and is available in `uv`'s versions above `v0.11.31`

SMoraisAnsys:
This PR allows users additionally run `uv audit` in `ansys/actions/check-vulnerabilities`.

While `uv audit` is still experimental, it can surface both known vulnerabilities and project risk signals (like adverse project status/malware-related checks). The current implementation also assumes that one has a `uv` lockfile available to avoid resolving the dependencies before checking them. If not, the action will fail. The current parsing is simple because the feature is still experimental.

If required (e.g. false positive), vulnerabilities can be ignored. For example, I get an expected failure when testing this branch in [here](https://github.com/ansys/pyaedt/actions/runs/32746581741/job/97874030688) and using
```
[tool.uv.audit]
ignore = ["PYSEC-2026-2132"]
```
allows the `uv audit` command to pass locally without failing.

Close #1458

---

>[!NOTE]
Even if they `safety` and `uv audit` do not use the same DB (PyUp Safety vs OSV), we might think about replacing the former with the latter to answer #1342, see also #641 for more information on this `safety` replacement topic. However, **projects without a lockfile would lose coverage entirely** so this might be for laterrrrrr. Maybe we could rediscuss that when the audit feature is no longer considered experimental ? :)


AlejandroFernandezLuces (commented):
This is a great addition, thanks for taking care of it! I have a question about this. What do you think of integrating the `uv audit` as a pre-commit hook rather than in the vulnerability action? Or maybe even having it in both places?

From my pov, it looks doable and potentially keeps the vulnerability from reaching public code, since users should have pre-commit installed in local generally.

check-vulnerabilities/action.yml
MaxJPRey, line 1: Do you think we could redirect uv audit output with `tee` rather than `> file 2>&1` so the log also appears in the Actions job output. Right now a failure summary is only visible after downloading the artifact.
You might have done it on purpose.

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -115,6 +118,44 @@ def check_vulnerabilities():
...
+        uv_audit_known_vulnerabilities = 0
+        uv_audit_adverse_project_statuses = 0
+
+        uv_audit_log_path = Path("info_uv_audit.log")
+        if not uv_audit_log_path.exists():
+            raise FileNotFoundError("Uv audit was enabled but 'info_uv_audit.log' is missing.")
+        else:
+            uv_audit_output = uv_audit_log_path.read_text(encoding="utf-8", errors="replace")
+
+            vulnerabilities_marker = "Vulnerabilities:"
```
MaxJPRey: I might be wrong here, but the `"Vulnerabilities:" ` marker parsing uses str.find on the raw log, which will also match that literal string if it ever appears earlier -for instance in inside a package description or a future header change- . 
We could anchor it to a line start with
` (re.search(r"(?m)^Vulnerabilities:\s*$", ...))`
to make it robust to uv output tweaks.

If you go with this, please double check the regex mentioned above 😄

check-vulnerabilities/action.yml
SMoraisAnsys: It was done on purpose, it can be activated with the correct input like the other outputs. If you use `hide-log=false` (`true` by default) then you can see it in the logs.

check-vulnerabilities/check_vulnerabilities.py
SMoraisAnsys: Good point, I've modified the code in a99b4295

check-vulnerabilities/action.yml
```diff
@@ -635,5 +710,7 @@ runs:
       if: inputs.upload-reports == 'true' || ( failure() && inputs.upload-reports == 'true' )
       with:
         name: vulnerability-results
-        path: ./info_*.json
+        path: |
+          ./info_*.json
+          ./info_uv_audit.log
```
ecoussoux-ansys: That's really a minor comment, but in order to upload the uv audit log file in case `upload-reports` is set to True, should you not first check if it exists, since the uv audit is optional and not enabled by default ?

Also maybe the name of that step could be updated to something like "Uploading safety and bandit results, as well as uv audit results if enabled".

ecoussoux-ansys (approved):
LGTM, just left a very minor comment.

check-vulnerabilities/action.yml
SMoraisAnsys: Step's name changed in 2b51684, thanks for pointing that out.
As for the file name being here or not, it's no a problem as if a file is not found the action only triggers a warning, see https://github.com/[actions/upload-artifact](https://github.com/actions/upload-artifact#inputs)#inputs. To avoid unwanted user warnings, I'll try to improve that in another commit. Thanks @ecoussoux-ansys

ecoussoux-ansys (commented):
Thanks for considering my previous comment @SMoraisAnsys !

That might not be relevant but I thought of something else: In the PR description you write that running uv audit assumes that a uv lock file is available.
I am not sure if the uv lock file is strictly needed to run the `uv audit`, but even if it's just the "recommended way", why do you prefer to assume this, rather than explicitly checking for its existence? In that case, if it appears to be missing a warning or an error could be raised and the "run uv audit checks" step could be skipped.

check-vulnerabilities/action.yml
```diff
@@ -613,6 +622,71 @@ runs:
...
         bandit ${CONFIGFILE} -r "${SOURCE_DIRECTORY}" -o info_bandit.json -f json --exit-zero
 
+    - uses: ansys/actions/_logging@main
+      if: ${{ inputs.use-uv-audit == 'true' }}
+      with:
+        level: "INFO"
+        message: >
+          Running uv audit security checks.
+
+    - name: "Detect uv audit malware-check config"
```
SMoraisAnsys, line 639: Add comment to highlight @ecoussoux-ansys 's comment on lock file detection.

check-vulnerabilities/action.yml
```diff
@@ -613,6 +629,71 @@ runs:
...
+      id: detect-uv-audit-malware-check
+      if: ${{ inputs.use-uv-audit == 'true' }}
+      shell: python
+      run: |
```
RobPasMue: We are trying to move the pure python logic to the python-utils folder if I am not mistaken. Could we move this there?

check-vulnerabilities/action.yml
```diff
@@ -613,6 +629,71 @@ runs:
...
+        with open(os.environ["GITHUB_OUTPUT"], "a", encoding="utf-8") as github_output:
+            github_output.write(
+                f"HAS_AUDIT_MALWARE_CHECK={'true' if has_audit_malware_check else 'false'}\n"
+            )
+
+    - name: "Suggest UV_MALWARE_CHECK fallback"
+      if: ${{ inputs.use-uv-audit == 'true' && steps.detect-uv-audit-malware-check.outputs.HAS_AUDIT_MALWARE_CHECK == 'false' }}
+      shell: bash
+      run: |
+        echo "::notice::Input use-uv-audit is enabled but malware-check is not set in pyproject.toml. A fallback UV_MALWARE_CHECK=1 will be used for uv audit. Consider adding malware-check to pyproject.toml to avoid this fallback."
```
RobPasMue: If we are selecting use-uv-audit (even if it was default) we should simply force the ``UV_MALWARE_CHECK=1`` option IMO and not simply rely on projects doing so

RobPasMue: This is a feature of the action we have ... so we simply force it. No need to announce it

check-vulnerabilities/action.yml
```diff
@@ -613,6 +629,71 @@ runs:
...
+        HAS_AUDIT_MALWARE_CHECK: ${{ steps.detect-uv-audit-malware-check.outputs.HAS_AUDIT_MALWARE_CHECK }}
+      run: |
+        ${ACTIVATE_VENV}
+
+        UV_AUDIT_LOG_PATH="info_uv_audit.log"
+        # If malware-check is not configured in pyproject.toml and no explicit env var
+        # was provided by the workflow, fallback to UV_MALWARE_CHECK=1 for this run.
+        if [[ "${HAS_AUDIT_MALWARE_CHECK}" != 'true' && -z "${UV_MALWARE_CHECK}" ]]; then
+          export UV_MALWARE_CHECK=1
+        fi
```
RobPasMue: Same, simply force the value

check-vulnerabilities/check_vulnerabilities.py
```diff
@@ -323,15 +361,25 @@ def check_vulnerabilities():
...
+
+    # Additional advisory checks for uv-audit
+    if UV_AUDIT_ENABLED:
+        print(f"Total 'uv audit' known vulnerabilities: {uv_audit_known_vulnerabilities}")
+        print(f"Total 'uv audit' adverse project statuses: {uv_audit_adverse_project_statuses}")
+        print("Note: 'uv audit' advisories may contain duplicates of 'safety' advisories.")
+        print("*****************************************************************************")
+        if uv_audit_vulnerability_details:
+            print("Uv audit vulnerabilities:\n")
+            print(uv_audit_vulnerability_details)
```
RobPasMue: I think we should integrate the reporting with the safety and bandit outputs... not as a separate stage of the report. Can you embed them in the proper location with the conditional usage?

moe-ad (approved):
@SMoraisAnsys shouldn't we document this for local runs, [similar to bandit and safety](https://actions.docs.ansys.com/version/stable/vulnerability-actions/index.html#check-vulnerabilities-action)?

I think we should. If you agree, we just need to update [`generate_advisory_files`](https://github.com/ansys/actions/blob/04a81b4d6c17739b20e2b77f550affa843f640d7/check-vulnerabilities/check_vulnerabilities.py#L388) with an initial run of `uv audit ...`, so that `info_uv_audit.log` will be generated. Otherwise, [this line](https://github.com/ansys/actions/blob/04a81b4d6c17739b20e2b77f550affa843f640d7/check-vulnerabilities/check_vulnerabilities.py#L128) would cause failure locally. And of course update the documentation section above.

Other than the above comments, the rest LGTM and I am approving.

check-vulnerabilities/action.yml
moe-ad: Doing this also means **detect-uv-audit-malware-check** step can be removed completely.

check-vulnerabilities/action.yml
```diff
@@ -630,10 +712,25 @@ runs:
           python "${GITHUB_ACTION_PATH}/check_vulnerabilities.py"
         fi
 
-    - name: "Uploading safety and bandit results"
+    - name: "Compute upload paths"
+      if: inputs.upload-reports == 'true' || ( failure() && inputs.upload-reports == 'true' )
```
MaxJPRey: The right side is a strict subset of the left side:
`failure() && inputs.upload-reports == 'true' `can never be true when `inputs.upload-reports == 'true'` alone is already `true`.
I think we can simplify by:
`if: ${{ always() && inputs.upload-reports == 'true' }}`

`always()` might be needed if you want the artifacts uploaded even after an early failure.

check-vulnerabilities/action.yml
```diff
@@ -630,10 +712,25 @@ runs:
           python "${GITHUB_ACTION_PATH}/check_vulnerabilities.py"
         fi
 
-    - name: "Uploading safety and bandit results"
+    - name: "Compute upload paths"
```
MaxJPRey: The current step always runs and builds a multi-line `UPLOAD_PATHS` even when uv-audit is disabled. We could drop the step and add  the conditional path directly in the artifact step.

check-vulnerabilities/action.yml
SMoraisAnsys: I'm fine with that, I used this approach cause I saw a recent PR with that kind of content and we also have it in multiple places. I'll handle that in a new commit, thanks for the remark :)

check-vulnerabilities/action.yml
SMoraisAnsys: Good point, I didn't think about allowing people to not activate it. It's more how the use of this check evolved and how it can now be configured :D I'll update that.

RobPasMue (approved):

doc/source/vulnerability-actions/index.rst
moe-ad, line 52: This section needs to be updated with `DEPENDENCY_CHECK_UV_AUDIT_ENABLED` right?

check-vulnerabilities/action.yml
```diff
@@ -613,6 +629,46 @@ runs:
...
+        level: "INFO"
+        message: >
+          Running uv audit security checks.
+
+    - name: "Detect uv audit malware-check config"
+      id: detect-uv-audit-malware-check
+      if: ${{ inputs.use-uv-audit == 'true' }}
+      shell: bash
+      run: |
+        python ${GITHUB_ACTION_PATH}/../python-utils/detect_uv_audit_malware_check.py
```
moe-ad, line 644: Maybe I am missing something, but my understanding is that both `UV_MALWARE_CHECK` env var and `audit.malware-check` in pyproject.toml are essentially just flags for turning on the malware scanning feature, right?

If so, then the `detect_uv_audit_malware_check.py` script shouldn't be require anymore since we want to force `UV_MALWARE_CHECK=1` in any case. You see what I mean?

python-utils/detect_uv_audit_malware_check.py
```diff
@@ -0,0 +1,72 @@
...
+    elif malware_check_value is False:
+        malware_check_output = "false"
+
+    with Path(os.environ["GITHUB_OUTPUT"]).open("a", encoding="utf-8") as github_output:
+        github_output.write(f"HAS_AUDIT_MALWARE_CHECK={'true' if has_malware_check else 'false'}\n")
+        github_output.write(f"AUDIT_MALWARE_CHECK_VALUE={malware_check_output}\n")
+
+
+if __name__ == "__main__":
+    main()
```
moe-ad, line 72: Based on my previous comment, the file shouldn't be needed.

doc/source/vulnerability-actions/index.rst
```diff
@@ -146,6 +147,16 @@ Documentation on how to address common vulnerabilities can be found in the
...
+.. note::
+
+   When the ``use-uv-audit`` option is enabled, ``uv audit`` runs **in addition** to
+   ``safety`` and ``bandit``. It checks the project's locked dependencies (via ``uv.lock``)
+   against the `OSV <https://osv.dev>`_ advisory database, and can also perform a
+   malware scan when ``malware-check`` is configured in ``[tool.uv.audit]`` (or via the
+   ``UV_MALWARE_CHECK=1`` environment variable as a fallback). If no such configuration is
+   found and the ``UV_MALWARE_CHECK`` environment variable is not set, the malware scan is performed
+   by default.
+
```
moe-ad, line 163: Based on my previous comment, this note's content will need to be updated (i.e. regarding `UV_MALWARE_CHECK` vs `audit.malware-check`)

moe-ad (commented):
I left a comment regarding `UV_MALWARE_CHECK` env variable vs `audit.malware-check` in pyproject.toml. Those two are essentially the same thing based on my understanding. So if we are already forcing `UV_MALWARE_CHECK=1`, then we shouldn't need to have any logic relating to detecting the value of `audit.malware-check`, right?

Unless I am misunderstanding something, in which case the PR should be good to go.

check-vulnerabilities/action.yml
SMoraisAnsys: You are totally right. I'm just keeping it around because if someone has already defined that value and it doesn't align with our will to force the check then, at least, they would be warned. From `uv`'s documentation I did read that env var behavior overrights the configuration in `pyproject.toml`.
I also think that teaching people about that potential parameter in the `pyproject.toml` can be beneficial if this audit ever reaches the precommit ecosystem :) Also, if the check is run locally, having the project set with the right config would avoid defining an environment variable.

I'm open to removing it if you don't think it brings enough value.

doc/source/vulnerability-actions/index.rst
SMoraisAnsys: Indeed, thanks for pointing that ! Addressed in [94f32d6](https://github.com/ansys/actions/pull/1498/commits/94f32d6af22878fdf17b39e6ccfc4c1e7d0b0ad2).

check-vulnerabilities/action.yml
moe-ad: There is indeed value in teaching people about it's existence.

My follow-up thought about this is then that if someone explicitly sets `malware-check = false`, I think we should at least respect it. Because we currently [don't](https://github.com/ansys/actions/blob/94f32d6af22878fdf17b39e6ccfc4c1e7d0b0ad2/check-vulnerabilities/action.yml#L658). That brings the additional value of giving folks a mechanism of influencing the audit process via the config, and that directly makes the existence of `detect_uv_audit_malware_check.py` even more justified. What do you think?

SMoraisAnsys: That was my original though and implementation but you guys convinced :p
Now that I'm convinced I would advocate that we keep using the malware check when users want to activate `uv audit` through our action. First, I'm not sure there is a good reason to deactivate it (I might be short sighted here). Second, if one really wants to have this behavior then one can simply use our default action behavior (without modifying the `use-uv-audit` input) and run a custom `uv audit` command.

PR #1500, opened 2026-08-25
fix: Make version check optional for private pypi publish

### LINKED ISSUE #1499: Tag verification for private pypi release should be optional

### 🐞 Description of the bug

In #1111 a verification check was added that the published version of a package was identical to the ref triggering the workflow. This makes sense for public releases but is breaking our long established workflow of pushing dev builds to the private pypi instance from main.

This needs to be optional for the private pypi index.

### 📝 Steps to reproduce

Rerun previously successful workflow and observe it now does not work

Linux

n/a (this template isnt very useful)

3.13

### 📦 Installed packages

```shell
ansys/actions@11.0.3
```

da1910:
Addresses #1499 

Makes the verification step optional for private pypi publish, the inner action has a required `verify-tag-version` input. The outer action has an optional input with the same name which defaults ot true.

### DIFF
--- a/_release-pypi/action.yml
+++ b/_release-pypi/action.yml
@@ -81,6 +81,13 @@ inputs:
     required: true
     type: boolean
 
+  verify-tag-version:
+    description: |
+      Verify that the git ref that triggered the workflow is the same as the version
+      of the package being uploaded.
+    required: true
+    type: boolean
+
   use-trusted-publisher:
     description: |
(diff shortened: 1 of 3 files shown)


RobPasMue (approved):
LGTM!

jorgepiloto (approved):
Thanks for implementing the issue, @da1910. All good on my side.

PR #1501, opened 2026-08-25
docs: document multiple-file Vale input for doc-style

### LINKED ISSUE #730: Document how to pass multiple files to Vale
This is a continuation of https://github.com/ansys/actions/issues/432, in particular of  this comment:

```console
      - name: "Install jq"
        run: |
          # Used to format Vale input as a JSON-formatted list of files
          sudo apt install -y jq

      - name: "Collect desired files"
        run: |
          # Find all .rst files excluding those in doc/source/api/
          RST_FILES=$(find doc/source -type f -name "*.rst" ! -path "doc/source/api/*")

          # Find all .py files inside the examples/ directory, excluding those with an underscore
          PY_FILES=$(find examples -type f -name "*.py" ! -name "*_*")

          # Combine both file lists and convert them to a JSON array
          # TODO: include Python files too. See Vale issue
          # https://github.com/errata-ai/vale/issues/858
          # VALE_FILES=$(echo -e "$RST_FILES\n$PY_FILES" | jq -R . | jq -s .)
          VALE_FILES=$(echo -e "$RST_FILES" | jq -R . | jq -s .)
          echo "VALE_FILES=$(echo $VALE_FILES)" >> $GITHUB_ENV
```

jorgepiloto:
This updates the `doc-style` action docs to show how to lint an explicit list of files with Vale.

- Clarifies that the `files` input accepts either a directory path or a JSON-formatted array of file paths.
- Adds a `doc-style-multiple-files.yml` example that builds a JSON file list with `jq` and passes it through `env.VALE_FILES`.

Closes #730.

### DIFF
--- a/.github/workflows/ci_cd_pr.yml
+++ b/.github/workflows/ci_cd_pr.yml
@@ -166,9 +166,26 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: "Collect files for Vale as JSON"
+        id: collect-vale-files
+        shell: bash
+        run: |
+          python - <<'PY' >> "${GITHUB_OUTPUT}"
+          import json
+          from pathlib import Path
+
+          files = sorted(
+              str(path)
+              for path in Path("doc/source").rglob("*.rst")
+              if not path.is_relative_to("doc/source/api")
+          )
+          print(f"vale_files={json.dumps(files)}")
+          PY
+
       - uses: ./doc-style
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          files: ${{ steps.collect-vale-files.outputs.vale_files }}
           checkout: false
 
   doc-build:
--- a/doc/source/style-actions/examples/doc-style-multiple-files.yml
+++ b/doc/source/style-actions/examples/doc-style-multiple-files.yml
@@ -5,19 +5,23 @@ doc-style:
     contents: read
     pull-requests: write
   steps:
+    - name: "Install Git and clone project"
+      uses: actions/checkout@v6
+
     - name: "Install jq"
       run: |
         # Used to format Vale input as a JSON-formatted list of files
         sudo apt install -y jq
 
     - name: "Collect files for Vale"
+      id: collect-files
       run: |
         RST_FILES=$(find doc/source -type f -name "*.rst" ! -path "doc/source/api/*")
         VALE_FILES=$(echo -e "$RST_FILES" | jq -R . | jq -s .)
-        echo "VALE_FILES=$(echo "$VALE_FILES")" >> "$GITHUB_ENV"
+        echo "VALE_FILES=$(echo "$VALE_FILES")" >> "${GITHUB_OUTPUT}"
 
     - name: "Running documentation style checks"
       uses: ansys/actions/doc-style@{{ version }}
       with:
         token: ${{ '{{ secrets.GITHUB_TOKEN }}' }}
-        files: ${{ '{{ env.VALE_FILES }}' }}
+        files: ${{ '{{ steps.collect-files.outputs.VALE_FILES }}' }}
--- a/doc/source/style-actions/examples/doc-style-multiple-files.yml
+++ b/doc/source/style-actions/examples/doc-style-multiple-files.yml
@@ -14,11 +14,20 @@ doc-style:
         sudo apt install -y jq
 
     - name: "Collect files for Vale"
-      id: collect-files
-      run: |
-        RST_FILES=$(find doc/source -type f -name "*.rst" ! -path "doc/source/api/*")
-        VALE_FILES=$(echo -e "$RST_FILES" | jq -R . | jq -s .)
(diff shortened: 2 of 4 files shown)


RobPasMue (approved):
Nice example!

SMoraisAnsys (approved):

doc/source/style-actions/examples/doc-style-multiple-files.yml
```diff
@@ -0,0 +1,23 @@
...
+  permissions:
+    contents: read
+    pull-requests: write
+  steps:
+    - name: "Install jq"
+      run: |
+        # Used to format Vale input as a JSON-formatted list of files
+        sudo apt install -y jq
+
+    - name: "Collect files for Vale"
```
moe-ad, line 16: There should be a checkout step before this, otherwise this step would fail.

doc/source/style-actions/examples/doc-style-multiple-files.yml
```diff
@@ -0,0 +1,23 @@
...
+      run: |
+        RST_FILES=$(find doc/source -type f -name "*.rst" ! -path "doc/source/api/*")
+        VALE_FILES=$(echo -e "$RST_FILES" | jq -R . | jq -s .)
+        echo "VALE_FILES=$(echo "$VALE_FILES")" >> "$GITHUB_ENV"
```
moe-ad: The example should reflect current best practice. Write to GITHUB_OUTPUT instead and update the next step correspondingly.

moe-ad (commented):
Thanks for this @jorgepiloto. I left a few comments.

Also can we test this just to make sure it works? I know we have faced issues in the past concerning passing json-formatted data to nested action calls. We can simply extend the [existing ci test](https://github.com/ansys/actions/blob/0a1dfcd67e28c5434e835d320bed184fc6913bcf/.github/workflows/ci_cd_pr.yml#L159) as a minimum.

moe-ad (changes requested):
I am requesting changes just to make sure we don't mistakenly merge given the existing approvals.

moe-ad (approved):

doc/source/style-actions/examples/doc-style-multiple-files.yml
```diff
@@ -0,0 +1,27 @@
...
+      run: |
+        # Used to format Vale input as a JSON-formatted list of files
+        sudo apt install -y jq
+
+    - name: "Collect files for Vale"
+      id: collect-files
+      run: |
+        RST_FILES=$(find doc/source -type f -name "*.rst" ! -path "doc/source/api/*")
+        VALE_FILES=$(echo -e "$RST_FILES" | jq -R . | jq -s .)
+        echo "VALE_FILES=$(echo "$VALE_FILES")" >> "${GITHUB_OUTPUT}"
```
moe-ad:
```suggestion
        id: collect-files
        shell: bash
        run: |
          python - <<'PY' >> "${GITHUB_OUTPUT}"
          import json
          from pathlib import Path

          files = sorted(
              str(path)
              for path in Path("doc/source").rglob("*.rst")
              if not path.is_relative_to("doc/source/api")
          )
          print(f"VALE_FILES={json.dumps(files)}")
          PY
```

Thanks for the changes. I guess it makes sense to change the example to match what we've tested. Just a recommendation, I have approved regardless.

.github/workflows/ci_cd_pr.yml
```diff
@@ -166,9 +166,26 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: "Collect files for Vale as JSON"
+        id: collect-vale-files
+        shell: bash
```
SMoraisAnsys, line 171: For consistency, could you refactor this step to use `shell: python` instead ? That's how we are handling that in other actions.

doc/source/style-actions/examples/doc-style-multiple-files.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: actions/checkout@v6
+
+    - name: "Install jq"
+      run: |
+        # Used to format Vale input as a JSON-formatted list of files
+        sudo apt install -y jq
+
+    - name: "Collect files for Vale"
+        id: collect-files
+        shell: bash
```
SMoraisAnsys: Same comment even if consistency might not be a requirement for an example

.github/workflows/ci_cd_pr.yml
moe-ad: Doable, yes. But that approach will make writing to GITHUB_OUTPUT not as straightforward as just doing it through bash. So I would advice leaving it as is.

SMoraisAnsys: Feel free to discard my comment then but I would emphasize to improve consistency in other actions then.

Note that, AFAIK, moving to `shell: python` would only require to replace `print(f"vale_files={json.dumps(files)}")` with
```python
    with open(os.environ["GITHUB_OUTPUT"], "a") as f:
        f.write(f"vale_files={json.dumps(files)}\n")
```
But again, as long as we are consitent, I'm fine with it.

clatapie (approved):

Issue #1457, opened 2026-08-06
refactor: add internal action to handle detection of build backend, uv lockfile, dependency groups usage, ...


SMoraisAnsys:
Mainly a follow up to the discussion [here](https://github.com/ansys/actions/pull/1427#discussion_r3721496533).

Having everything in a single job would help a lot with maintenance of our actions. This might add a tiny overhead to existing actions because some extra check might be performed but the maintenance would be WAYYYYY easier in the long run. Also, playing with inputs might allow to reduce overhead greatly :)

PR #1503, opened 2026-08-26
fix: build-library for projects not using uv lockfiles

moe-ad:
The changes we made in #1487 has the unforeseen side-effect that `uv run ...` only works with virtual environments.

So for projects not using uv lockfiles but still leveraging `use-uv`, `uv run ...` will create a fresh venv, leading to an error since the library will not be available in this new venv. See [this example](https://github.com/ansys-internal/ansys-turbo-components/actions/runs/32879390493/job/97910350539).

Thanks to @Acohen-work for reporting this.

### DIFF
--- a/build-library/action.yml
+++ b/build-library/action.yml
@@ -222,9 +222,9 @@ runs:
       env:
         TAG_REF: ${{ github.ref_name }}
         LIBRARY_NAME: ${{ inputs.library-name }}
-        USE_UV: ${{ inputs.use-uv }}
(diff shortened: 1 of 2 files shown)


SMoraisAnsys (approved):
LGTM

Wondering if we shouldn't just use venv everywhere to avoid this kind of side effects. Might be an addition to https://github.com/ansys/actions/issues/1457 scope

ecoussoux-ansys (approved):
LGTM !

PR #1512, opened 2026-08-26
docs: update vulnerability documentation on safety local scanning

moe-ad:
I recently had to do local scanning in `pydpf-core` and discovered that `safety` dropped `requests` for `httpx`. The implication is that the documentation on local scanning with safety needs to be updated wrt the environment variables that need to be set, hence this PR.


RobPasMue (approved):
Nice catches!

PR #1516, opened 2026-08-28
ci: update dependabot configuration

SMoraisAnsys:
Tiny update of the dependabot configuration because `pypa/gh-action-pypi-publish` is no longer an action that we leverage in any of our actions (see https://github.com/ansys/actions/pull/758 for more information). Now, it is only used in our CI.
Other minor changes are performed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -71,6 +71,7 @@ updates:
       cicd-actions:
         patterns:
           - "micnncim/action-label-syncer"
+          - "pypa/gh-action-pypi-publish"
       # Github actions that are used in our composite actions
       github-actions:
         patterns:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -104,15 +106,13 @@ updates:
...
-          - "pypa/gh-action-pypi-publish"
           - "softprops/action-gh-release"
       # Missed non ansys actions that should by assigned to a group
       must-be-assigned-actions:
         patterns:
-          - peter-evans/find-comment
-          - pyvista/setup-headless-display-action
-          - vimtor/action-zip
-          - dependabot/fetch-metadata
+          - "peter-evans/find-comment"
```
MaxJPRey, line 113: Out of curiosity, was it working without the quotes?

MaxJPRey (approved):

.github/dependabot.yml
SMoraisAnsys: Yes it did, see https://github.com/ansys/actions/pull/1291 :)

RobPasMue (approved):

PR #1518, opened 2026-08-31
chore: remove `skip-existing` input for release-pypi-private

### LINKED ISSUE #1515: Deprecate `skip-existing` input in `release_pypi_*` actions or keep?
Behind the scenes, the `skip-existing` input depends on twine supporting the `--skip-existing` flag, however, twine dropped support since `v6.2.0`. We have bumped twine version in Ansys Actions over the months (now pinned to v7.0.0) and we haven't deprecated this input. This is causing failures such as [this](https://github.com/ansys-internal/ansys-turbo-components/actions/runs/32970293647/job/98535413016) in the pipelines of projects.

I see two approach:
  - We simply deprecate this option. If libraries really need this option, they can pin to `v10`.
  - Since only `release-pypi-private` is effectively the only functioning PyPI release action, do we pin twine to `v6.1.0` forever? Because I think a lot of repos use this option, and it is the reason why `skip-existing`  is true by default. But I get that this option might not make sense.

Thoughts @jorgepiloto, @RobPasMue, @SMoraisAnsys?

moe-ad:
Closes #1515.

Given that the public release-* actions have been deprecated and the private PyPI did not support the option to begin with, I think it makes sense to just silently remove the option. That would resolve the ongoing issues and we don't have to worry about deprecation book keeping.

### DIFF
--- a/_release-pypi/action.yml
+++ b/_release-pypi/action.yml
@@ -74,13 +74,6 @@ inputs:
     required: true
     type: boolean
 
-  skip-existing:
-    description: |
-      Continue uploading files if one already exists. Only valid when uploading
-      to PyPI. Other implementations may not support this.
-    required: true
-    type: boolean
-
   verify-tag-version:
     description: |
       Verify that the git ref that triggered the workflow is the same as the version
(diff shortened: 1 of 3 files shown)


jorgepiloto (approved):

SMoraisAnsys (commented):
Is there a reason why only the private one receives this behavior ? Given @moe-ad 's statement, I think that it would make sense to also align the other release actions. All the more given that this PR has been merged.

PR #1527, opened 2026-09-01
refactor: reduce release note body size and add cmd

SMoraisAnsys:
Refactor the release note body that can be autogenerated to reduce it's size (we have many repos where the number of elements is large. An example is illustrated below with `pyansys-geometry's` latest release note

image

Also, the content of the release note contains extra information to let users know that they can directly download attestations from the projets repo `$REPO_URL/attestations`. See https://github.com/ansys/actions/pull/1527#issuecomment-5512022089 for the result on a dummy repo.


release-github/action.yml
```diff
@@ -487,13 +488,12 @@ runs:
...
-            filename=$(basename "$artifact")
-            echo "\$ gh attestation verify --owner ${GITHUB_REPOSITORY_OWNER} ${filename}"
-          done
-
+          echo "gh attestation verify <file-path of downloaded artifact> --repo ${GITHUB_REPOSITORY}"
+          echo "\`\`\`"
+          echo "You can also download the attestation from [GitHub](${REPO_URL}/attestations) and verify against that directly:"
+          echo ""
+          echo "\`\`\`bash"
+          echo "gh attestation verify <file-path of downloaded artifact> --bundle <file-path of downloaded attestation>"
```
RobPasMue: As I stated in my previous comment - let's add only 1 example as part of the body, so that users know how to verify it instead of figuring out the substitution themselves

RobPasMue:
```suggestion
          echo "gh attestation verify <file-path-of-downloaded-artifact> --repo ${GITHUB_REPOSITORY}"
          echo "\`\`\`"
          echo "You can also download the attestation from [GitHub](${REPO_URL}/attestations) and verify against that directly:"
          echo ""
          echo "\`\`\`bash"
          echo "gh attestation verify <file-path-of-downloaded-artifact> --bundle <file-path-of-downloaded-attestation>"
```

RobPasMue (approved):
LGTM! thanks for this enhancement @SMoraisAnsys !

ecoussoux-ansys (approved):

release-github/action.yml
```diff
@@ -34,6 +34,10 @@ description: |
     setup guide, https://actions.docs.ansys.com/version/dev/migrations/docs-changelog-setup.html,
     the changelog for your release is included in the release notes on GitHub.
 
+    If you are using the attestation features, we recommend you add the ``attestations: read``
+    permission to your GitHub token. Doing so will allow this action to add an example of
+    attested artifacts to the release notes.
+
     .. important:: **Required GitHub Permissions**
 
        - **contents**: ``write`` - Required to create GitHub releases and upload release assets
```
moe-ad, line 39:
```suggestion
       - **contents**: ``write`` - Required to create GitHub releases and upload release assets
       - **attestations**: ``read`` (optional) - Adding this permission is required when attestation features are being used i.e. when ``add-artifact-attestation-notes`` input is ``true``.
```

release-github/action.yml
```diff
@@ -34,6 +34,10 @@ description: |
     setup guide, https://actions.docs.ansys.com/version/dev/migrations/docs-changelog-setup.html,
     the changelog for your release is included in the release notes on GitHub.
 
+    If you are using the attestation features, we recommend you add the ``attestations: read``
+    permission to your GitHub token. Doing so will allow this action to add an example of
+    attested artifacts to the release notes.
+
```
moe-ad:
```suggestion
```

I recommend moving this to the permissions section.

moe-ad (approved):
LGTM!

I left a minor comment, just for consistency regarding where permission related information should be documented.

release-github/action.yml
SMoraisAnsys: I've not added it because we section title was `Required GitHub Permissions`. I can add another 'informational' section if you want. Would that work for you @moe-ad

release-github/action.yml
SMoraisAnsys: Waiting for an answer to the discussion above

release-github/action.yml
moe-ad: I guess the using "... (optional): ..." as proposed in my suggestion should be enough rather than creating a separate section?
Both approach works fine in any case from my perspective.

SMoraisAnsys: IMO we can mix them if we remove 'Required' from the section title. Otherwise, splitting makes more sense because optional and required don't mix well semantically.

PR #1531, opened 2026-09-02
feat: LFS support during checkout

### LINKED ISSUE #1437: Add optional git-lfs support to default checkout
### Description

It would be helpful to add an optional parameter that enables Git LFS support during the default checkout process for repositories with large files requiring LFS.

Example usage:

```yaml
- uses: ansys/actions/<action>
  with:
    git-lfs: true
```

Default value:

```yaml
git-lfs: false
```

### Expected Behavior

When `git-lfs` is enabled, the action should perform the equivalent of a `git lfs pull` (or otherwise ensure that all LFS-managed files are downloaded and available) after the repository checkout.

### Motivation

Many repositories rely on Git LFS for large assets. Currently, users need to add the checkout step manually with that option and provide `checkout: false` option. Providing built-in support through an optional parameter would simplify workflows while preserving the current behavior by default.

_No response_

_No response_

moe-ad:
Closes #1437.

This PR exposes a `checkout-lfs` input for some actions so that projects leveraging git LFS can use those actions without adding preliminary steps.

I handpicked actions where it makes sense to support this input for the following reasons:
- `doc-build`: documentation assets can be tracked through LFS.
- `tests-pytest`: large binary inputs for tests are natural LFS candidates.
- `build-library` and `build-wheelhouse`: `.dll` and `.so` files are distributed in some libraries, and those are also good candidates for LFS.

@ansys/pyansys-core, let me know if it makes sense for any other action to support this input.

### DIFF
--- a/build-library/action.yml
+++ b/build-library/action.yml
@@ -84,6 +84,16 @@ inputs:
     default: true
     type: boolean
 
+  checkout-lfs:
+    description: |
+      Whether to download Git LFS-tracked files during the checkout step.
+      Enable this when the repository stores package data files tracked with
+      Git LFS that must be included in the built sdist/wheel. Only applies
+      when ``checkout`` is ``true``. Default value is ``false``.
+    required: false
+    default: false
+    type: boolean
+
   validate-build:
     description: |
       Whether to validate the build process. If ``true``, the build process is
@@ -147,6 +157,7 @@ runs:
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
         persist-credentials: false
+        lfs: ${{ inputs.checkout-lfs }}
 
     - name: "Set up Python ${{ inputs.python-version }}"
       if: ${{ inputs.skip-python-setup == 'false' }}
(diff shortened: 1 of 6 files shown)


MaxJPRey (commented):
Agreed with the actions targeted.

Could `check-licenses` and `check-vulnerabilities` be additional candidates, as they need the actual file contents, including during project installation.

`check-licenses` installs the project before checking dependency licenses.
`check-vulnerabilities` installs the local library by default.

RobPasMue:
> `check-licenses` installs the project before checking dependency licenses.

For checking the licenses we shouldn't need any large files AFAIK.. only the pyproject.toml/setup.py files and lockfiles. I wouldn't expose it there unless explicitly requested by someone

> `check-vulnerabilities` installs the local library by default.

Same here as before - the vulnerability checks don't require checking out large files AFAIK.

RobPasMue (approved):

MaxJPRey:
> > `check-licenses` installs the project before checking dependency licenses.
> 
> For checking the licenses we shouldn't need any large files AFAIK.. only the pyproject.toml/setup.py files and lockfiles. I wouldn't expose it there unless explicitly requested by someone
> 
> > `check-vulnerabilities` installs the local library by default.
> 
> Same here as before - the vulnerability checks don't require checking out large files AFAIK.

Valid point. Agreed.

SMoraisAnsys (approved):

PR #1532, opened 2026-09-02
fix: reorder `doc-deploy-stable` action steps to allow uv caching

### LINKED ISSUE #1073: Ensure actions/checkout is performed before astral-sh/setup-uv
I was looking into a failed stable doc deployment and noticed that the action `doc-deploy-stable` is installing `uv` before performing a checkout of the repo. This shows the following warning in the logs
Image
and I think this highlights that if a user wants to reuse `uv` cache, it will not work as expected because `setup-uv` does not have access to the information necessary for its cache management.

A simple fix should be to reorder the steps of the action. Note that this might also concern other actions and anyone handling this issue should have a look in the whole repo if possible :)

moe-ad:
Closes #1073.

A light PR, I have just reordered action steps. I have checked all other actions and only `doc-deploy-stable` is affected.

### DIFF
--- a/doc-deploy-stable/action.yml
+++ b/doc-deploy-stable/action.yml
@@ -191,33 +191,6 @@ runs:
       with:
         level: "INFO"
         message: >
-          Collect the version number from the tag or the branch name according
-          to the reference type. Tags names are expected to follow
-          'v<MAJOR>.<MINOR>' or 'v<MAJOR>.<MINOR>.<PATCH>'. Branches must follow
-          'release/<MAJOR>.<MINOR>'.
-
-    - name: "Set up Python ${{ inputs.python-version }}"
-      uses: ansys/actions/_setup-python@main
-      with:
-        python-version: ${{ inputs.python-version }}
-        use-cache: ${{ inputs.use-python-cache }}
-        provision-uv: ${{ inputs.use-uv }}
-        prune-uv-cache: ${{ inputs.use-python-cache != 'true' }}
-
-    - name: "Install packaging"
-      shell: bash
-      env:
-        INSTALL_COMMAND: ${{ inputs.use-uv == 'true' && 'uv pip install --no-managed-python --system' || 'python -m pip install' }}
-      run: |
-        ${INSTALL_COMMAND} --upgrade pip
-        ${INSTALL_COMMAND} -r ${GITHUB_ACTION_PATH}/requirements.txt
-
-    # ------------------------------------------------------------------------
-
-    - uses: ansys/actions/_logging@main
-      with:
-        level: "INFO"
-        message: >
           Checkout the repository branch for deploying the documentation. If this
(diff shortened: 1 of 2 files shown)


doc-deploy-stable/action.yml
```diff
@@ -239,6 +212,33 @@ runs:
...
+      with:
+        level: "INFO"
+        message: >
+          Collect the version number from the tag or the branch name according
+          to the reference type. Tags names are expected to follow
+          'v<MAJOR>.<MINOR>' or 'v<MAJOR>.<MINOR>.<PATCH>'. Branches must follow
+          'release/<MAJOR>.<MINOR>'.
+
+    - name: "Set up Python ${{ inputs.python-version }}"
+      uses: ansys/actions/_setup-python@main
```
jorgepiloto:
```suggestion
      uses: $/_setup-python
```

jorgepiloto (changes requested):
I requested changes for one blocking compatibility issue. In doc-deploy-stable/action.yml, the new references to ansys/actions/_logging@main and ansys/actions/_setup-python@main decouple internal behavior from the caller pinned ansys/actions release. This means workflows pinned to a stable tag can still pull updates from main, which breaks reproducibility and can introduce unreviewed behavior changes. Please keep these internal action references tied to the same release as this action instead of main.

doc-deploy-stable/action.yml
```diff
@@ -239,6 +212,33 @@ runs:
         repository: ${{ steps.get-repository-name.outputs.REPOSITORY }}
         token: ${{ inputs.token }} # zizmor: ignore[artipacked] , credential must be persisted in this case
 
+    # ------------------------------------------------------------------------
+
+    - uses: ansys/actions/_logging@main
```
jorgepiloto: Thanks for fixing the _setup-python reference. I still see one compatibility issue here: _logging now points to ansys/actions/_logging@main. That still decouples this action from a caller pinned ansys/actions release tag and can pull behavior from main. Please keep this internal reference tied to the same release as the parent action.

jorgepiloto (changes requested):
I re-reviewed after your update. Thanks for fixing the _setup-python reference. I still have one blocking compatibility concern: doc-deploy-stable/action.yml now references ansys/actions/_logging@main, which can drift independently from a caller pinned ansys/actions release tag. Please bind this internal reference to the same release as the parent action instead of main.

doc-deploy-stable/action.yml
jorgepiloto:
```suggestion
    - uses: $/_logging
```

jorgepiloto (approved):
Thanks for fixing this, @moe-ad.

jorgepiloto (approved):

SMoraisAnsys (approved):

RobPasMue (approved):

PR #1533, opened 2026-09-02
fix: tag check for releases

clatapie:
This error was hit in PySam SysML2 when trying to make a release (https://github.com/ansys/pysam-sysml2/actions/runs/33525442329/job/100165544141) with the following error:

```
  Git tag version: 0.3.2
  Found artifact: dist/ansys-sam-sysml2-artifacts/ansys_sam_sysml2-0.3.2-py3-none-any.whl
  Artifact version: 2
  Version mismatch!
    Git tag version:     0.3.2
    Artifact version:    2
```


moe-ad (approved):
Thanks for the fix @clatapie. We didn't envisage a situation where the project name contains a number.

Tested the new filters locally and they work fine. I guess you need a patch release immediately this gets merged?

PR #1535, opened 2026-09-02
refactor: update check-licenses to exclude non-runtime dependencies during checks

### LINKED ISSUE #1066: License checker is not doing the right thing

### 🐞 Description of the bug

License checker should be checking the packages installed on a customer's machine, not pinned development only dependencies, unless the license terms of those packages specifically prohibit commercial use etc.

In the case of PyMaterials Manager the action is complaining about several dependencies being LGPL and friends, for example astroid which is a transitive dependency of sphinx-automodapi. This is used only to generate the documentation, and is not installed on a customer's environment.

Simply checking the lock file, or the environment when poetry install has been used is not the way to determine license compliance, realistically for locked environments we ought to install only the project, excluding any extras or other groups before running the reporting tool.

### 📝 Steps to reproduce

https://github.com/ansys/pymaterials-manager/actions/runs/19696138640/job/56421521678

Linux

_No response_

3.12

### 📦 Installed packages

```shell
n/a
```

moe-ad:
Closes #1066.

>[!NOTE]
>Whoever reviews this PR should also check #1563.

#### Poetry-related changes
So, when the issue was originally opened, `pymaterials-manager` was still using `poetry`, however, the project has since migrated to `uv`, which really complicated my life as there was no way to easily reproduce anymore... :)

Nonetheless, I looked into it and the main issue stems from the fact that poetry automatically installs all non-optional groups (check [this section](https://python-poetry.org/docs/managing-dependencies/#optional-groups) to find out more about "optional groups" in poetry lingo) when executing `poetry install` (see [this section](https://python-poetry.org/docs/managing-dependencies/#installing-group-dependencies)). To install only runtime dependencies, we need to pass `--only main` as I have done.

#### UV-related changes
`uv` also behaves similarly, in that it automatically syncs with the `dev` group included. Fine, we already pass `--no-dev` but I have updated that to `--no-default-groups` to make things more robust. This is because `uv` actually reads the [default-groups setting](https://docs.astral.sh/uv/reference/settings/#default-groups) to determine what groups to automatically include during a sync (and `dev` just happens to be the default value). So if a project reconfigures this setting to include other groups, `--no-dev` will no longer be sufficient.

#### Others
Given that an external virtual environment can be passed to the `check-licenses` action via the `activate-venv` input (which we have no control over), I have added a warning log to inform users about their role in ensuring the provided virtual environment contains only the packages whose licenses need to be checked.


check-licenses/action.yml
```diff
@@ -159,6 +159,19 @@ runs:
...
+      uses: $/_logging
+      with:
+        level: "WARNING"
+        message: >
+          The license check inspects every package installed in the provided
+          virtual environment. Make sure the environment activated via
+          activate-venv only contains consumer-facing dependencies (project
+          runtime and selected extras). Dependency groups (dev, doc, test, ...)
+          and build tooling (build, wheel, pyproject-hooks, ...) installed in
+          this environment also be reported, which may not be desired.
```
SMoraisAnsys:
```suggestion
          this environment will also be reported, which may not be desired.
```

SMoraisAnsys (approved):
Thanks for the changes @moe-ad ! LGTM, please let others also review the PR and do not merge it too fast (see [this comment](https://github.com/ansys/actions/pull/1562#issuecomment-5677847583))

jorgepiloto (approved):
Thanks for adding this, @moe-ad

RobPasMue (approved):

clatapie (approved):

PR #1548, opened 2026-09-07
ci: rework subaction usage to leverage local path

SMoraisAnsys:
Seems like we can leverage [local path](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#runsstepsuses) by using `uses: $/sub_action_name` (maybe even `uses: ./sub_action_name`).

This change would greatly help with the migration topic.

>[!NOTE]
The documentation on this approach seems to be quite recent (July 30th), see [this PR](https://github.com/github/docs/pull/45347/changes).

---

Edit: extra modification have been performed following https://github.com/ansys/actions/pull/1548#pullrequestreview-5138887312 If you look at the content of our PR workflows, it's now a mix of `uses: ./my-action` and `uses: $/my-action`. The reason for that is that we keep using `uses: ./my-action` in cases where we mock the behavior of an external repo using our action. For example, first we checkout `pymapdl`, then we checkout the repo in another directory, then we use `uses: ./actions/my-action` to explicitly force the execution of the action from the second checkout.


.github/workflows/ci_cd_release.yml
```diff
@@ -17,98 +17,8 @@ permissions: {}
 
 jobs:
 
-  rolling-release:
```
RobPasMue, line 20: Why was this reverted? This will still need to be true. We need the rolling release mechanism to generate the three types of tags we accept

RobPasMue (changes requested):

.github/workflows/ci_cd_release.yml
SMoraisAnsys: Totally true, I miss read part of the workflow and though it was in charge of something else. Thanks for catching that

jorgepiloto (commented):
Awesome! Thanks a lot, @SMoraisAnsys

clatapie (approved):
This will be very useful!

moe-ad (changes requested):
Two immediate feedbacks:
-  We need to update things similarly in the `ci_cd_*.yml` workflows. I mean steps such as [this](https://github.com/ansys/actions/blob/bcf2b662500dfcd31dcd9ae189ae7f3fc665de34/.github/workflows/ci_cd_pr.yml#L27-L36).
- Taking into consideration the changes in this PR, we can simplify the [`update_tag_references.py`](https://github.com/ansys/actions/blob/bcf2b662500dfcd31dcd9ae189ae7f3fc665de34/python-utils/update_tag_references.py) helper script we use during release.

MaxJPRey (approved):

jacobevansgit (approved):
Sweet!

Revathyvenugopal162 (approved):

PR #1549, opened 2026-09-09
refactor: use lockfile to generate wheelhouse

SMoraisAnsys:
**I'm not sure if I already raised that point but given that I couldn't find any history in GitHub, I'm opening this PR as both a POC and discussion place.**

---

This PR changes consist in using the project lockfile when building the wheelhouse for `uv`-based projects. This was already done with `poetry` and I don't see a reason not to do the same with `uv`. 

Pros:
- Keeps the wheelhouse reproducible for `uv`-based projects.
- Solves an issue related to `pip` resolution when developers define their dependencies with wide ranges (or no ranges at all). Indeed, in some cases we could end up with a very old version of a package or not finding one at all due to how `pip` does his resolution. An example of such scenario can be found in [here](https://github.com/ansys/pyaedt-toolkits-common/actions/runs/34333466933/job/102407399853?pr=495).
- No resolution required since the uv lock file already contains one (so faster run)

Cons:
- If the lockfile stale or isn't aligned with the project metadata, things could go wrong. But in that case, we have another bigger problem :)

### DIFF
--- a/build-wheelhouse/action.yml
+++ b/build-wheelhouse/action.yml
@@ -467,6 +467,8 @@ runs:
         INSTALL_TARGET: ${{ steps.specific-target-requested.outputs.install_target }}
         WORKING_DIRECTORY: ${{ inputs.working-directory }}
         GITHUB_WORKSPACE: ${{ github.workspace }}
+        USE_UV_LOCKFILE: ${{ steps.auto-detect-uv-lockfile.outputs.USE_UV_LOCKFILE }}
+        TARGET: ${{ inputs.target }}
       run: |
         ${ACTIVATE_VENV}
         cd ${WORKING_DIRECTORY}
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):
Awesome idea - thanks @SMoraisAnsys

jorgepiloto (approved):
Awesome!

AlejandroFernandezLuces (approved):

jorgepiloto (changes requested):
I can't approve this yet because wheelhouse generation for uv projects is not guaranteed to stay lockfile-driven.

request: `build-wheelhouse/action.yml` (hunk around line 479). In the new lockfile branch, `uv export` is called without `--frozen` or `--locked`:

`uv export $EXPORT_OPTIONS --output-file=wheel_reqs.txt`

This action already uses `uv sync --frozen` earlier, but this export step can still re-resolve when project metadata and `uv.lock` diverge, so the wheelhouse can be generated from a different dependency set than the checked-in lockfile. Please add `--frozen` (or `--locked`) to `uv export` so the wheelhouse stays reproducible and consistent with the install step.

jorgepiloto (approved):
Thanks for the update. I can confirm the suggestion is now applied in `build-wheelhouse/action.yml` with:

`uv export $EXPORT_OPTIONS --output-file=wheel_reqs.txt --frozen`

This keeps wheelhouse generation aligned with the lockfile path, so I'm approving.

PR #1551, opened 2026-09-09
refactor!: add setup python step in doc-style

SMoraisAnsys:
Imagine a runner where python was not already available 👀

This PR addresses that use case and aligns with what is already performed in other actions.

>[!WARNING]
This will change the behavior faced by users. I'm not expecting issues on users side but... Before we were using installed python, now we install it on the fly by default. This could be added to `v12` or we update the default approach.

>[!NOTE]
Easier approach would be to simply define that python is expected to be available and simply update the action's documentation.

### DIFF
--- a/doc-style/action.yml
+++ b/doc-style/action.yml
@@ -46,6 +46,40 @@ inputs:
 
   # Optional inputs
 
+  python-version:
+    description: |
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):

jorgepiloto (approved):
Thanks @SMoraisAnsys

moe-ad (approved):

PR #1552, opened 2026-09-10
refactor: remove quarto setup action

SMoraisAnsys:
Refactor _doc-build-linux to remove the need to use `quarto-dev/quarto-actions/setup`.

Generation went well in geometry, see below
image


_doc-build-linux/action.yml
```diff
@@ -203,11 +203,28 @@ runs:
...
-    - name: Install Quarto
+    - name: Install Quarto and TinyTeX
       if: ${{ inputs.needs-quarto == 'true' }}
-      uses: quarto-dev/quarto-actions/setup@8a96df13519ee81fd526f2dfca5962811136661b # v2.2.0
-      with:
-        tinytex: true
+      shell: bash
+      env:
+        QUARTO_VERSION: "1.10.18"
+        QUARTO_DEB_SHA256: "4bdf5a17df300003beb2f8f0e4dfe568e2ca0ff91318220c8537d2655015c430"
```
RobPasMue, line 211: We are going to completely forget about updating this... Not sure if it is a good idea.

SMoraisAnsys: This is currently what we are doing with Syft and we update it every year. I can create a workflow to track those updates one/twice a year if you want. Other option if to use the latest release but I would prefer to keep a pinned version.

RobPasMue: Setting up a reminder workflow - or a workflow that runs 1/2 times a year and updates it - works for me. Thanks @SMoraisAnsys

jorgepiloto (approved):

_doc-build-linux/action.yml
SMoraisAnsys: Added in 5912a2ef5ec596c4fe18ebd606266d72a1a1e610 (with minor modif to follow zizmor feedback)

.github/workflows/semestrial-tooling-update.yml
RobPasMue, line 1: Might it not be easier to open a PR with the update for the deps needed?

SMoraisAnsys: I'd rather avoid doing that as it would include a lot of code, require that we don't change the LOCs associated to the tooling, and also require that we only propose releases older than 1 week. I'm fearing that the investment to create the workflow with a PR generation won't pay the cost of it's creation / maintenance. I'll give it a try though

RobPasMue: Hmm... with a YML parser it shouldn't be that hard... you only need to find the proper entry and then, once you have it... search and replace.

I agree it might be more work and coding, but we might benefit from it honestly - and avoid having to do manual checks.

Also, I think we could probably skip the cooldown logic here to be honest.. but if you feel like implementing it, go ahead

.github/workflows/semestrial-tooling-update.yml
```diff
@@ -0,0 +1,73 @@
...
+          - Quarto
+
+          > [!NOTE]
+          > This tracking issue is intentionally lightweight and should be reviewed every six months to avoid stale toolchain versions.
+          EOF
+
+          gh issue create \
+            --title "$ISSUE_TITLE" \
+            --label "dependencies,ci" \
+            --assignee "@ansys/pyansys-core" \
```
MaxJPRey: In assignee  we supplie a team "@ansys/pyansys-core" where the CLI expects a user login.
Does it work? If not, we can remove that option and mention the team in the issue body,

.github/workflows/semestrial-tooling-update.yml
```diff
@@ -0,0 +1,73 @@
...
+          EXISTING_ISSUE=$(gh issue list \
+            --state open \
+            --search "$ISSUE_TITLE" \
+            --limit 20 \
+            --json number,title \
+            --jq ".[] | select(.title == \"$ISSUE_TITLE\") | .number" | head -n 1 || true)
+
+          if [[ -n "$EXISTING_ISSUE" ]]; then
+            echo "An open issue already exists: #$EXISTING_ISSUE"
+            exit 0
```
MaxJPRey: `exit 0`, ends only that step.
The next step’s `if: success()` therefore still permits creation

.github/workflows/semestrial-tooling-update.yml
SMoraisAnsys: Updated the PR to perform a PR. I've modified the call to add a reviewer and teams are allowed,
image

.github/workflows/semestrial-tooling-update.yml
SMoraisAnsys: I tested the python file locally and it worked correctly (I also took the opportunity to updated the Syft version on the fly).

python-utils/update_tooling_versions.py
```diff
@@ -0,0 +1,217 @@
...
+            "-H",
+            "User-Agent: actions",
+            f"https://api.github.com/repos/{repo}/releases?per_page={per_page}",
+        ],
+        text=True,
+    )
+    return json.loads(payload)
+
+
+def get_eligible_release(repo: str, cutoff_seconds: int = ONE_WEEK_SECONDS) -> str | None:
```
RobPasMue: I would avoid default inputs in these scripts. Simply use the module constant whenever you need it.

python-utils/update_tooling_versions.py
```diff
@@ -0,0 +1,217 @@
...
+def apply_tooling_updates(cutoff_seconds: int = ONE_WEEK_SECONDS) -> None:
+    """Resolve the eligible releases and update the action YAML files in place."""
+    syft_tag, syft_git_sha, quarto_tag = resolve_tooling_updates(cutoff_seconds=cutoff_seconds)
+    if syft_tag is None and quarto_tag is None:
+        print("No eligible release found for Syft or Quarto. Nothing to update.")
+
+    if syft_tag is not None:
+        update_syft_version(Path("build-wheelhouse/action.yml"), syft_tag, syft_git_sha)
+    if quarto_tag is not None:
+        update_quarto_version(Path("_doc-build-linux/action.yml"), quarto_tag)
```
RobPasMue: Wouldn't it be easier to do a grep on all files? That way if we ever add it in a different place we don't need to adapt it here

RobPasMue (approved):
Overall looks good - thanks @SMoraisAnsys!

RobPasMue (approved):

PR #1553, opened 2026-09-11
build: update dependencies

jorgepiloto:
This updates pinned dependency versions used by build and release workflows, and refreshes the lockfile so dependency definitions stay aligned across the repository.


SMoraisAnsys (changes requested):
Small request but could you also extend the pyproject.toml file with a configuration for cooldown on project ? This would be usefull since we do update things manually from time to time.

AFAIK we only have one dependency that is corporate so it would be something like
```
[tool.uv]
exclude-newer = "1 week"

[tool.uv.exclude-newer-package]
ansys-sphinx-theme = false
```

pyproject.toml
```diff
@@ -100,6 +100,12 @@ doc = [
     "tabulate==0.10.0"
 ]
 
+[tool.uv]
+exclude-newer = "1 week"
```
SMoraisAnsys, line 104: @jorgepiloto can you call `uv lock` again ? If I remember correctly, the added cfg to the pyproject.toml file should also appear in the lock file

SMoraisAnsys (approved):
Sry for the dealy, I forgot to add the submit button last week.
Except for this comment, LGTM :)

pyproject.toml
SMoraisAnsys: Yeah, see https://github.com/ansys/pyaedt/blob/main/uv.lock#L11-L20 for example

jorgepiloto: Done in https://github.com/ansys/actions/pull/1553/commits/8942e3c0c81a4bb5e55efc2e021bbe339b5fbc48

SMoraisAnsys: Yeah, saw that and resolved the comment :) Thanks @jorgepiloto

moe-ad (approved):

PR #1562, opened 2026-09-15
chore: update refs in workflows to self-repository syntax

moe-ad:
Follow-up to #1548 and #1550. The good thing is that the latest version of `zizmor` now includes this check (caught the missed updates in #1558).


jorgepiloto (approved):
Awesome, thanks @moe-ad

SMoraisAnsys:
Hey @moe-ad this distinction (`$/` and `./`) in our PR workflows was performed on purpose as explained in #1548 description.
```
If you look at the content of our PR workflows, it's now a mix of uses: ./my-action and uses: $/my-action.
The reason for that is  that we keep using uses: ./my-action in cases where we mock the behavior of an
external repo using our action. For example, first we checkout pymapdl, then we checkout the repo in
another directory, then we use uses: ./actions/my-action to explicitly force the execution of the action
from the second checkout.
```

I'm fine with changing that decision but it would have been great to wait a bit before merging to get more reviews (ofc this doesn't apply to urgent fixes and so on)

RobPasMue:
Yep, agreeing with @SMoraisAnsys comment - also, the contents in this PR could've waited so that others can review it. Especially since it is regarding a new way of calling the actions. We could've anticipated that the missing changes might have been on purpose..

Was there any urgency on merging? Also.. I thought we agreed internally to slow down on merging to this specific repository for internal reasons.

moe-ad:
> Hey @moe-ad this distinction (`$/` and `./`) in our PR workflows was performed on purpose as explained in #1548 description.
> 
> ```
> If you look at the content of our PR workflows, it's now a mix of uses: ./my-action and uses: $/my-action.
> The reason for that is  that we keep using uses: ./my-action in cases where we mock the behavior of an
> external repo using our action. For example, first we checkout pymapdl, then we checkout the repo in
> another directory, then we use uses: ./actions/my-action to explicitly force the execution of the action
> from the second checkout.
> ```
> 
> I'm fine with changing that decision but it would have been great to wait a bit before merging to get more reviews (ofc this doesn't apply to urgent fixes and so on)

I missed your comment in #1548. I thought the goal was to update all references, so when `zizmor` flagged the remaining "legacy" references earlier today, I just thought you must have missed those.

> Yep, agreeing with @SMoraisAnsys comment - also, the contents in this PR could've waited so that others can review it. Especially since it is regarding a new way of calling the actions. We could've anticipated that the missing changes might have been on purpose..
> 
> Was there any urgency on merging? Also.. I thought we agreed internally to slow down on merging to this specific repository for internal reasons.

My bad on merging too early. I considered it a trivial PR and just merged so I could move forward with #1558.

-------------------------------
FYI @SMoraisAnsys that we still achieve a similar "behavior mocking" with the new Syntax. From what I understand, the new syntax is a cleaner way to do the same thing + the extra benefit that the self-reference will be respected in caller workflows + more robustness of course (no need for [weird hacks](https://github.com/ansys/actions/blob/12a5c264b20d2006954e97d7978592d9d353bd66/.github/workflows/ci_cd_pr.yml#L520) we used to have, which is another added value).

PR #1563, opened 2026-09-15
refactor: exclude build and wheel from generated wheelhouse

moe-ad:
>[!NOTE]
>Whoever reviews this PR should also check #1535.

While working on #1535, I discovered a "bug" (can I call it that?) in how we are installing `build-wheelhouse` action requirements (i.e. `build` and `wheel` in this case). To summarize, simply running `poetry add <package>` implicitly means "include this package in the main dependencies of the project" ==> `build` and `wheel` will be included in the wheelhouse.

For example, this is the wheelhouse generated for `pytwin` in the testing workflows prior to this PR:
https://github.com/ansys/actions/actions/runs/34959003573?pr=1535
image

That has been fixed with this PR:
https://github.com/ansys/actions/actions/runs/34962405659?pr=1563
image


jorgepiloto (approved):
Thanks for fixing this, @moe-ad

SMoraisAnsys (approved):

build-wheelhouse/action.yml
```diff
@@ -453,8 +453,9 @@ runs:
       run: |
         ${ACTIVATE_VENV}
         if [[ "${BUILD_BACKEND}" == 'poetry' ]]; then
+          # Scope build tooling to a non-main group so poetry export skips it and it stays out of the wheelhouse.
           grep -v '^\s*#' "${GITHUB_ACTION_PATH}/requirements.txt" | grep -v '^\s*$' | grep -v '^\s' | while read -r line; do
-            poetry add "${line//\'/\"}" --lock
+            poetry add "${line//\'/\"}" --group build-and-wheel --lock
```
RobPasMue, line 458: Ha! Nice hack 😄

RobPasMue (approved):
LGTM!

PR #1567, opened 2026-09-16
feat: allow doc-changelog to select Towncrier config file

### LINKED ISSUE #952: Provide option to specify configuration file for `doc-changelog` action

For repositories containing a `pyproject.toml` and `towncrier.toml`, allow the user to specify using the `towncrier.toml` file. Currently, if both of these files exists it automatically defaults to the `pyproject.toml`

Requested by @ansjhenry 

Add an optional input to the `doc-changelog` action to specify the name of the configuration file. By default, it will be `pyproject.toml`

_No response_

jorgepiloto:
This change adds a `config-file` input to `ansys/actions/doc-changelog` so maintainers can choose which Towncrier config file to use. The action now defaults to `pyproject.toml`, supports `towncrier.toml` when explicitly set, and fails with a clear error if the selected file or `[tool.towncrier]` section is missing.

It also wires the selected config file through both Towncrier config-reading steps and updates the usage example. Closes #952.


doc-changelog/action.yml
```diff
@@ -380,10 +392,11 @@ runs:
 
         repo_name = os.environ.get("REPO_NAME")
         org_name = os.getenv("GITHUB_REPOSITORY_OWNER")
+        towncrier_config_file = os.environ.get("TOWNCRIER_CONFIG_FILE")
```
SMoraisAnsys:
```suggestion
        towncrier_config = pathlib.Path(os.environ.get("TOWNCRIER_CONFIG_FILE", "pyproject.toml"))
```
Aligning with line 355 changes. If `os.environ.get("TOWNCRIER_CONFIG_FILE")` is prefered, then this one can stay and the other line need to be updated. I'm fine with both.

SMoraisAnsys (commented):
LGTM. Since we don't test the use of towncrier.toml in our CI, could you check that the changes work in a test repo ?

doc-changelog/action.yml
jorgepiloto: Thanks for suggesting this, applied in 6b33974b

clatapie (approved):
LGTM with the same test question as @SMoraisAnsys

klmcadams (approved):
LGTM with same testing comment as @SMoraisAnsys & @clatapie

PR #1568, opened 2026-09-16
feat: improve releasing with `doc-deploy-changelog` action

### LINKED ISSUE #626: Bug located in how the Changelog Action interacts with GitHub Environments

### 🐞 Description of the bug

Our setup for publishing with the trusted publisher workflow leverages environments to secure access to publishing new versions of packages. The intended workflow for environments with approvals is that the person triggering the workflow is not able to approve their own request.

The doc deploy changelog workflow circumvents this security by triggering a follow up workflow run as the pyansys bot, this means a user is then able to approve their own publish. This may be acceptable or it may not, but ideally the follow up run would be triggered as the same user who initially pushed the tag.

### 📝 Steps to reproduce

Publish a tag on a repository that has trusted publisher enabled with an environment

Windows

_No response_

3.9

### 📦 Installed packages

n/a

moe-ad:
Closes #626.
This PR updates `doc-deploy-pr` to be usable in scenarios where users want to prevent multiple workflow runs, such as in #626. I will try to test with a sample project, but this can be reviewed now regardless. At least, I will like to know what people think about the approach before investing further time.

The reasonable approach I could think of is adapting `doc-deploy-changelog` in way that makes it compatible with single workflow runs, without too much extra configuration by the user. The idea is to have a "same-workflow" switch that controls legacy behavior (i.e. multi-workflow) vs single-workflow behavior. In the later case the action will produce an output containing the release SHA, which the user can then use to setup the remaining release steps.

#### Summary
A couple of things to keep in mind:
  - Our actions already have the "knobs" needed to make this work, especially the `checkout` input that is exposed in most.
  - `GITHUB_TOKEN` behaves such that it will not create a new workflow run when used to push to a PR (see [relevant section in the docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow)). That behavior is also important for making this work.
  - The main limitation of @RobPasMue's suggestion in #626 is that there is no way to have a centralized mechanism to obtain a token that impersonates each user that pushes a tag (at least I can't see how myself, maybe someone might have fresh ideas).

Having said that, these are the changes I made:
- I have added a `same-workflow` input to `doc-deploy-changelog` whose purpose is to simply control whether the action should exit with error or not.
- If this input is true, the action will produce a `release-commit-sha` output that will point to the new commit that contains the changes made by the action (i.e. the updated changelog file where all the fragments have been combined).
- If this input is true, users are expected to have used `GITHUB_TOKEN` (which will not trigger a re-run) instead of `PYANSYS_CI_BOT_TOKEN` to ensure the current workflow keeps running. You are probably thinking "but we are already not exiting with error, why is this necessary...", well, remember workflows can still be canceled because of concurrency?
- Every subsequent job in the release workflow must then be configured to work from `release-commit-sha`. This is where the `checkout` input comes in handy. Basically users can do a checkout step before calling an action and pass `checkout: false`. The will ensure everything runs from the new commit.
- `release-github` action has also been updated such that it supports the single-workflow scenario. It now exposes a `tag-name` input which is passed to the underlying action being called behind the scenes.
- And to the most important part of the PR, documentation. This is actually the core of the PR, going through it will also help whoever is reviewing the PR understand the approach better.
  - A new `release-workflow.rst` file has been added that juxtaposes the legacy mode and the single workflow mode.
  - Existing documentation of `doc-deploy-changelog` action has been updated.


doc/source/conf.py
```diff
@@ -62,10 +62,10 @@
...
         "version_match": switcher_version,
     },
-    "cheatsheet": {
-        "file": "cheat_sheet.qmd",
-        "title": "Actions cheat sheet",
-    },
+    # "cheatsheet": {
+    #     "file": "cheat_sheet.qmd",
+    #     "title": "Actions cheat sheet",
+    # },
```
jorgepiloto:
```suggestion
    "cheatsheet": {
        "file": "cheat_sheet.qmd",
        "title": "Actions cheat sheet",
    },
```

jorgepiloto (approved):
Thanks for adding this, Mo. I don't se any blocking issues despite restoring the cheatsheet.

jorgepiloto:
> Thanks for adding this, Mo. I don't se any blocking issues despite restoring the cheatsheet.

Done in https://github.com/ansys/actions/pull/1568/commits/628ae397e9ce05b208ed572f238b1109a7b0656f

moe-ad:
> Thanks for adding this, Mo. I don't se any blocking issues despite restoring the cheatsheet.

My bad. I commented it out to test building the doc locally and forgot to revert. Thanks for restoring.

RobPasMue (approved):
Works for me although I see it a bit "hacky" not gonna lie... From a developer's expectations... if I am running a workflow for tag "vX.Y.Z" and then the tag gets recreated... but my workflow keeps running pointing to the old tag (but the actions seem to use the new tag/commit)... this seems pretty weird and unexpected.

For me, cancelling the workflow and re-executing is definiteñy more fail-safe and consistent. But if anybody wants to use this.. up to them. I am expecting lots of issues, misconfigurations and debugging sessions for people that enable this.

moe-ad:
> Works for me although I see it a bit "hacky" not gonna lie... From a developer's expectations... if I am running a workflow for tag "vX.Y.Z" and then the tag gets recreated... but my workflow keeps running pointing to the old tag (but the actions seem to use the new tag/commit)... this seems pretty weird and unexpected.
> 
> For me, cancelling the workflow and re-executing is definiteñy more fail-safe and consistent. But if anybody wants to use this.. up to them. I am expecting lots of issues, misconfigurations and debugging sessions for people that enable this.

It is a bit hacky, agreed. But as you said, people that choose to use this will be well aware of what to expect. I will also request a review from @da1910 to drop his two cents on what he thinks.

moe-ad:
#### Tests

##### Legacy (i.e. multi-workflow) using this branch + GH environment where "Prevent self-review" is activated.
- First workflow run - https://github.com/ansys/test-repo-moe/actions/runs/35891305234
- Second workflow run triggered by doc-deploy-changelog - https://github.com/ansys/test-repo-moe/actions/runs/35891417632. Of course as @da1910 reported, I was able to approve this workflow myself:
image

This test confirms two things: what was reported and that the legacy behavior is not impacted by the changes in this PR. So existing users do not have to do anything to their workflows when this is released.

##### Single workflow approach (i.e. using changes in this branch) + GH environment where "Prevent self-review" is activated 
- Workflow run - https://github.com/ansys/test-repo-moe/actions/runs/35889727415
- In this workflow run, I wasn't able to approve as expected. I had to use admin privileges to skip the approval:
image

As can be seen, all the jobs executed correctly. Most importantly, if you check the [v0.9.0 GH release](https://github.com/ansys/test-repo-moe/releases#release-v0.9.0) (what was released via single workflow approach), it points to the correct SHA (`78b08e1...`) on the [release branch](https://github.com/ansys/test-repo-moe/commits/release/0.9/), and the PR for updating the changelog file on the main branch [gets opened](https://github.com/ansys/test-repo-moe/pull/20) as usual.

SMoraisAnsys:
If this PR can wait next week, I'd gladly have a look when I'm back :) Otherwise, I'll see the result afterward :D

PR #1569, opened 2026-09-16
fix(code-style): align uv install behavior with poetry

### LINKED ISSUE #926: code-style has different behavior between uv and poetry backends

### 🐞 Description of the bug

If code-style is run on a project using poetry, and `skip-install` is set to False, then the entire project is installed with all dependencies:

https://github.com/ansys/actions/blob/1240aef678c0335bbc502c6f8097e17d1e514246/code-style/action.yml#L239

If however, code-style is run on a project using uv with the same configuration, then only the dependencies in the wheel are installed:

https://github.com/ansys/actions/blob/1240aef678c0335bbc502c6f8097e17d1e514246/code-style/action.yml#L241

We were previously depending on the poetry behavior, because as part of the project installation we install mypy as a dev dependency. mypy needs to be installed in the environment itself to work. Since mypy is a dev dependency, it's installed with a poetry project, but not installed with a uv project.

uv should probably just do a `uv sync`, such that the behavior is more similar to the poetry case.

### 📝 Steps to reproduce

See an example here https://github.com/ansys-internal/grantami-system/actions/runs/16201141925/job/45740217086 of CI running, where mypy is not available after project install

Windows

N/A

3.12

### 📦 Installed packages

```shell
Actions v10.0.12
```

jorgepiloto:
When `skip-install` is `false` and a `uv.lock` file is present, `code-style` now runs `uv sync --frozen` instead of `uv sync --frozen --no-dev`. This installs dev dependencies like the Poetry backend and closes the behavior gap reported in #926.

Closes #926


doc/source/changelog/926.fixed.md
```diff
@@ -0,0 +1 @@
+Align uv project installation in `code-style` with Poetry by installing dev dependencies when `uv.lock` is present
```
SMoraisAnsys, line 1: Please pay attention to the PR content, there are quite a few containing an extra fragment file :/

SMoraisAnsys (approved):
LGTM except for my comment

PR #1570, opened 2026-09-16
fix: run doc-style dependency install in an isolated virtualenv

jorgepiloto:
This updates doc-style to create a local .venv before installing action dependencies and prepend that environment to GITHUB_PATH. The install step now runs in an isolated environment regardless of how Python is provided to the job.


doc-style/action.yml
```diff
@@ -225,6 +225,16 @@ runs:
...
 
+    - name: "Create isolated virtual environment"
+      shell: bash
+      run: |
+        python -m venv .venv
+        if [[ ${RUNNER_OS} == 'Windows' ]]; then
+          echo "${GITHUB_WORKSPACE}/.venv/Scripts" >> $GITHUB_PATH # zizmor: ignore[github-env]
+        else
+          echo "${GITHUB_WORKSPACE}/.venv/bin" >> $GITHUB_PATH # zizmor: ignore[github-env]
+        fi
```
SMoraisAnsys: The step is never leveraged in the rest of the file. I assume that the next job should be the one using the venv, right ?

SMoraisAnsys (changes requested):

doc-style/action.yml
jorgepiloto: This approach should work since it takes preference from the Python of the system. However, I think I'll follow the per-step activation approach that we have in other actions. I feel its safer than modifying the `GITHUB_PATH`.

SMoraisAnsys: You are right, I missed that! I would advise to keep things consistent with other places for the moment. We can perform this modification in a follow up PR :)

jorgepiloto: I just modified it in https://github.com/ansys/actions/pull/1570/commits/88a58fee969a4f6e9585863943f568e74ba4b973 so it is consistent with the rest of the actions.

doc-style/action.yml
```diff
@@ -225,15 +225,40 @@ runs:
...
 
+    - name: "Expose rst2html on PATH"
+      shell: bash
+      run: |
+        if [[ ${RUNNER_OS} == 'Windows' ]]; then
+          cp "${GITHUB_WORKSPACE}/.venv/Scripts/rst2html.exe" "${pythonLocation}/Scripts/rst2html.exe"
+        else
+          sudo ln -sf "${GITHUB_WORKSPACE}/.venv/bin/rst2html" /usr/local/bin/rst2html
+        fi
+
```
SMoraisAnsys, line 252: Why are those LOC added to the PR ?

SMoraisAnsys (approved):

